PR #4, opened 2021-12-16
Add Notepad++ & ComparePlus plugin

Ana06:
`notepadplusplus.vm` is a wrapper for the community package `notepadplusplus` that updates the config file and disable auto-updates. The original package I have migrated had an uninstaller that reenabled the updates. But we can't really know if they were enabled before installing the package (in case Notepad++ was already installed) and we keep Notepad++ that may not have been installed (as it is installed by the dependency). So I would suggest to keep things simple and do not provide an uninstaller in this case. Any concerns with this?

I have also added a `notepadplusplus-compare.vm` package to install the ComparePlus plugin for Notepad++ that I like to use.

### DIFF
--- a/packages/common.vm/tools/vm.common/vm.common.psm1
+++ b/packages/common.vm/tools/vm.common/vm.common.psm1
@@ -693,7 +693,7 @@ function VM-Add-To-Right-Click-Menu {
     }
     Set-ItemProperty -LiteralPath "HKCR:\$key\shell\$menuKey\command" -Name '(Default)' -Value $command -Type String
   } catch {
-    FE-Write-Log "ERROR" "Failed to add $menuKey to right-click menu"
+    VM-Write-Log "ERROR" "Failed to add $menuKey to right-click menu"
   }
 }
 
@@ -724,6 +724,6 @@ function VM-Remove-From-Right-Click-Menu {
       Remove-Item -LiteralPath "HKCR:\$key\shell\$menuKey" -Recurse
     }
   } catch {
-    FE-Write-Log "ERROR" "Failed to remove $menuKey from right-click menu"
+    VM-Write-Log "ERROR" "Failed to remove $menuKey from right-click menu"
   }
 }
--- a/packages/notepadpp.plugin.compare.vm/notepadplusplus.vm.nuspec
+++ b/packages/notepadpp.plugin.compare.vm/notepadplusplus.vm.nuspec
@@ -1,0 +1,13 @@
+<?xml version="1.0"?>
+<package xmlns="http://schemas.microsoft.com/packaging/2015/06/nuspec.xsd">
+  <metadata>
+    <id>notepadpp.plugin.compare.vm</id>
+    <version>2.0.1</version>
+    <description>ComparePlus plugin for Notepad++</description>
+    <authors>Pavel Nedev</authors>
+    <dependencies>
+      <dependency id="common.vm" />
+      <dependency id="notepadpp.vm" />
+    </dependencies>
+  </metadata>
+</package>
--- a/packages/notepadpp.plugin.compare.vm/tools/chocolateyinstall.ps1
+++ b/packages/notepadpp.plugin.compare.vm/tools/chocolateyinstall.ps1
@@ -1,0 +1,29 @@
+$ErrorActionPreference = 'Stop'
+Import-Module vm.common -Force -DisableNameChecking
+
+try {
+  $toolName = "ComparePlugin"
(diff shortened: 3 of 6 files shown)


packages/notepadplusplus.vm/tools/chocolateyInstall.ps1
```diff
@@ -0,0 +1,32 @@
+$ErrorActionPreference = 'Stop'
```
MalwareMechanic, line 1: File is a duplicate and has an uppercase letter. I wonder if we can have GH hook to check for lowercase file paths?

packages/notepadplusplus.vm/tools/chocolateyinstall.ps1
```diff
@@ -0,0 +1,32 @@
...
+      $notepad = Get-Process -Name "notepad++" -ErrorAction SilentlyContinue
+      if ($notepad -ne $null) {
+        $notepad | ForEach-Object {$_.CloseMainWindow() | Out-Null}
+      }
+      Start-Sleep -Seconds 2
+    }
+  }
+
+  if (-Not (Test-Path -Path $configFile)) {
+    Write-Host "Can't find config.xml to modify"
```
MalwareMechanic, line 25: Instead of `Write-Host`, use `VM-Write-Log "ERROR" "Can't find config.xml to modify"`

packages/notepadplusplus-compare.vm/tools/chocolateyinstall.ps1
```diff
@@ -0,0 +1,27 @@
...
+  $packageArgs = @{
+    packageName    = ${Env:ChocolateyPackageName}
+    unzipLocation  = $toolDir
+    url            = $zipUrl
+    checksum       = $zipSha256
+    checksumType   = 'sha256'
+    url64bit       = $zipUrl_64
+    checksum64     = $zipSha256_64
+  }
+  Install-ChocolateyZipPackage @packageArgs
```
MalwareMechanic, line 24: Is there a DLL or file we can check for via `VM-Assert-Path` to verify the files were successfully installed?

MalwareMechanic (changes requested):

packages/notepadplusplus-compare.vm/tools/chocolateyinstall.ps1
```diff
@@ -0,0 +1,27 @@
+$ErrorActionPreference = 'Stop'
+Import-Module vm.common -Force -DisableNameChecking
+
+try {
+  $pluginsDir= Join-Path ${Env:ProgramFiles} "Notepad++\plugins" -Resolve
+  $toolDir = New-Item "C:\Program Files\Notepad++\plugins\ComparePlugin" -itemtype directory
```
MalwareMechanic, line 6: Let's use the `$pluginsdir` instead of a hard-coded path, like:
`$toolDir = New-Item "$pluginsDir\ComparePlugin" -itemtype directory`

MalwareMechanic (changes requested):

packages/notepadplusplus.vm/tools/chocolateyInstall.ps1
Ana06: We could. Although this is something easy to fix if we miss it. I would like to first focus on automatic updates or fixes.

packages/notepadplusplus.vm/tools/chocolateyinstall.ps1
Ana06: I can use `Write-Host` 👍  Although I think this is more a `WARN` than a `ERROR`, notepad is still installed, just the updates are not disabled. 🤔

packages/notepadplusplus-compare.vm/tools/chocolateyinstall.ps1
```diff
@@ -0,0 +1,29 @@
...
+    packageName    = ${Env:ChocolateyPackageName}
+    unzipLocation  = $toolDir
+    url            = $zipUrl
+    checksum       = $zipSha256
+    checksumType   = 'sha256'
+    url64bit       = $zipUrl_64
+    checksum64     = $zipSha256_64
+  }
+  Install-ChocolateyZipPackage @packageArgs
+  VM-Assert-Path (Join-Path $toolDir $toolName.dll)
```
MalwareMechanic, line 26: Slight nitpick, can we use double quotes for: `"$toolName.dll"`

MalwareMechanic (changes requested):

MalwareMechanic (approved):

PR #7, opened 2021-12-20
Use a fix commit to avoid failures

Ana06:
Use a fix commit instead of master branch to avoid that a future update changes the hash.

### DIFF
--- a/packages/asreproast.vm/asreproast.vm.nuspec
+++ b/packages/asreproast.vm/asreproast.vm.nuspec
@@ -2,7 +2,7 @@
 <package xmlns="http://schemas.microsoft.com/packaging/2015/06/nuspec.xsd">
   <metadata>
     <id>asreproast.vm</id>
-    <version>0.0.0.20211029</version>
+    <version>0.0.0.20180925</version>
     <authors>HarmJ0y</authors>
     <description>Project that retrieves crackable hashes from KRB5 AS-REP responses for users without kerberoast preauthentication enabled.</description>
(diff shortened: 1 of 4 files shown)


packages/asreproast.vm/asreproast.vm.nuspec
```diff
@@ -2,11 +2,11 @@
 <package xmlns="http://schemas.microsoft.com/packaging/2015/06/nuspec.xsd">
   <metadata>
     <id>asreproast.vm</id>
-    <version>0.0.0.20211029</version>
+    <version>0.0.0.20180925</version>
```
Ana06, line 5: As this package hasn't been pushed yet, I can use the tool creation date.

packages/peid.vm/peid.vm.nuspec
```diff
@@ -2,7 +2,7 @@
 <package xmlns="http://schemas.microsoft.com/packaging/2015/06/nuspec.xsd">
   <metadata>
     <id>peid.vm</id>
-    <version>0.95</version>
+    <version>0.95.20081103</version>
```
Ana06, line 5: I have used the tool official date. I could also use today's date, but I though that looks nicer and as the package hasn't been pushed yet, it is not really important.

MalwareMechanic: @Ana06 Make sure to only put dates in the fourth segment 😄

MalwareMechanic (approved):

Issue #6, opened 2021-12-17
Add linter


Ana06:
Ideas for linting:
- [ ] Powershell style linter
- [ ] Ensure nuspec is modified is packaged is modified
- [ ] Ensure file paths use lowercase

Any other ideas?

mr-tz:
Make the linter output easy to see, currently lines like `Progress: Downloading PSScriptAnalyzer 1.20.0... 0%` add a lot of noise.

PR #8, opened 2021-12-20
Ensure nuspec is modified if package is modified

Ana06:
This actions fails if a package has been modified, but the nuspec hasn't. It doesn't check that the version of the nuspec has been (correctly) modified, but let's start simple and improve it after some testing.

Partially addresses: https://github.com/mandiant/VM-Packages/issues/6

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -7,8 +7,35 @@ on:
     branches: [ main ]
 
 jobs:
+  check_nuspec:
+    runs-on: windows-2016
+    steps:
+      - name: Checkout code
+        uses: actions/checkout@v2
+      - name: Get changed files
+        id: files
+        uses: Ana06/get-changed-files@v2.0.0
+        with:
+          filter: 'packages/*/**'
+      - name: Check nuspec is modified for every modified package
+        # It runs only if there are modified files
+        if: steps.files.outputs.added_modified != ''
+        run: |
+          $changed_files = "${{ steps.files.outputs.added_modified }}".Split(" ")
+          $nuspecs = [string]@($changed_files | where { $_ -Like "*.nuspec" })
+          $others = @($changed_files | where { $_ -NotLike "*.nuspec" })
+          foreach ($file in $others) {
+              if ($file -match "packages/[^/]*") {
+                  $package = $matches.0
+                  if (!$nuspecs.contains($package)) {
+                      Write-Error "The version in $package needs to be modified"
+                  }
+              }
+          }
+
   build_test_upload:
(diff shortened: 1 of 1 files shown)


Ana06:
I will remove 7d3b446 before merging. I have added as it may help with the review (as you can see how it fails).

.github/workflows/ci.yml
```diff
@@ -7,8 +7,36 @@ on:
...
+        if: steps.files.outputs.added_modified != ''
+        run: |
+          $changed_files = "${{ steps.files.outputs.added_modified }}".Split(" ")
+          $nuspecs = @($changed_files | where { $_ -Like "*.nuspec" })
+          $others = @($changed_files | where { $_ -NotLike "*.nuspec" })
+          foreach ($file in $others) {
+              if ($file -match "packages/([^/]*)") {
+                  $package = $matches.1
+                  $nuspec = "packages/$package/$package.nuspec"
+                  if (!$nuspecs.contains($file)) {
```
MalwareMechanic, line 31: I think the path `"packages/$package/$package.nuspec"` may not always hold true (e.g., see path for [notepadpp.vm's .nuspec file](https://github.com/mandiant/VM-Packages/blob/main/packages/notepadpp.vm/notepadplusplus.vm.nuspec)). Also wouldn't it be easier to check if a `$nuspecs` is not empty?

Ana06: > I think the path "packages/$package/$package.nuspec" may not always hold true (e.g., see path for notepadpp.vm's .nuspec file)

isn't that nuspec name wrong? I think the directory and the nuspec should have the same name. 🤔 

> Also wouldn't it be easier to check if a $nuspecs is not empty?
we could. But if you edit two packages and only one nuspec is updated, then the action would work. Ideailly we should check that the version has been correctly updated. Remember that if the action success in a PR merge, the package is pushed to myget.

Ana06: One easier alternative would be to keep $nuspecs as string and check it contains `packages/$package/`. This may be a more robust solution. 🤔

MalwareMechanic (approved):

.github/workflows/ci.yml
MalwareMechanic: I like the new version 😄

Ana06:
> > I think the path "packages/$package/$package.nuspec" may not always hold true (e.g., see path for notepadpp.vm's .nuspec file)
> 
> isn't that nuspec name wrong? I thought the directory and the nuspec should have the same name. 🤔 

@MalwareMechanic should we fix this? 🤔

PR #9, opened 2021-12-20
Do not check hash for updates with same URL

Ana06:
Some projects use the same download URL for all versions of the package. This causes that the hash changes and the package breaks. Ignore checksums for these packages.

Note we are already not checking hashes for all installed software. We currently do something similar for some Python libraries. We should make clear in the project documentation that there is no guarantee the packages are benign. These packages are intended to be used to analyze malware that may be inherently dangerous.

### DIFF
--- a/packages/common.vm/tools/vm.common/vm.common.psm1
+++ b/packages/common.vm/tools/vm.common/vm.common.psm1
@@ -482,11 +482,11 @@ function VM-Install-From-Zip {
     [string] $category,
     [Parameter(Mandatory=$true, Position=2)]
     [string] $zipUrl,
-    [Parameter(Mandatory=$true, Position=3)]
+    [Parameter(Mandatory=$false)]
     [string] $zipSha256,
-    [Parameter(Mandatory=$false, Position=4)]
+    [Parameter(Mandatory=$false)]
     [string] $zipUrl_64,
-    [Parameter(Mandatory=$false, Position=5)]
+    [Parameter(Mandatory=$false)]
     [string] $zipSha256_64,
     [Parameter(Mandatory=$false)]
     [bool] $consoleApp=$false,
--- a/packages/hashmyfiles.vm/tools/chocolateyinstall.ps1
+++ b/packages/hashmyfiles.vm/tools/chocolateyinstall.ps1
@@ -6,11 +6,9 @@ try {
   $category = 'Utilities'
 
   $zipUrl = "https://www.nirsoft.net/utils/hashmyfiles.zip"
-  $zipSha256 = "89db49ec6a3e50f1d76da97ac1289272d1b09b9a330d36f65fdbe1f010f1ae8b"
   $zipUrl_64 = "https://www.nirsoft.net/utils/hashmyfiles-x64.zip"
-  $zipSha256_64 = "be2dc5b9613b72ca44e60b7a1b5332593a868079638ded37cc3ad120e7182b0b"
 
-  $executablePath = (VM-Install-From-Zip $toolName $category $zipUrl $zipSha256 $zipUrl_64 $zipSha256_64)[-1]
+  $executablePath = (VM-Install-From-Zip $toolName $category $zipUrl -zipUrl_64 $zipUrl_64)[-1]
   VM-Add-To-Right-Click-Menu $toolName "HashMyFiles" "`"$executablePath`" `"%1`"" "file"
   VM-Add-To-Right-Click-Menu $toolName "HashMyFiles" "`"$executablePath`" `"%1`"" "directory"
 } catch {
--- a/packages/uniextract2.vm/uniextract2.vm.nuspec
+++ b/packages/uniextract2.vm/uniextract2.vm.nuspec
@@ -2,7 +2,7 @@
 <package xmlns="http://schemas.microsoft.com/packaging/2011/08/nuspec.xsd">
   <metadata>
     <id>uniextract2.vm</id>
-    <version>2.0.0.5</version>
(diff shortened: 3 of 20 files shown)


Ana06:
I haven't changed MAP (which is also using a URL for all versions) because it uses HTTP (http://sandsprite.com/CodeStuff/map_setup.exe) and Chocolatey complains (with good reasons). I have opened an issue for this one: https://github.com/mandiant/VM-Packages/issues/11

packages/common.vm/tools/vm.common/vm.common.psm1
```diff
@@ -482,7 +482,7 @@ function VM-Install-From-Zip {
     [string] $category,
     [Parameter(Mandatory=$true, Position=2)]
     [string] $zipUrl,
-    [Parameter(Mandatory=$true, Position=3)]
+    [Parameter(Mandatory=$false, Position=3)]
```
MalwareMechanic, line 485: Does this actually work? Seems like it's used down at like 514. Perhaps the empty/null value allows it to not use the checksum?

itzdan (approved):

packages/common.vm/tools/vm.common/vm.common.psm1
Ana06: You can see it working in the tests: https://github.com/mandiant/VM-Packages/runs/4585639925?check_suite_focus=true

MalwareMechanic: Neat! 📷

packages/hashmyfiles.vm/tools/chocolateyinstall.ps1
```diff
@@ -6,9 +6,7 @@ try {
   $category = 'Utilities'
 
   $zipUrl = "https://www.nirsoft.net/utils/hashmyfiles.zip"
-  $zipSha256 = "89db49ec6a3e50f1d76da97ac1289272d1b09b9a330d36f65fdbe1f010f1ae8b"
   $zipUrl_64 = "https://www.nirsoft.net/utils/hashmyfiles-x64.zip"
-  $zipSha256_64 = "be2dc5b9613b72ca44e60b7a1b5332593a868079638ded37cc3ad120e7182b0b"
 
   $executablePath = (VM-Install-From-Zip $toolName $category $zipUrl $zipSha256 $zipUrl_64 $zipSha256_64)[-1]
```
MalwareMechanic, line 11: Can you remove the unreferenced variables: `$zipSha256` and `$zipSha256_64`?

MalwareMechanic (changes requested):

packages/hashmyfiles.vm/tools/chocolateyinstall.ps1
Ana06: If I do that, it doesn't work anymore 😮

MalwareMechanic: Ah, I bet it's because we're using positional arguments and marking them as not mandatory... so we should probably remove the positional component and make them be explicitly set if used. Otherwise, we'll need to pass in null/empty values for the positions. If we do remove the positional component, it may affect other packages as well that would then need to explicitly set them.

MalwareMechanic: Idea:

Switch positional arguments 3 & 4 in: https://github.com/mandiant/VM-Packages/blob/3556aab8fd482483c0da4d3f732ca88d7e3cce0e/packages/common.vm/tools/vm.common/vm.common.psm1#L485-L488 then update the packages that use them accordingly. This will allow the hashes to be optional and if used, they don't have to be explicitly declared.

Ana06: @MalwareMechanic 
> Switch positional arguments 3 & 4

But there may be other packages that specify `$zipSha256` and not `zipUrl_64`. I think we need that both arguments are not positional.

packages/uniextract2.vm/uniextract2.vm.nuspec
```diff
@@ -2,7 +2,7 @@
 <package xmlns="http://schemas.microsoft.com/packaging/2011/08/nuspec.xsd">
   <metadata>
     <id>uniextract2.vm</id>
-    <version>2.0.0.5</version>
+    <version>2.0.0.20220113</version>
```
Ana06, line 5: This version is tricky. From Chocolatey documentation:

> So if the package was ruby and the version was 2.0.0-p353, the package is 2.0.0.35300 (adding the two zeroes at the end)

But there is one case that Chocolatey documentation doesn't cover.  As I understand the documentation `Ruby 2.7.0-preview3` would be `2.7.0.300` and `Ruby 2.7.0-rc1` (released after `Ruby 2.7.0-preview3`) would be `Ruby 2.7.0.100`, which would conflict with `Ruby 2.7.0-preview1`. Or I am understanding something wrong?

I have use the date format while we think how to resolve this. The previous version was already not following Chocolatey documentation (as the fourth segment needed two extra 0s).

packages/hashmyfiles.vm/tools/chocolateyinstall.ps1
MalwareMechanic: Hmm you're probably right. We'd need to update the current packages to use the explicit syntax instead of using positional arguments

Ana06: I have done it, check last commit 😉

MalwareMechanic (approved):

Issue #3, opened 2021-12-16
Add priority packages


Ana06:
Add the following priority packages:
- [X] cyberchef (cyberchef.vm - #15)
- [x] notepadplusplus (notepadplusplus.vm - #4)
- [X] pd64.exe (processdump.vm - #19)
- [x] wireshark (wireshark.vm - #12)

PR #12, opened 2021-12-20
Add wireshark.vm

Ana06:
Add new package: Wireshark. It is a wrapper for the community package.

Partially addresses: https://github.com/mandiant/VM-Packages/issues/3

### DIFF
--- a/packages/wireshark.vm/tools/chocolateyinstall.ps1
+++ b/packages/wireshark.vm/tools/chocolateyinstall.ps1
@@ -1,0 +1,15 @@
+$ErrorActionPreference = 'Stop'
+Import-Module vm.common -Force -DisableNameChecking
+
+try {
+  $toolName = 'wireshark'
+  $category = 'Networking Tools'
+
+  $shortcutDir = Join-Path ${Env:TOOL_LIST_DIR} $category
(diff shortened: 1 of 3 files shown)


packages/wireshark.vm/tools/chocolateyinstall.ps1
```diff
@@ -0,0 +1,10 @@
+$ErrorActionPreference = 'Stop'
+Import-Module vm.common -Force -DisableNameChecking
+
+$toolName = 'wireshark'
+$category = 'Networking Tools'
+
+$shortcutDir = Join-Path ${Env:TOOL_LIST_DIR} $category
+$shortcut = Join-Path $shortcutDir "Wireshark.lnk"
+$executablePath = Join-Path ${Env:ProgramFiles} "Wireshark\Wireshark.exe" -Resolve
+Install-ChocolateyShortcut -shortcutFilePath $shortcut -targetPath $executablePath -RunAsAdmin
```
MalwareMechanic, line 10: Can we wrap all this in a try/catch in case the `$executablePath` doesn't resolve we'll log it? Also, can you assert that the shortcut was actually created?

MalwareMechanic (changes requested):

MalwareMechanic (approved):

Issue #2, opened 2021-12-16
Allow to have a different default console


Ana06:
We are currently opening console applications using cmd.
```
$executableCmd  = Join-Path ${Env:WinDir} "system32\cmd.exe"
```
I think it would be nice to use a environment variable to allow to change the default console.

Ana06:
Same for the default browser. 🤔

MalwareMechanic:
> Same for the default browser. 🤔

Regarding my comment in https://github.com/mandiant/VM-Packages/pull/15#issuecomment-998445610, I was meaning to say we could have additional configuration packages to allow users to further configure Chrome. For instance, if they wanted the default browser to be Chrome then they could include that package. I think for Win10 you have to manually set your default browser anyway?

MalwareMechanic:
I realized the other tricky part is properly handling paths with spaces correctly. We have this down well for `cmd.exe`, we'd need to figure out for `powershell.exe` and `cmder.exe` the proper way to handle that.

Then we should probably create a function in `vm.common.psm1` to have a big switch statement to handle the bulk logic of:

```
$executableCmd  = Join-Path ${Env:WinDir} "system32\cmd.exe"
$executableDir  = Join-Path ${Env:UserProfile} "Desktop"
$executableArgs = "/K `"cd `"$executableDir`" && `"$executablePath`" --help`""
Install-ChocolateyShortcut -shortcutFilePath $shortcut -targetPath $executableCmd -Arguments $executableArgs -WorkingDirectory $executableDir -IconLocation $executablePath
```

so the users can simply pass in the parameters and not have to worry about handling paths with spaces for each console type.

Ana06:
Closing in favor of https://github.com/mandiant/VM-Packages/issues/971

PR #15, opened 2021-12-21
Add new package for CyberChef

MalwareMechanic:
Partially addresses: #3

### DIFF
--- a/packages/cyberchef.vm/cyberchef.vm.nuspec
+++ b/packages/cyberchef.vm/cyberchef.vm.nuspec
@@ -1,0 +1,13 @@
+<?xml version="1.0" encoding="utf-8"?>
+<package xmlns="http://schemas.microsoft.com/packaging/2015/06/nuspec.xsd">
+  <metadata>
+    <id>cyberchef.vm</id>
+    <version>9.32.3</version>
+    <authors>GCHQ</authors>
+    <description>The Cyber Swiss Army Knife - a web app for encryption, encoding, compression, data analysis, and more.</description>
+    <dependencies>
+      <dependency id="common.vm" />
+      <dependency id="googlechrome" />
+    </dependencies>
+  </metadata>
+</package>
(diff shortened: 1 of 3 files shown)


MalwareMechanic:
Will add additional config.chrome.* packages later. For instance:

- config.chrome.defaultbrowser.vm (to make default browser)
- config.chrome.removeshortcut.com (for desktop shortcut)

packages/cyberchef.vm/cyberchef.vm.nuspec
```diff
@@ -0,0 +1,13 @@
+<?xml version="1.0" encoding="utf-8"?>
+<package xmlns="http://schemas.microsoft.com/packaging/2015/06/nuspec.xsd">
+  <metadata>
+    <id>cyberchef.vm</id>
+    <version>9.32.3</version>
+    <authors>GCHQ</authors>
+    <description>The Cyber Swiss Army Knife - a web app for encryption, encoding, compression, data analysis, and more.</description>
+    <dependencies>
+      <dependency id="common.vm" />
+      <dependency id="googlechrome" />
```
Ana06, line 10: QUESTION: why are we using googlechrome vs firefox for example?

Ana06 (approved):
👍

packages/cyberchef.vm/cyberchef.vm.nuspec
Ana06: Added here: https://github.com/mandiant/VM-Packages/issues/2#issuecomment-998685033

MalwareMechanic: CyberChef supports:
- Google Chrome 50+
- Mozilla Firefox 38+

So I picked Chrome since it also has good debugging tools. Since it's a dependency we'd need to pick one or the other, but realistically we probably don't need both. Thoughts?

Ana06: The dependency makes it more difficult to make it configurable. It was just an idea, let's keep it like it is by now.

PR #17, opened 2021-12-21
Run test in different Windows versions & fix notepad & fix get-changed-files

Ana06:
We can only use windows-2016 till March 15, 2022. Ensure the packages work with other versions as well. I have also fixed notepad (see conversation in https://github.com/mandiant/VM-Packages/pull/8#pullrequestreview-836500590) in this PR to be able to test the windows versions. Otherwise there is nothing to build.

As the fix for notepad is done in a renamed file, I also needed to fix get-changed-files to include renamed files in `renamed`, `modified` and `added_modified`. I have released the fix in [Ana06/get-changed-files@v2.1.0](https://github.com/Ana06/get-changed-files/releases/tag/v2.1.0)


.github/workflows/ci.yml
```diff
@@ -14,7 +14,7 @@ jobs:
         uses: actions/checkout@v2
       - name: Get changed files
         id: files
-        uses: Ana06/get-changed-files@v2.0.0
+        uses: Ana06/get-changed-files@renamed
```
Ana06, line 17: I need to release v2.1.0 with this change.

.github/workflows/ci.yml
```diff
@@ -77,7 +81,7 @@ jobs:
           Exit 0
       - name: Push all built packages to MyGet
         # Only push packages on master if they were built (not if testing was skipped)
-        if: steps.test.outcome == 'success' && github.event_name == 'push'
+        if: steps.test.outcome == 'success' && github.event_name == 'push' && matrix.os == 'windows-2019'
```
MalwareMechanic, line 84: Is there a reason we're only looking for a specific version of Windows OS here?

Ana06: yes, to avoid pushing the packages 3 times

Ana06: should I add a comment to clarify it? Or do you have a better idea how to achieve this in a more beautiful way? 🤔

MalwareMechanic: Perhaps just a comment so I don't forget later 😄

MalwareMechanic: @Ana06 at'ing you so this doesn't get lost

MalwareMechanic (approved):

PR #19, opened 2021-12-22
Add procdump.vm

Ana06:
~Add processdump.vm. Note that it is called `procdump.exe` and not `pd32.exe` anymore.~

Closes https://github.com/mandiant/VM-Packages/issues/3

### DIFF
--- a/packages/common.vm/common.vm.nuspec
+++ b/packages/common.vm/common.vm.nuspec
@@ -2,7 +2,7 @@
 <package xmlns="http://schemas.microsoft.com/packaging/2015/06/nuspec.xsd">
   <metadata>
     <id>common.vm</id>
-    <version>0.0.0.20220222</version>
+    <version>0.0.0.20220223</version>
     <description>Common libraries for VM-packages</description>
     <authors>Mandiant</authors>
   </metadata>
--- a/packages/processdump.vm/tools/chocolateyinstall.ps1
+++ b/packages/processdump.vm/tools/chocolateyinstall.ps1
@@ -1,0 +1,43 @@
+$ErrorActionPreference = 'Stop'
+Import-Module vm.common -Force -DisableNameChecking
+
+try {
+  $toolName = "pd"
+  $category = "Utilities"
+
+  $toolDirParent = $(Split-Path -parent $MyInvocation.MyCommand.Definition)
+  $toolDir = Join-Path $toolDirParent $toolName
+  $shortcutDir = Join-Path ${Env:TOOL_LIST_DIR} $category
+  $zipUrl = "http://www.split-code.com/files/pd_v2_1.zip"
+  $zipSha256 = "2060f6713c082e78b2c18e6cb4e195602296af7ff9b096931f5f9f70a145b487"
+
+  # Remove files from previous zips for upgrade
+  VM-Remove-PreviousZipPackage ${Env:chocolateyPackageFolder}
+
+  $packageArgs = @{
+    packageName   = ${Env:ChocolateyPackageName}
+    unzipLocation = $toolDir
+    url           = $zipUrl
+    checksum       = $zipSha256
+    checksumType   = 'sha256'
+  }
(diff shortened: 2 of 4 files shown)


MalwareMechanic:
@Ana06 Seems you're using Sysinternal's "procdump" instead of: https://github.com/glmcdona/Process-Dump which has a download link on http://split-code.com/processdump.html. From the split-code website:

> Process Dump: Dump memory modules to disk
> by geoff mcdonald
>
> Process Dump is a Windows reverse-engineering tool to dump malware memory components back to disk for analysis. It uses an aggressive import reconstruction approach to make analysis easier, and supports 32 and 64 bit modules. Dumping of regions without PE headers is supported and in these cases PE headers and import tables will automatically be generated. Process Dump supports creation and use of a clean-hash database, so that dumping of clean files such as kernel32.dll can be skipped.

Process Dump comes in .zip format and supports Windows x86 and x64:
  -   Download: pd.exe v2.1 for Windows 32 and 64 bit
 
This tool depends on Microsoft Visual C++ 2015 Redistributable:
  -   Microsoft Visual C++ 2015 Redistributable Package

So it seems it does indeed need that dependency 🤮

Ana06:
It seems we have a problem:

Screenshot 2021-12-22 at 19 50 37

@mandiant/flare-vm  any ideas how to solve this?

MalwareMechanic:
> It seems we have a problem:
> 
> Screenshot 2021-12-22 at 19 50 37
> 
> @mandiant/flare-vm any ideas how to solve this?

Perhaps try disabling Windows Defender before attempting installation to see if that is the root cause. From https://theitbros.com/managing-windows-defender-using-powershell/, there are a few options that may disable it or uninstall it for Window Servers:

>If you want to disable Defender real-time protection:
>
>Set-MpPreference -DisableRealtimeMonitoring $true

>To completely disable Windows Defender on a computer, use the command:
>
>New-ItemProperty -Path “HKLM:\SOFTWARE\Policies\Microsoft\Windows Defender” -Name DisableAntiSpyware -Value 1 -PropertyType DWORD -Force

>You can uninstall Windows Defender on Windows Server 2016\2019 using PowerShell. Just run the following command in the elevated PowerShell session:
>
>Uninstall-WindowsFeature -Name Windows-Defender

packages/processdump.vm/tools/chocolateyinstall.ps1
```diff
@@ -0,0 +1,48 @@
+$ErrorActionPreference = 'Stop'
+Import-Module vm.common -Force -DisableNameChecking
+
+try {
+  # Disable Windows Defender
+  Set-MpPreference -DisableRealtimeMonitoring $true
+  New-Item -Path "HKLM:\SOFTWARE\Policies\Microsoft\Windows Defender" -Force
+  New-ItemProperty -Path "HKLM:\SOFTWARE\Policies\Microsoft\Windows Defender" -Name DisableAntiSpyware -Value 1 -PropertyType DWORD -Force
+  Uninstall-WindowsFeature -Name Windows-Defender
```
Ana06, line 9: Any of this options seems to work 😞

MalwareMechanic:
@Ana06 can you give this a quick look over to make sure I didn't miss anything?

packages/common.vm/common.vm.nuspec
```diff
@@ -2,7 +2,7 @@
 <package xmlns="http://schemas.microsoft.com/packaging/2015/06/nuspec.xsd">
   <metadata>
     <id>common.vm</id>
-    <version>3.0.1</version>
+    <version>0.0.0.20220113</version>
```
Ana06, line 5: I would prefer to merge https://github.com/mandiant/VM-Packages/pull/21 before this PR as it changes many files and to ensure this PR doesn't break the linter (if we can make that PR work 🤞). Then we should increase this (to today's date) to avoid conflicts with that PR (otherwise this PR won't push the package as it has the same version.

Ana06 (commented):
LGTM 👍

Issue #1, opened 2021-12-16
Update Python3 libraries


Ana06:
should we install the same packages in python 3 as in python 2? For example pefile and vivisect could be installed in Python 3 (maybe they didn't supported py3 when this file was created).

PR #23, opened 2022-02-21
Improve libraries.python packages and fix test suite 🎉 

Ana06:
Currently the test suite is broken because a version higher than Python 3.9 was already installed (GH actions include several versions of Python in the tools cache directory including Python 3.10.1) and stringsifter doesn't work in Python > 3.9 (as current last version of scikit-learn doesn't support Python 10). Ensure pip uses the correct Python version to fix this problem.

In addition, the following improvements are included in this PR:
- Do not install Python 3 in the Python 2 libraries package and vice versa
- Use library from pip when possible
- Use a fixed version when getting library from GitHub
- Sort Python libraries alphabetically and try to install same packages in
Python 2 and 3.
- Remove flask and networkx as they don't seem to be malware analysis related. Please speak up if you disagree.

Closes https://github.com/mandiant/VM-Packages/issues/1

### DIFF
--- a/packages/libraries.python2.vm/libraries.python2.vm.nuspec
+++ b/packages/libraries.python2.vm/libraries.python2.vm.nuspec
@@ -6,14 +6,12 @@
     <description>Metapackage that requires the dependencies below:
 - vcbuildtools.vm
 - python2
-- python3
 </description>
     <authors>Mandiant, Microsoft, Python Software Foundation</authors>
     <dependencies>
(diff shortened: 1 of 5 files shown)


Ana06:
This PR needs https://github.com/mandiant/VM-Packages/pull/24

Ana06:
Rebased after #24, CI is green! 🎉 Merging! 🚢

PR #31, opened 2022-02-25
Enhance package template script

mr-tz:
- adds template to aid creation of metapackes
- hopefully makes adding of new templates easier
- code format using isort and black (no more 2 space indents 🤣)

PR #32, opened 2022-02-25
add apktool metapackage

mr-tz:
generated using changes suggested in #31

relies on existing apktool package

### DIFF
--- a/packages/apktool.vm/apktool.vm.nuspec
+++ b/packages/apktool.vm/apktool.vm.nuspec
@@ -2,7 +2,7 @@
 <package xmlns="http://schemas.microsoft.com/packaging/2015/06/nuspec.xsd">
   <metadata>
     <id>apktool.vm</id>
-    <version>2.6.0.20220225</version>
+    <version>2.6.0</version>
     <authors>Connor Tumbleson, Ryszard Wisniewski</authors>
     <description>A tool for reverse engineering 3rd party, closed, binary Android apps.</description>
     <dependencies>
--- a/packages/apktool.vm/apktool.vm.nuspec
+++ b/packages/apktool.vm/apktool.vm.nuspec
@@ -1,0 +1,13 @@
+<?xml version="1.0" encoding="utf-8"?>
+<package xmlns="http://schemas.microsoft.com/packaging/2015/06/nuspec.xsd">
+  <metadata>
+    <id>apktool.vm</id>
+    <version>2.6.0.20220225</version>
+    <authors>Connor Tumbleson, Ryszard Wisniewski</authors>
+    <description>A tool for reverse engineering 3rd party, closed, binary Android apps.</description>
+    <dependencies>
+      <dependency id="common.vm" />
+      <dependency id="apktool" version="[2.6.0]" />
+    </dependencies>
+  </metadata>
+</package>
--- a/packages/apktool.vm/tools/chocolateyinstall.ps1
+++ b/packages/apktool.vm/tools/chocolateyinstall.ps1
@@ -1,0 +1,22 @@
+$ErrorActionPreference = 'Stop'
+Import-Module vm.common -Force -DisableNameChecking
+
+try {
+  $toolLnk = 'apktool.lnk'
+  $category = 'Android'
+  $shimPath = 'bin\apktool.exe'
+
+  $shortcutDir = Join-Path ${Env:TOOL_LIST_DIR} $category
+  $shortcut = Join-Path $shortcutDir $toolLnk
+
+  $executablePath = Join-Path ${Env:ChocolateyInstall} $shimPath -Resolve
+  $executableIcon = ""  # no icon
+  $executableCmd  = Join-Path ${Env:WinDir} "system32\cmd.exe"
+  $executableDir  = Join-Path ${Env:UserProfile} "Desktop"
+  $executableArgs = "/K `"cd `"$executableDir`" && `"$executablePath`""
+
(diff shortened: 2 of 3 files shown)


MalwareMechanic (approved):
Looks good to me! Haven't tested it locally, but all tests pass!

packages/apktool.vm/apktool.vm.nuspec
```diff
@@ -0,0 +1,13 @@
+<?xml version="1.0" encoding="utf-8"?>
+<package xmlns="http://schemas.microsoft.com/packaging/2015/06/nuspec.xsd">
+  <metadata>
+    <id>apktool.vm</id>
+    <version>2.6.0.20220225</version>
```
Ana06, line 5: Date shouldn't be included here as the forth segment is not used in the dependency. From the [documentation](https://github.com/mandiant/VM-Packages/wiki/Contributing#metapackages):
> When a metapackage is named after and installs/configures a specific tool (for instance cygwin.vm that installs/configures cygwin) lock the dependency version (e.g., version="[3.2.0]") and use the locked version as the metapackage's version. If the tool's version uses the 4th segment (e.g., version="[2.0.1.20210213]") update the metapackage's version to use the current date in the 4th segment (i.e., <version>2.0.1.20220224</version>).

Or is the documentation not clear enough? 🤔

packages/apktool.vm/apktool.vm.nuspec
```diff
@@ -0,0 +1,13 @@
...
+    <authors>Connor Tumbleson, Ryszard Wisniewski</authors>
```
Ana06, line 6: The dependecy only has Ryszard Wisniewski as author, not sure if we should follow that as I see in GitHub that Connor Tumbleson is the current maintainer.

packages/apktool.vm/tools/chocolateyinstall.ps1
```diff
@@ -0,0 +1,22 @@
+$ErrorActionPreference = 'Stop'
+Import-Module vm.common -Force -DisableNameChecking
+
+try {
+  $toolLnk = 'apktool.lnk'
```
Ana06, line 5: In most packages we define `$toolName` and `$shortcut = Join-Path $shortcutDir "$toolName.lnk"` (it is also [documented](https://github.com/mandiant/VM-Packages/wiki/Contributing#best-practices)).  I think keeping packages as similar as possible makes it easier to read and maintain.

packages/apktool.vm/tools/chocolateyinstall.ps1
```diff
@@ -0,0 +1,22 @@
...
+  $category = 'Android'
+  $shimPath = 'bin\apktool.exe'
+
+  $shortcutDir = Join-Path ${Env:TOOL_LIST_DIR} $category
+  $shortcut = Join-Path $shortcutDir $toolLnk
+
+  $executablePath = Join-Path ${Env:ChocolateyInstall} $shimPath -Resolve
```
Ana06, line 12: In other packages (for example [cygwin.vm](https://github.com/mandiant/VM-Packages/tree/ee3608a53dff436bb87cac531e93d1f2d5175e40/packages/cygwin.vm)), we ensure `$toolBin` exists doing this in two steps:
```
  $toolBin = Join-Path $toolDir "bin" -Resolve
  $executablePath = Join-Path $toolBin "$toolName.exe" -Resolve
```

I am not sure if there is a reason to do this in two steps instead of how @mr-tz has done it. @MalwareMechanic do you know?

Ana06 (commented):
🎉

packages/apktool.vm/tools/chocolateyinstall.ps1
```diff
@@ -0,0 +1,22 @@
...
+  $executableIcon = ""  # no icon
```
Ana06, line 13: You can remove this (and also from `Install-ChocolateyShortcut`)

packages/apktool.vm/apktool.vm.nuspec
mr-tz: leaving as is, although in the future I will copy straight from the dependency

packages/apktool.vm/tools/chocolateyinstall.ps1
mr-tz: thanks for the note, left this here as is

mr-tz (commented):
addressed feedback in bfeab5e

mr-tz:
good to go?

packages/apktool.vm/apktool.vm.nuspec
MalwareMechanic: It doesn't hurt to perform additional due diligence and add more authors 😄

packages/apktool.vm/apktool.vm.nuspec
MalwareMechanic: If we're going to wind up using the fourth segment anyway, perhaps we should update the documentation to say to always use the fourth segment as the current date? @Ana06

packages/apktool.vm/tools/chocolateyinstall.ps1
MalwareMechanic: I personally don't think it should matter, we were likely being over cautious before lol As long as the path is validated I don't think it matters if we do it one or two steps -- one step is shorter 😄  Now if we need to reuse the "bin" then it might make sense to have it in a separate variable, but that's not needed here.

PR #35, opened 2022-03-10
Remove windows-2016 from GH actions

Ana06:
The Windows 2016 runner image will be removed from GitHub-hosted runners
on March 15, 2022:
https://github.blog/changelog/2021-10-19-github-actions-the-windows-2016-runner-image-will-be-removed-from-github-hosted-runners-on-march-15-2022

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -8,7 +8,7 @@ on:
 
 jobs:
   lint:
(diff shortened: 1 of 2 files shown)


Ana06:
ouch! it is march 15 and the daily run is failing because of this. Let's merge this!

Issue #41, opened 2022-04-21
`notepadplusplus.vm` fails sometimes with `access is denied`


Ana06:
It seems that installing `notepadplusplus.vm` fails sometimes with `access is denied`: https://github.com/mandiant/VM-Packages/runs/6088818786?check_suite_focus=true It successes in some daily runs and fails in others.

I forgot to add the upload of logs to the daily in https://github.com/mandiant/VM-Packages/pull/22, I'll send a PR in a moment that may help us figuring out what the problem is.

mr-tz:
I don't think this happened for a longer time. Can we close this?

PR #22, opened 2022-02-18
Upload Chocolatey logs as artifacts

Ana06:
Upload the Chocolatey logs as they can be useful to debug errors.
Screen Shot 2022-02-21 at 18 44 58

PR #42, opened 2022-04-21
Upload Chocolatey logs in daily run

Ana06:
Upload the Chocolatey logs as artifacts in the daily run. I forgot this on https://github.com/mandiant/VM-Packages/pull/22.

This may help with https://github.com/mandiant/VM-Packages/issues/41.

I don't like the duplication in both actions, but I can't come up with a better way to do this. We could move the log paths to a file if we find ourselves modifying this often, but by now let's leave like it is.


Ana06:
It seems @MalwareMechanic is busy. I am going to merge this without review, as we have the same code in the PR run already.

PR #64, opened 2022-08-31
Feat/linter

### LINKED ISSUE #43: Add linter in Python
The Python script should check the following:
- [x] Ensure file paths use lowercase
- [x] Verify correct versioning: 3 first segments of the tool's version, fourth segment reserved for the date of hotfix
- [x] Verify if nuspec version is not increased (by geting both revisions)
- [x] Verify .nuspec has only the required fields
- [x] Verify the package depends on “common.vm” 
- [x] Verify .ps1 imports the “common.vm” 
- [x] Check if first line is “error action stop”

Previously documented in https://github.com/mandiant/VM-Packages/issues/6 (splitting that issue up now)

### LINKED ISSUE #44: Clear up Powershell linter
From @mr-tz comment in https://github.com/mandiant/VM-Packages/issues/6#issuecomment-1050131278:

> Make the linter output easy to see, currently lines like Progress: Downloading PSScriptAnalyzer 1.20.0... 0% add a lot of noise.

This refers to the [Powershell linter](https://github.com/mandiant/VM-Packages/blob/main/scripts/lint.ps1) and it is not related to the new [Python linter](https://github.com/mandiant/VM-Packages/issues/43).

mr-tz:
closes #43 
closes #44 

With the exception of "Verify if nuspec version is not increased (by getting both revisions)". Is this handled via the `check_nuspec` job in the CI workflow?!

### DIFF
--- a/packages/libraries.python2.vm/tools/chocolateyinstall.ps1
+++ b/packages/libraries.python2.vm/tools/chocolateyinstall.ps1
@@ -1,4 +1,4 @@
-$ErrorActionPreference = 'Continue'
+$ErrorActionPreference = 'Stop'
 Import-Module vm.common -Force -DisableNameChecking
 
 function LoadPackages {
--- a/packages/libraries.python2.vm/tools/chocolateyinstall.ps1
+++ b/packages/libraries.python2.vm/tools/chocolateyinstall.ps1
@@ -1,4 +1,4 @@
-$ErrorActionPreference = 'Stop'
+$ErrorActionPreference = 'Continue'
 Import-Module vm.common -Force -DisableNameChecking
 
 function LoadPackages {
--- a/packages/ollydbg.vm/ollydbg.vm.nuspec
+++ b/packages/ollydbg.vm/ollydbg.vm.nuspec
@@ -2,7 +2,7 @@
 <package xmlns="http://schemas.microsoft.com/packaging/2015/06/nuspec.xsd">
   <metadata>
     <id>ollydbg.vm</id>
-    <version>1.11</version>
+    <version>1.10.0.20220908</version>
     <description>OllyDbg is a 32-bit assembler level analysing debugger for Windows. Emphasis on binary code analysis makes it particularly useful in cases where source is unavailable.</description>
     <authors>Oleh Yuschuk</authors>
     <dependencies>
--- a/packages/ollydbg.vm/ollydbg.vm.nuspec
+++ b/packages/ollydbg.vm/ollydbg.vm.nuspec
@@ -2,7 +2,7 @@
 <package xmlns="http://schemas.microsoft.com/packaging/2015/06/nuspec.xsd">
   <metadata>
     <id>ollydbg.vm</id>
-    <version>1.10</version>
+    <version>1.11</version>
     <description>OllyDbg is a 32-bit assembler level analysing debugger for Windows. Emphasis on binary code analysis makes it particularly useful in cases where source is unavailable.</description>
     <authors>Oleh Yuschuk</authors>
     <dependencies>
--- a/packages/ollydbg.vm/ollydbg.vm.nuspec
+++ b/packages/ollydbg.vm/ollydbg.vm.nuspec
@@ -3,7 +3,6 @@
   <metadata>
     <id>ollydbg.vm</id>
     <version>1.10</version>
-    <tags>reverse engineering debugger disassembler binary analysis</tags>
     <description>OllyDbg is a 32-bit assembler level analysing debugger for Windows. Emphasis on binary code analysis makes it particularly useful in cases where source is unavailable.</description>
     <authors>Oleh Yuschuk</authors>
     <dependencies>
--- a/scripts/test/lint.py
+++ b/scripts/test/lint.py
@@ -276,6 +276,7 @@ def lint(path) -> Dict[str, list]:
             violations = list()
 
             path = pathlib.Path(os.path.join(root, name))
+            logger.info("linting %s", path)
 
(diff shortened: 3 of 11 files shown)


packages/libraries.python2.vm/tools/chocolateyinstall.ps1
```diff
@@ -1,4 +1,4 @@
-$ErrorActionPreference = 'Continue'
+$ErrorActionPreference = 'Stop'
```
mr-tz, line 1: I assume that was here for a reason? Should I exclude this and the Py3 library package from this check?

Ana06: I think this is because we install several python libraries with this package and we want to install as many as possible even if one fails (as they are independent). So, yes, exclude them from the check.

Ana06:
> With the exception of "Verify if nuspec version is not increased (by getting both revisions)". Is this handled via the check_nuspec job in the CI workflow?!

No, the workflow just check that the nuspec is modified. It was a quick way to avoid we forget to update the version. But we only check if the file is modified, not if the version was increased. If this is difficult to do with this linter, we could maybe add it easier to the current workflow. 🤔

Ana06 (commented):
The script looks great, thanks @mr-tz! I would personally appreciate some more detailed commit messages, but it is not a blocker.

packages/libraries.python2.vm/tools/chocolateyinstall.ps1
mr-tz: baae48a

mr-tz:
I've added a lint for the version update. Changed files are obtained via git.

packages/ollydbg.vm/ollydbg.vm.nuspec
```diff
@@ -3,7 +3,6 @@
   <metadata>
     <id>ollydbg.vm</id>
```
mr-tz, line 4: Does an update of only the `.nuspec` file require a version update? If so, I need to modify the linter.

Ana06: Yes. Imagine we update the authors because they are wrong. Then the updated package won't be pushed if the version is not updated.

The previous CI check is a quick fix and is not covering this case.

mr-tz: updated in 26653af

packages/ollydbg.vm/ollydbg.vm.nuspec
```diff
@@ -2,8 +2,7 @@
 <package xmlns="http://schemas.microsoft.com/packaging/2015/06/nuspec.xsd">
   <metadata>
     <id>ollydbg.vm</id>
-    <version>1.10</version>
-    <tags>reverse engineering debugger disassembler binary analysis</tags>
+    <version>1.11</version>
```
mr-tz, line 5: is this the proper version update?

Ana06: no, you are not updating the tool, so you shouldn't update the first 3 segments as they are the tool version.

From our [documentation](https://github.com/mandiant/VM-Packages/wiki/Contributing#version):

> When do I update the a package's version string?
> 
> The version needs to be increased every time you modify an existing package. For example, if you fix a bug in a package, you need to update the nuspec version string. Our guidance is to add the current date to the 4th segment using the format YYYMMDD. For example, 1.6.2 becomes 1.6.2.20220113 and 3.02 becomes 3.02.0.20220113. If the date is already in the 4th segment, you can increment it to the next day (e.g., 20220113 -> 20220114) and when the PR is merged it should be good to go.

maybe the linter can catch this 🤔

scripts/test/lint.py
```diff
@@ -0,0 +1,327 @@
...
+)
+
+
+def lint_install(path):
+    return run_lints(INSTALL_LINTS, path)
+
+
+def lint(path) -> Dict[str, list]:
+    ret = {}
+    for root, dirs, files in os.walk(path):
```
Ana06, line 270: what about directories without files? Do we need to ensure they are also lowercase @MalwareMechanic?

scripts/test/lint.py
```diff
@@ -0,0 +1,327 @@
...
+
+        version = metadata.getElementsByTagName("version")[0]
+        version = version.firstChild.data.split(".")
+        if len(version) > 4:
+            print(f"{path} more than four version segments: {version}")
+            return True
+
+        if len(version) == 4:
+            try:
+                datetime.datetime.strptime(version[3], "%Y%m%d")
```
Ana06, line 121: we could also check that segments 1, 2 and 3 are a number

packages/ollydbg.vm/ollydbg.vm.nuspec
mr-tz: 61f4a87
I don't think the linter can automatically catch this here.

scripts/test/lint.py
mr-tz: 667015e

scripts/test/lint.py
mr-tz: 100b7d6

mr-tz: added, if we have to adjust later for `rc` or `beta` or whatever, it's an easy fix

Ana06 (approved):
This looks great, thanks for all the work @mr-tz!!! 💘

 If you don't want to squash the commits, I propose we merge this PR with `squash and merge` as there are several `cleanup` and `fix` commits.

mr-tz:
Thanks for all the helpful feedback!

Issue #67, opened 2022-09-12
CI: send PR fails with "line 16: unexpected EOF while looking for matching `"'"


mr-tz:
### What's the problem?

See https://github.com/mandiant/VM-Packages/runs/8303111845?check_suite_focus=true#step:5:20

May be an issue with the line breaks?

### Steps to Reproduce

add `send PR` label to METAPACKAGE issue

### Environment

CI

### Detected by test suite

No

Ana06:
The problem is the `'` in the description as we are using `'` to escape the json string. I have manually changed `'` by `''` in https://github.com/mandiant/VM-Packages/issues/66 and then the action work.

PR #69, opened 2022-09-12
ci: update pkg from json

mr-tz:
closes #67

PR #71, opened 2022-09-12
Do no create file in new_package workflow

Ana06:
Restoring change introduced in https://github.com/mandiant/VM-Packages/pull/69 as the generated file gets added to the commit and this file is not needed. The code also doesn't fix https://github.com/mandiant/VM-Packages/issues/67, see https://github.com/mandiant/VM-Packages/issues/67#issuecomment-1243838565.

### DIFF
--- a/.github/workflows/new_package.yml
+++ b/.github/workflows/new_package.yml
@@ -25,8 +25,7 @@ jobs:
           # The Python script prints the package name so that we can easily
           # echo it to $GITHUB_ENV. Alternatively we could parse if with a
           # regular string from the JSON string
-          echo '${{ steps.issue-parser.outputs.jsonString }}' > pkg.json
-          pkg_name=$(./scripts/utils/create_package_template_from_json.py --json_file pkg.json)
+          pkg_name=$(./scripts/utils/create_package_template_from_json.py '${{ steps.issue-parser.outputs.jsonString }}')
           echo "pkg_name=$pkg_name" >> $GITHUB_ENV
       - name: Create Pull Request
         uses: peter-evans/create-pull-request@v4


Ana06:
@mr-tz not sure if we should delete the option to read from file you added to the Python script in https://github.com/mandiant/VM-Packages/pull/69.

I am also not sure what the best way to fix https://github.com/mandiant/VM-Packages/issues/67 is. We could delete the file after using it if that helps us somehow. 🤔

.github/workflows/new_package.yml
```diff
@@ -25,8 +25,7 @@ jobs:
           # The Python script prints the package name so that we can easily
           # echo it to $GITHUB_ENV. Alternatively we could parse if with a
           # regular string from the JSON string
-          echo '${{ steps.issue-parser.outputs.jsonString }}' > pkg.json
-          pkg_name=$(./scripts/utils/create_package_template_from_json.py --json_file pkg.json)
+          pkg_name=$(./scripts/utils/create_package_template_from_json.py '${{ steps.issue-parser.outputs.jsonString }}')
```
mr-tz, line 28:
```suggestion
          pkg_name=$(./scripts/utils/create_package_template_from_json.py --json_string '${{ steps.issue-parser.outputs.jsonString }}')
```

mr-tz: We should escape the JSON chars, maybe via `jq`? See https://stackoverflow.com/questions/10053678/escaping-characters-in-bash-for-json

mr-tz: Alternatively, deleting the created file should work?!

Ana06: yes, we could delete the file after using it.

Ana06: I think `jq` is not installed. Maybe using https://github.com/sergeysova/jq-action

Ana06:
Fixed in https://github.com/mandiant/VM-Packages/pull/73

PR #72, opened 2022-09-13
fix: delete json file

mr-tz:
alternative to #71 deleting the file after usage

### DIFF
--- a/.github/workflows/new_package.yml
+++ b/.github/workflows/new_package.yml
@@ -32,7 +32,7 @@ jobs:
           # The Python script prints the package name so that we can easily
           # echo it to $GITHUB_ENV. Alternatively we could parse if with a
           # regular string from the JSON string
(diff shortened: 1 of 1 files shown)


MalwareMechanic (approved):
Seems short and easy, I like it

mr-tz:
Bummer we cannot really test it.

mr-tz:
@Ana06 should we go with this (and test in production 😄)?

Ana06 (commented):
Note this doesn't fix https://github.com/mandiant/VM-Packages/issues/67, we still need to manually escape `'`. So it is basically the same as https://github.com/mandiant/VM-Packages/pull/71. A fix for #67 would be the ideal solution, but if we don't have time for it now, we should merge #71 or #72 (no preference, they are the same).

mr-tz:
superseded by #73

PR #73, opened 2022-09-14
Fix PR from JSON with special characters

### LINKED ISSUE #67: CI: send PR fails with "line 16: unexpected EOF while looking for matching `"'"

mr-tz:
remove tmp json file after use and closes #67 
see example success run at https://github.com/mr-tz/VM-Packages/actions/runs/3052783651/jobs/4922573479

I think the issue was in the `system` call. There may still be a security issue here potentially running user input directly, but it's mitigated via the ENV usage and passing the arguments to the script now.

### DIFF
--- a/.github/workflows/new_package.yml
+++ b/.github/workflows/new_package.yml
@@ -27,6 +27,7 @@ jobs:
           # regular string from the JSON string
           echo '${{ steps.issue-parser.outputs.jsonString }}' > pkg.json
           pkg_name=$(./scripts/utils/create_package_template_from_json.py --json_file pkg.json)
+          rm pkg.json
           echo "pkg_name=$pkg_name" >> $GITHUB_ENV
       - name: Create Pull Request
         uses: peter-evans/create-pull-request@v4
--- a/.github/workflows/new_package.yml
+++ b/.github/workflows/new_package.yml
@@ -20,13 +20,19 @@ jobs:
         id: issue-parser
         with:
           template-path: .github/ISSUE_TEMPLATE/new_package.yml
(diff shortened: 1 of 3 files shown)


Ana06 (commented):
> see example success run at https://github.com/mr-tz/VM-Packages/actions/runs/3052783651/jobs/4922573479

The run actually fails, but it seems it is because it doesn't have permissions to create the PR, #67 appears to be fixed 🎉 thanks for testing it @mr-tz! 💘 🎉 

---

> I think the issue was in the system call. There may still be a security issue here potentially running user input directly, but it's mitigated via the ENV usage and passing the arguments to the script now.

I don't know what you mean. Can you please elaborate?

---
Screen Shot 2022-09-14 at 18 21 39
@mr-tz is it ok for you if we merge this with `squash and merge`? Otherwise can you please rewrite this commit message? It seems you kept some old commit messages while rebasing.

mr-tz:
There may have been potential for command injections (which the error with the `'` also indicated).
Can you review and squash and merge, please?

Ana06:
> There may have been potential for command injections (which the error with the ' also indicated).

Right, good thing to take into account! Note that only people with permissions to add labels to issues can trigger this action.

mr-tz:
Thank you!

> Note that only people with permissions to add labels to issues can trigger this action.

Yes, that's good.

PR #74, opened 2022-09-15
fix: only add packages directory

mr-tz:


.github/workflows/new_package.yml
```diff
@@ -40,6 +40,8 @@ jobs:
         with:
           title: ':robot: Add ${{env.pkg_name}}.vm'
           body: 'Automated PR from ${{github.event.issue.html_url}} :rocket:'
+          add-paths:
```
Ana06, line 43: As there is only one path, I would put it in the same line

.github/workflows/new_package.yml
```diff
@@ -40,6 +40,8 @@ jobs:
...
+            packages/*
```
mr-tz, line 44:
```suggestion
          add-paths: packages/*
```

Ana06 (approved):

Ana06:
wait, we should also remove the 
```
# remove tmp file
rm pkg.json
```

That's not needed anymore. Sorry didn't realised the first time I checked it.

mr-tz:
Good point, handled in [361c1f1](https://github.com/mandiant/VM-Packages/pull/74/commits/361c1f18009a46eb38efac9ef14eb9291cda7a2d)

Ana06 (approved):
thanks!

PR #78, opened 2022-09-16
Fix daily run

mr-tz:
Fixes errors like
```
Cannot find path 'D:\a\VM-Packages\VM-Packages\packages\@{Name=010editor.vm}' because it does not
     | exist.
```


mr-tz:
See success in https://github.com/mandiant/VM-Packages/actions/runs/3066685911/jobs/4952138591

Ana06 (approved):
Thanks Moritz!

PR #83, opened 2022-10-21
Automatic package updates

Ana06:
Add workflow that runs `scripts/utils/update_package.py` for every package. `scripts/utils/update_package.py` identifies packages downloaded from a GitHub release URL, queries the latest version and updates:
- The version in `tools/chocolateyinstall.ps1`, including the download URL.
- The SHA256 hashes in `tools/chocolateyinstall.ps1`
- The version in the nuspec

The workflow is run twice a week and creates a PR with a commit for every package.

See example run here: https://github.com/Ana06/VM-Packages/actions/runs/3297886819
See PR example here: https://github.com/Ana06/VM-Packages/pull/17

Screen Shot 2022-10-21 at 15 50 05

The workflow tests changes before committing. The workflow logs packages that failed to update, but the action still succeeds and sends a PR.
Screen Shot 2022-10-21 at 15 49 37

The workflow uploads the output of testing every package to the assets.
Screen Shot 2022-10-21 at 15 51 06

### DIFF
--- a/.github/workflows/update_package.yml
+++ b/.github/workflows/update_package.yml
@@ -1,0 +1,63 @@
+name: Update package
+
+on:
+  schedule:
+    # Runs Wednesdays and Sundays at a weird time (22:21 UTC) to avoid
+    # delays during periods of high loads of GitHub Actions workflow runs.
+    - cron: '21 22 * * 0,3'
+
+jobs:
+  update:
+    runs-on: windows-2022
+    steps:
+      - name: Checkout code
+        uses: actions/checkout@v3
+      - name: Install python dependency 'requests'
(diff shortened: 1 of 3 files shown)


mr-tz (approved):
LGTM! For future easier maintainance it may be helpful to run the script thorugh black/isort and comment the regexes a bit.

Ana06:
@mr-tz 
> to run the script thorugh black/isort 
Do you mean locally or setting up CI? I have run `black -l 120` locally and the imports are sorted alphabetically (which I think it is what isort does). 

> comment the regexes a bit.

The regexes are a bit ugly, good idea! 👍

PR #59, opened 2022-07-08
Add common functions

day1player:
Added "VM-Install-Single-Exe" and "VM-Install-Single-Ps1" functions to common module. Also created templates for each of these types of packages in the python helper script.

closes #93

Issue #38, opened 2022-04-21
Categories


Ana06:
We are very close to start using these packages in FLARE-VM. We have mentioned a few times that we want to give a thought to the current categories, trying to create more concrete categories that make it easier to find tools. 

I think these are the current categories:
- Android
- Debuggers
- Delphi
- Developer Tools
- Disassemblers
- dotNet
- Flash
- Forensic
- Hex Editors
- Java
- Javascript
- Net
- Office
- PDF
- Pentest
- PowerShell
- Python
- Text Editors
- Utilities
- VB
- Web Application

This is also needed for https://github.com/mandiant/VM-Packages/issues/25

@mandiant/flare-vm ideas?

mr-tz:
Some ideas:
1. move `Flash` tools into `Utilities`, I don't think we really need Flash tools anymore, anyway
2. rename `Net` to `Network`/`Networking`? I always have to look twice with `dotNet` and `Net`
3. add `Dynamic` analysis with things like procmon, apimonitor, etc.?
4. add `Packers` with `upx`, `DiE`, etc.?
5. add `PE Viewers`?

3.-5. would help to split up `Utilities` more, but I'm not sure there's real benefit to having more or less categories.

MalwareMechanic:
Update packages based on category additions/renames from: https://github.com/mandiant/VM-Packages/pull/59

mr-tz:
My proposal:
- Android
- Debuggers
- Delphi
- **Developer Tools -> move to Utilities**
- Disassemblers
- dotNet
- **Flash -> move to Utilities**
- Forensic
- Hex Editors
- Java
- Javascript
- **Net -> rename to Networking**
- Office
- PDF
- **Pentest -> remove and see below?**
- PowerShell
- Python
- Text Editors
- Utilities
- VB
- Web Application

And from #59 (Commando), add:
- Active Directory
- Command & Control
- Evasion
- Exploitation
- Information Gathering
- Password Attacks
- Vulnerability Analysis
- Wordlists

mr-tz:
However, I also believe that it should be "allowed" and easy to change/rename/move categories as we cannot know now what tools we'll have and how to organize them best.

Should we create a separate file to track all current categories (and document what they roughly contain)?

mr-tz:
- Android
- Debuggers
- Delphi
- Disassemblers
- dotNet
- Forensic
- Hex Editors
- Java
- Javascript
- Networking
- Office
- PDF
- PE
[...]

PR #94, opened 2022-11-14
Feat/new package url exe

mr-tz:
merge after #59


scripts/utils/create_package_template.py
```diff
@@ -27,7 +27,7 @@
     "Hex Editors",
     "Java",
     "Javascript",
-    "Net",
+    "Networking",
```
Ana06, line 30: are these the final categories? They are different from https://github.com/mandiant/VM-Packages/issues/38#issuecomment-1313874710

Note that modifying the categories also needs to update our current GH actions and the packages that are using the old categories (and maybe somewhere else?).

Ana06 (changes requested):
The review was related to #59, didn't realized before this included #59. I have added the review there.

PR #98, opened 2022-11-16
Update dependencies automatically

Ana06:
Update dependencies in "Update package" workflow. See it working here:
https://github.com/Ana06/VM-Packages/pull/21
Allow to also trigger the workflow manually. The workflow still runs automatically on Wednesdays.

Note that the script only update fixed dependencies and many of our dependencies are not fixed. The script relies on metapackages having just one dependency with same name and version (only first 3 segments). I have renamed notepad++ to follow this. **We should document this. Should we rename/adapt other packages?**


mr-tz (approved):

mr-tz:
The naming and version fixing should be documented and optionally checked via the linter.

PR #104, opened 2022-11-18
Test all packages in daily

Ana06:
Test all packages in daily so that we now all packages that fail. 😄 

Reduce the chocolatey output with the `-r` option to make it easier to read the now longer output. Also make it easier to download the log with failing packages.

Unrelated to CI: Do not ask for confirmation to build packages. It annoyed me during manual testing. :wink:

### DIFF
--- a/.github/workflows/daily.yml
+++ b/.github/workflows/daily.yml
@@ -18,7 +18,7 @@ jobs:
         uses: actions/checkout@v3
       - name: Build and test all modified packages
         id: test
-        run: scripts/test/test_install.ps1
+        run: scripts/test/test_install.ps1 -all
       - name: Upload chocolatey logs to artifacts
         uses: actions/upload-artifact@v3
         if: always()
--- a/scripts/test/test_install.ps1
+++ b/scripts/test/test_install.ps1
@@ -1,11 +1,12 @@
 # Build the packages in the 'packages' directory given as argument (or all if none provided) into the 'built_pkgs'.
-# Install the built packages
+# Install the built packages. If a package install fails and the $all switch is not provided,
+# the rest of the packages are not installed
 
 # Examples
 ## ./test_install
 ## ./test_install '7zip.vm 010editor.vm'
 
-param ([string] $package_names=$null)
+param ([string] $package_names=$null, [switch] $all)
 
 # Error Code Definitions
 # ----------------------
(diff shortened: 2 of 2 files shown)


.github/workflows/daily.yml
```diff
@@ -18,7 +18,13 @@ jobs:
         uses: actions/checkout@v3
       - name: Build and test all modified packages
         id: test
-        run: scripts/test/test_install.ps1
+        run: scripts/test/test_install.ps1 -all
+      - name: Upload log to artifacts
+        uses: actions/upload-artifact@v3
+        if: always()
+        with:
+          name: failed_packages.txt
```
MalwareMechanic, line 26: I don't think this file will ever exist since we're not testing `flarevm.installer.vm`

scripts/test/test_install.ps1
```diff
@@ -40,15 +41,14 @@ $exclude_tests = @("flarevm.installer.vm")
...
-    choco install $package -y -s "'.;https://www.myget.org/F/vm-packages/api/v2;https://community.chocolatey.org/api/v2/'" --no-progress --force
-    if ($validExitCodes -notcontains $LASTEXITCODE) { # Abort with the first failing install
-        Write-Host -ForegroundColor Red "[ERROR] Failed to install $package"
-        # Restore the original location
-        Set-Location -Path $org_dir -PassThru | Out-Null
-        Exit 1
+    choco install $package -y -r -s "'.;https://www.myget.org/F/vm-packages/api/v2;https://community.chocolatey.org/api/v2/'" --no-progress --force
+    if ($validExitCodes -notcontains $LASTEXITCODE) {
+        $failed = $true
+        if (-not $all) { break } # Abort with the first failing install
```
MalwareMechanic, line 47:
```suggestion
        if (-not $all.IsPresent) { break } # Abort with the first failing install
```

.github/workflows/daily.yml
MalwareMechanic: We may want to manually create this file like:

```
$failedPackages = "failed_packages.txt"

..[SNIP]...

if ($validExitCodes -notcontains $LASTEXITCODE) {
    Add-Content $failedPackages $package

...[SNIP]...
```

MalwareMechanic (changes requested):

.github/workflows/daily.yml
Ana06: it exists, we were already uploading it. Check the daily uploads: https://github.com/mandiant/VM-Packages/actions/runs/3467207358

scripts/test/test_install.ps1
Ana06: It works without the `IsPresent`. I see on the internet that people use `IsPresent` `-eq $true` and nothing, is there a reason we prefer one of this options?

MalwareMechanic: To me it reads better to have `<switch_variable_name>.IsPresent`

mr-tz (approved):

PR #106, opened 2022-11-20
Update category from Utils to Utilities

MalwareMechanic:


mr-tz (approved):
Good catch. Should we add a lint for the categories?

Unrelated(?) test fail:
```
Installing the following packages:
True
```

PR #109, opened 2022-11-21
Fix test exclusions array

MalwareMechanic:


Ana06 (approved):
Thanks for fixing this @MalwareMechanic! 👍

PR #112, opened 2022-11-22
Fix host survey, AV portion

MalwareMechanic:
Seems to work sometimes, added a "TODO" to address later


mr-tz (approved):

mr-tz (commented):
This works for me now. I get:
```
VM AV Details
-----
AntiVirusProduct classname does not exist...

```

PR #116, opened 2022-11-25
Fix dependency for notepadpp plugin

MalwareMechanic:


MalwareMechanic:
Somehow `notepadplusplus.vm` is missing from the package repository

mr-tz (approved):

mr-tz:
Updated missing package via #114

PR #120, opened 2022-11-29
Update libraries.python* packages to use XML

MalwareMechanic:
- Updated libraries.python*.vm to use XML instead of JSON.
- Investigated the error "py -2 not available". Turns out Python3 installs "Python Launcher for Windows" (see: https://docs.python.org/3/using/windows.html#python-launcher-for-windows). We previously had the `libraries.python2.vm` package depend on Python3 -- this was the reason. I recommend we stick with that, even though it's slightly counterintuitive for simplicity. Added a comment in `libraries.python2.vm`'s nuspec file to document this reasoning.
- Removed `stringsifter` from `libraries.python3.vm`'s `modules.xml` file since I couldn't get it to build locally. Until we figure out how, or the stringsifter folks update their installer, I'm in favor of skipping for now.


MalwareMechanic:
Linter says: `FAIL: packages\python2.vm\python2.vm.nuspec - the version has not been updated: update the version in the nuspec file` but I didn't update this file?

mr-tz:
@MalwareMechanic rebasing this should fix the linter fail

PR #123, opened 2022-11-29
Update notepadplusplus to wait on config file

MalwareMechanic:
Also updated the spacing from two to four spaces

### DIFF
--- a/packages/notepadplusplus.vm/tools/chocolateyinstall.ps1
+++ b/packages/notepadplusplus.vm/tools/chocolateyinstall.ps1
@@ -28,7 +28,7 @@ try {
     # Wait for config.xml to exist after notepad++ exits
     $count = 0
     while (-Not (Test-Path -Path $configFile)) {
-        Start-Sleep 1
+        Start-Sleep -Seconds 1
         if ($count++ -eq 10) {
           break
         }
--- a/packages/notepadplusplus.vm/tools/chocolateyinstall.ps1
+++ b/packages/notepadplusplus.vm/tools/chocolateyinstall.ps1
@@ -2,31 +2,45 @@ $ErrorActionPreference = 'Stop'
 Import-Module vm.common -Force -DisableNameChecking
 
 try {
-  # Path to Notepad++'s configuration file
-  $configFile = Join-Path ${Env:APPDATA} "Notepad++\config.xml"
+    # Path to Notepad++'s configuration file
+    $configFile = Join-Path ${Env:APPDATA} "Notepad++\config.xml"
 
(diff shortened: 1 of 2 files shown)


MalwareMechanic:
@mr-tz linter is saying the version needs to match the fixed version, but when we update a package we always need to increase the version. I propose we update the guidance since it's not clear for metapackages from:

> When a metapackage is named after and installs/configures a specific tool (for instance cygwin.vm that installs/configures cygwin) lock the dependency version (e.g., version="[3.2.0]") and use the locked version as the metapackage's version. If the tool's version uses the 4th segment (e.g., version="[2.0.1.20210213]") update the metapackage's version to use the current date in the 4th segment (i.e., <version>2.0.1.20220224</version>).


to:
> When a metapackage is named after and installs/configures a specific tool (for instance cygwin.vm that installs/configures cygwin) lock the dependency version (e.g., version="[3.2.0]") and use the locked version as the metapackage's version. If the tool's version uses the 4th segment (e.g., version="[2.0.1.20210213]") update the metapackage's version to use the current date (YYYYMMDD) in the 4th segment (i.e., <version>2.0.1.20220224</version>). Additionally, when a change is made to a metapackge, always use the current date in the 4th segment.

Main change is the line:
**Additionally, when a change is made to a metapackge, always use the current date in the 4th segment.**

packages/notepadplusplus.vm/tools/chocolateyinstall.ps1
```diff
@@ -2,31 +2,45 @@ $ErrorActionPreference = 'Stop'
...
-  }
 
-  if (-Not (Test-Path -Path $configFile)) {
-    VM-Write-Log "WARN" "Can't find config.xml to modify"
-  } else {
-    # Update the config file and disable auto-updates
-    (Get-Content -Path $configFile) -Replace '("noUpdate".*?">)no', '$1yes' | Set-Content -Path $configFile -Force | Out-Null
-  }
+    if (-Not (Test-Path -Path $configFile)) {
+        VM-Write-Log "WARN" "Can't find Notepad++'s config.xml to disable auto-updates, skipping..."
```
mr-tz, line 38: ah, that confused me before (now I know what you meant :D)

mr-tz (approved):
pending the linter fix

packages/notepadplusplus.vm/tools/chocolateyinstall.ps1
mr-tz: good change here

mr-tz:
updating the linter now

mr-tz:
linter update addressed in #124

PR #117, opened 2022-11-28
Display Failures in daily run

mr-tz:
Example output:

```cmd
$ Select-String "Failures" -Context 0,1 .\chocolatey.log
> chocolatey.log:64081:2022-11-28 03:53:17,123 2916 [ERROR] - Failures
  chocolatey.log:64082:2022-11-28 03:53:17,123 2916 [ERROR] -  - notepadpp.plugin.compare.vm (exited 1) -
notepadpp.plugin.compare.vm not installed. An error occurred during installation:
> chocolatey.log:69998:2022-11-28 03:56:38,391 4216 [ERROR] - Failures
  chocolatey.log:69999:2022-11-28 03:56:38,391 4216 [ERROR] -  - python2.vm (exited -1) - Error while running
'C:\ProgramData\chocolatey\lib\python2.vm\tools\chocolateyinstall.ps1'.
> chocolatey.log:78824:2022-11-28 03:59:09,324 6592 [ERROR] - Failures
  chocolatey.log:78825:2022-11-28 03:59:09,324 6592 [ERROR] -  - wireshark (exited -1) - Error while running
'C:\ProgramData\chocolatey\lib\wireshark\tools\chocolateyInstall.ps1'.
```

PR #125, opened 2022-11-29
Add additional log sources to view after install

MalwareMechanic:


mr-tz (approved):

packages/flarevm.installer.vm/tools/chocolateyinstall.ps1
```diff
@@ -76,7 +76,10 @@ try {
     # Display installer log if available
     $logPath = Join-Path ${Env:VM_COMMON_DIR} "log.txt"
     if ((Test-Path $logPath)) {
-        Write-Host "[-] Please see installer log for any errors: $logPath" -ForegroundColor Yellow
+        Write-Host "[-] Please check the following logs for any errors:" -ForegroundColor Yellow
```
Ana06, line 79: This is written to `C:\ProgramData\_VM\log.txt`, right? I would add all the logs we mention there to the artifacts of the actions and display `C:\ProgramData\_VM\log.txt` in the action log. Related to https://github.com/mandiant/VM-Packages/pull/117

MalwareMechanic: This message is being displayed to the user in the PowerShell Window. We could also add this message to the `log.txt` file.

MalwareMechanic: @Ana06  Also add the following logs to the artifacts: `%PROGRAMDATA%\chocolatey\lib-bad\*\install_log.txt`

Packages may create a per-install log to use for whatever they wish (see: https://github.com/mandiant/VM-Packages/blob/f27bacad8c5e658fec4434a5f1f57122ebcb7d94/packages/common.vm/tools/vm.common/vm.common.psm1#L169). The packages `libraries.python*.vm` use this to redirect output from pip when installing Python modules. It'd be good to grab these per-install log files as well. You'll likely need to use a regex to find them, but they should be in a subdirectory within `lib-bad` somewhere.

MalwareMechanic: Documented in https://github.com/mandiant/VM-Packages/issues/133

PR #131, opened 2022-11-30
Remove python2.vm

MalwareMechanic:


mr-tz (approved):

mr-tz:
Should we set a reminder to manually delete removed packages from the feed?

MalwareMechanic:
We should have a GitHub Action that compares packages in the GitHub repository to those hosted on MyGet and let us know if they differ. I think that's the best solution.

MalwareMechanic:
Documented in https://github.com/mandiant/VM-Packages/issues/132

PR #141, opened 2022-12-01
Update installer log and add complete notice

MalwareMechanic:


packages/flarevm.installer.vm/tools/chocolateyinstall.ps1
```diff
@@ -86,6 +89,40 @@ try {
         Start-Sleep 5
         & notepad.exe $logPath
     }
+
+    # Let users know installation is complete by setting background, playing win sound, and display message box
+    Set-ItemProperty 'HKCU:\Control Panel\Colors' -Name Background -Value "0 0 0" -Force | Out-Null
+    $backgroundImage = "${Env:VM_COMMON_DIR}\background.png"
```
mr-tz, line 95: how's this file put there?

packages/flarevm.installer.vm/tools/chocolateyinstall.ps1
```diff
@@ -86,6 +89,40 @@ try {
         Start-Sleep 5
         & notepad.exe $logPath
     }
+
+    # Let users know installation is complete by setting background, playing win sound, and display message box
```
mr-tz, line 93: going above and beyond here 🤣

mr-tz (approved):
looks good, I need to see this in action :)

packages/flarevm.installer.vm/tools/chocolateyinstall.ps1
MalwareMechanic: Currently in the `installer.ps1` see:
 https://github.com/mandiant/flare-vm/blob/4923163b9ec97e3e863aa714e156e5b5784d8713/install.ps1#L846

This was done so that the `flarevm.installer.vm` can remain "flavor" agnostic. But if we wanted specific installer packages per-VM "flavor", we could include the image in the `tools` folder of `flarevm.installer.vm`.

mr-tz: gotcha, thanks, that's cool (just didn't see the new function for this before the comment)

PR #142, opened 2022-12-02
fix: use category Networking

mr-tz:


MalwareMechanic:
@mr-tz Make sure to bump the versions... when does that get caught? Was a test supposed to fail?

MalwareMechanic (changes requested):
Increase the version in the `.nuspec` file plz

mr-tz:
oh yeah, the linter failed to fail this... investigating

mr-tz:
not sure what the linter bug was, but I renamed the file

MalwareMechanic (approved):

Ana06 (approved):

Issue #148, opened 2022-12-06
cyberchef install broken


rangerthegood:
### What's the problem?

The cyberchef install is failing in your testing automation with the following error.

Installing the following packages:
cyberchef.vm
By installing, you accept licenses for the packages.

GoogleChrome v108.0.5359.72 (forced) [Approved]
googlechrome package files install completed. Performing other installation steps.
Downloading googlechrome 64 bit
  from 'https://dl.google.com/tag/s/dl/chrome/install/googlechromestandaloneenterprise64.msi'

Download of googlechromestandaloneenterprise64.msi (90.95 MB) completed.
Error - hashes do not match. Actual value was 'F99755B41CFEB5289802FE163F871D5F9DE7DA10F02067816534BAE7E09AB8B5'.
ERROR: Checksum for 'C:\Users\runneradmin\AppData\Local\Temp\chocolatey\GoogleChrome\108.0.5359.72\googlechromestandaloneenterprise64.msi' did not meet '2f1561d3934261bb869ed1209117d82df202bf6095f6e643619b41f608311694' for checksum type 'sha256'. Consider passing the actual checksums through with --checksum --checksum64 once you validate the checksums are appropriate. A less secure option is to pass --ignore-checksums if necessary.
The install of googlechrome was NOT successful.
Error while running 'C:\ProgramData\chocolatey\lib\GoogleChrome\tools\chocolateyInstall.ps1'.
 See log for details.

cyberchef.vm v9.49.0.20221201 (forced)
cyberchef.vm package files install completed. Performing other installation steps.
Downloading cyberchef.vm 
  from 'https://github.com/gchq/CyberChef/releases/download/v9.49.0/CyberChef_v9.49.0.zip'

Download of CyberChef_v9.49.0.zip (37.58 MB) completed.
Hashes match.
Extracting C:\Users\runneradmin\AppData\Local\Temp\chocolatey\cyberchef.vm\9.49.0.20221201\CyberChef_v9.49.0.zip to C:\Tools\cyberchef...
C:\Tools\cyberchef
 The install of cyberchef.vm was successful.
  Software installed to 'C:\Tools\cyberchef'

Chocolatey installed 1/2 packages. 1 packages failed.
 See the log for details (C:\ProgramData\chocolatey\logs\chocolatey.log).

Failures
 - googlechrome (exited -1) - Error while running 'C:\ProgramData\chocolatey\lib\GoogleChrome\tools\chocolateyInstall.ps1'.
[...]

PR #149, opened 2022-12-06
Update cyberchef.vm.nuspec

rangerthegood:
bump googlechrome version to latest.

should fix: #148


mr-tz:
Awesome, thanks for the fix, this looks good to me.

PR #191, opened 2022-12-15
Fix README.md links

mwilliams31:
Getting a 404 on existing links for Issues and Wiki, which point to https://github.com/mandiant/VM-Packages/blob/main/issues and https://github.com/mandiant/VM-Packages/blob/main/wiki


mr-tz:
Thank you!

PR #201, opened 2023-01-02
NetworkMiner 2.8 released

erik4711:
Release blog post:
https://netresec.com/?b=231d523

PR #202, opened 2023-01-02
NetworkMiner 2.8 released

erik4711:


mr-tz:
Handling this in #201

Issue #212, opened 2023-01-05
linter `VersionNotUpdated` appears to fail in CI


mr-tz:
see e.g. in https://github.com/mandiant/VM-Packages/pull/211

mr-tz:
Suspect this could be an issue with git comparing between forks...

PR #214, opened 2023-01-07
Added 64-bit HollowsHunter.

t-mtsmt:
The 64-bit hollows_hunter was not specified, therefore it was added.

### DIFF
--- a/packages/hollowshunter.vm/hollowshunter.vm.nuspec
+++ b/packages/hollowshunter.vm/hollowshunter.vm.nuspec
@@ -2,7 +2,7 @@
 <package xmlns="http://schemas.microsoft.com/packaging/2015/06/nuspec.xsd">
   <metadata>
     <id>hollowshunter.vm</id>
-    <version>0.3.5</version>
+    <version>0.3.5.20230202</version>
     <authors>hasherezade</authors>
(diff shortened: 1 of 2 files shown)


mr-tz:
Thank you.
Looks like our CI has some issues at the moment:
- the "get changed file" step failed so this has not been tested in CI properly yet
- also the nuspec version needs to be updated which should have been caught by the linter (issue #212)

Ana06 (approved):
@t-mtsmt thanks for your PR. I have updated your commit to include the timestamp in the version (we need to always increase the version to update a package, see [our documentation](https://github.com/mandiant/VM-Packages/wiki/Package-Structure#version)). I have also changed`Added`by `Add` in the commit message and removed the period, as they are [good practices](https://chiamakaikeanyi.dev/how-to-write-good-git-commit-messages/). So we should be good to go now 😄

Ana06:
@mr-tz get-changed-files didn't fail. It relays on the linter failing if the nuspec is not updated (only looks for modified .nuspec). Thanks for fixing the linter issue! 😄

mr-tz:
Ah, right, I see the filter on the action now :)

Thanks for the update here.

mr-tz (approved):

PR #224, opened 2023-01-31
handle utf-8 and ignore case when searching dependency

mr-tz:
This should improve automatic updates for cyberchef (googlechrome vs. GoogleChrome) and other such cases.

### DIFF
--- a/scripts/utils/update_package.py
+++ b/scripts/utils/update_package.py
@@ -96,7 +96,7 @@ def update_github_url(package):
 # Metapackages have only one dependency and same name (adding `.vm`)  and version as the dependency
 def update_dependencies(package):
     nuspec_path = f"packages/{package}/{package}.nuspec"
-    with open(nuspec_path, "r") as file:
+    with open(nuspec_path, "r", encoding="utf-8") as file:
         content = file.read()
     matches = re.findall(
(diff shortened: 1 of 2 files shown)


mr-tz:
Coincidentally, this could help with some update issues seen in https://github.com/mandiant/flare-vm/issues/437.

scripts/utils/update_package.py
```diff
@@ -96,7 +96,7 @@ def update_github_url(package):
 # Metapackages have only one dependency and same name (adding `.vm`)  and version as the dependency
 def update_dependencies(package):
     nuspec_path = f"packages/{package}/{package}.nuspec"
-    with open(nuspec_path, "r") as file:
+    with open(nuspec_path, "r", encoding="utf-8") as file:
```
mr-tz, line 99: see error: https://github.com/mandiant/VM-Packages/actions/runs/4052201794/jobs/6971349734#step:5:70

MalwareMechanic (approved):
Oh, good find!

mr-tz:
Looks like it's working! https://github.com/mandiant/VM-Packages/pull/225

tuttimann:
For me there are still many package update failures as seen in https://github.com/mandiant/flare-vm/issues/437

mr-tz:
We haven't merged the proposed updates yet, so the packages are not available.
Hope @MalwareMechanic can shed some light on your referenced issue.

PR #226, opened 2023-02-03
improve dependency handling and use all lower case

mr-tz:
add lint and ensure update script comparison compares case insensitive

### DIFF
--- a/scripts/utils/update_package.py
+++ b/scripts/utils/update_package.py
@@ -119,7 +119,8 @@ def update_dependencies(package):
                     content,
                 )
                 updates = True
-                if dependency == package[:-3]:  # Metapackage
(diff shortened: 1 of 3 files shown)


scripts/utils/update_package.py
```diff
@@ -119,7 +119,8 @@ def update_dependencies(package):
                     content,
                 )
                 updates = True
-                if dependency == package[:-3]:  # Metapackage
+                # both should be all lowercase via the linter, but let's be sure here
+                if dependency.lower() == package[:-3].lower():  # Metapackage
```
Ana06, line 123: This shouldn't be needed as the linter is now checking this, but it should be more robust 👍

Ana06 (approved):
Thanks for fixing this @mr-tz!

PR #236, opened 2023-02-10
Add x64dbg DbgChild

thejoelpatrol:
Add a popular plugin for x64dbg that automatically attaches to any spawned child processes. 

This is the code of the tool: https://github.com/therealdreg/DbgChild

The zip package is downloaded from GitHub, but it's somewhat more complicated than your average zip because the files need to go in several different locations within the x64dbg directory.

### DIFF
--- a/packages/x64dbg.dbgchild.vm/tools/chocolateyinstall.ps1
+++ b/packages/x64dbg.dbgchild.vm/tools/chocolateyinstall.ps1
@@ -1,0 +1,62 @@
+$ErrorActionPreference = 'Stop'
+Import-Module vm.common -Force -DisableNameChecking
+
+try {
+    $tempDownloadDir = Join-Path ${Env:chocolateyPackageFolder} "temp_$([guid]::NewGuid())"
+    $toolDir = Join-Path ${Env:RAW_TOOLS_DIR} 'x64dbg\release' -Resolve
+    $packageArgs = @{
+        packageName   = ${Env:ChocolateyPackageName}
+        unzipLocation = $tempDownloadDir
+        url           = 'https://github.com/therealdreg/DbgChild/releases/download/beta10/DbgChild.Beta.10.zip'
+        checksum      = 'f17f588795d8f5f94d71335a8acfa58946bb03a94a5637be7f3e804c652ea2b4'
+        checksumType  = 'sha256'
+    }
+
+    VM-Remove-PreviousZipPackage ${Env:chocolateyPackageFolder}
+    Install-ChocolateyZipPackage @packageArgs
+    $unzippedDir = (Get-ChildItem -Directory $tempDownloadDir | Where-Object {$_.PSIsContainer} | Select-Object -f 1).FullName
+    VM-Assert-Path $unzippedDir
+
+    $archs = @("x32", "x64")
+    foreach ($arch in $archs) {
+        $archDstDir = Join-Path $toolDir "${arch}" -Resolve
+        $pluginDstDir = Join-Path $archDstDir 'plugins'
+        if (-Not (Test-Path $pluginDstDir -PathType Container)) {
+            New-Item -ItemType directory $pluginDstDir -Force -ea 0 | Out-Null
+        }
+        VM-Assert-Path $pluginDstDir
+
+        # Move 32/64-bit plugin DLL itself into the arch directory
+        $pluginSrcPath = Join-Path $unzippedDir "release\${arch}\plugins" -Resolve
+        Get-ChildItem -Path $pluginSrcPath -File | Move-Item -Destination $pluginDstDir -Force
+
+        # Note that we don't simply move all children including directories, because we don't want to overwrite plugins
+
+        # Move all the other arch-specific files
+        $archSrcPath = Join-Path $unzippedDir "release\${arch}" -Resolve
+        Get-ChildItem -Path $archSrcPath -File | Move-Item -Destination $archDstDir -Force
+        if (-Not(Test-Path "${archDstDir}\CPIDS" -PathType Container)) {
+            New-Item -ItemType directory "${archDstDir}\CPIDS" -Force -ea 0 | Out-Null
+        }
+    }
+
+    # Move the NewProcessWatcher and text files into the main x64dbg directory
+    $releaseSrcDir = Join-Path $unzippedDir 'release'
+
+    Get-ChildItem -Path $releaseSrcDir -File | Move-Item -Destination $toolDir -Force
+    if (-Not(Test-Path "${toolDir}\dbgchildlogs" -PathType Container)) {
+        Move-Item -Path "${releaseSrcDir}\dbgchildlogs" -Destination $toolDir
+    }
+
+    # Make sure at least one of the files in each dir ended up in the right place
+    VM-Assert-Path "${toolDir}\NewProcessWatcher.exe"
+    VM-Assert-Path "${toolDir}\x32\CreateProcessPatch.exe"
+    VM-Assert-Path "${toolDir}\x64\CreateProcessPatch.exe"
+    VM-Assert-Path "${toolDir}\x32\plugins\dbgchild.dp32"
+    VM-Assert-Path "${toolDir}\x64\plugins\dbgchild.dp64"
+
+    # $unzippedDir is in $tempDownloadDir, so this should clean up both of them
+    Remove-Item $tempDownloadDir -Recurse -Force -ea 0
+} catch {
+    VM-Write-Log-Exception $_
+}
--- a/packages/x64dbg.dbgchild.vm/tools/chocolateyuninstall.ps1
+++ b/packages/x64dbg.dbgchild.vm/tools/chocolateyuninstall.ps1
@@ -1,0 +1,15 @@
+$ErrorActionPreference = 'Continue'
+
+$toolDir = Join-Path ${Env:RAW_TOOLS_DIR} 'x64dbg\release'
+foreach ($file in $("NewProcessWatcher.exe", "readme_dbgchild.txt", "x64_post.unicode.txt", "x64_pre.unicode.txt", "x86_post.unicode.txt", "x86_pre.unicode.txt", "dbgchildlogs")) {
+    Remove-Item "${toolDir}\${file}" -Recurse -Force
+}
+
+$archFiles = @("CreateProcessPatch.exe", "DbgChildHookDLL.dll", "NTDLLEntryPatch.exe", "CPIDS")
+foreach ($arch in @("32", "64")) {
+    $toolDir = Join-Path ${Env:RAW_TOOLS_DIR} "x64dbg\release\x${arch}"
+    foreach ($file in $archFiles) {
+        Remove-Item "${toolDir}\${file}" -Recurse -Force
+    }
+    Remove-Item "${toolDir}\plugins\dbgchild.dp${arch}" -Force
+}
--- a/packages/x64dbg.dbgchild.vm/x64dbg.dbgchild.vm.nuspec
+++ b/packages/x64dbg.dbgchild.vm/x64dbg.dbgchild.vm.nuspec
@@ -1,0 +1,13 @@
+<?xml version="1.0" encoding="utf-8"?>
+<package xmlns="http://schemas.microsoft.com/packaging/2015/06/nuspec.xsd">
+  <metadata>
+    <id>x64dbg.dbgchild.vm</id>
+    <version>10</version>
+    <description>DbgChild is an x64dbg plugin to automatically attach to spawned child processes.</description>
+    <authors>Dreg (David Reguera Garcia)</authors>
+    <dependencies>
+      <dependency id="common.vm" />
+      <dependency id="x64dbg.vm" />
+    </dependencies>
+  </metadata>
+</package>
--- a/scripts/test/lint.py
+++ b/scripts/test/lint.py
@@ -304,6 +304,7 @@ class UsesInvalidCategory(Lint):
     EXCLUSIONS = [
         ".ollydumpex.vm",
         ".scyllahide.vm",
+        ".dbgchild.vm",
         "common.vm",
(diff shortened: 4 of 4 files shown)


packages/x64dbg.dbgchild.vm/tools/chocolateyinstall.ps1
```diff
@@ -0,0 +1,62 @@
+$ErrorActionPreference = 'Stop'
+Import-Module vm.common -Force -DisableNameChecking
+
+VM-Remove-PreviousZipPackage ${Env:chocolateyPackageFolder}
```
MalwareMechanic, line 4: Please move inside the try/catch

packages/x64dbg.dbgchild.vm/tools/chocolateyinstall.ps1
```diff
@@ -0,0 +1,62 @@
...
+        $archDstDir = Join-Path $toolDir "${arch}" -Resolve
+        $pluginDstDir = Join-Path $archDstDir 'plugins'
+        if (-Not (Test-Path $pluginDstDir -PathType Container)) {
+            New-Item -ItemType directory $pluginDstDir -Force -ea 0 | Out-Null
+        }
+        VM-Assert-Path $pluginDstDir
+
+        # Move 32/64-bit plugin DLL itself into the arch directory
+        $pluginSrcPath = Join-Path $toolSrcDir "release\${arch}\plugins" -Resolve
+        Get-ChildItem -Path $pluginSrcPath -File | Move-Item -Destination $pluginDstDir -Force -vb
```
MalwareMechanic, line 32: What are the `-vb` at the end of these lines?

packages/x64dbg.dbgchild.vm/tools/chocolateyinstall.ps1
```diff
@@ -0,0 +1,62 @@
...
+        
+        # Note that we don't simply move all children including directories, because we don't want to overwrite plugins
+
+        # Move all the other arch-specific files
+        $archSrcPath = Join-Path $toolSrcDir "release\${arch}" -Resolve
+        Get-ChildItem -Path $archSrcPath -File | Move-Item -Destination $archDstDir -Force -vb
+        if (-Not(Test-Path "${archDstDir}\CPIDS" -PathType Container)) {
```
MalwareMechanic, line 39: What would happen on an upgrade? The path may exist and then the move wouldn't happen.

Perhaps if the folder doesn't exists, create it. After the folder existence check, copy over the files from the source to the destination. This would allow the upgrade scenario to simply overwrite the pre-existing files.

packages/x64dbg.dbgchild.vm/tools/chocolateyinstall.ps1
```diff
@@ -0,0 +1,62 @@
...
+} catch {
+    VM-Write-Log-Exception $_
+}
+
+# Make sure at least one of the files in each dir ended up in the right place
+VM-Assert-Path "${toolDir}\NewProcessWatcher.exe"
+VM-Assert-Path "${toolDir}\x32\CreateProcessPatch.exe"
+VM-Assert-Path "${toolDir}\x64\CreateProcessPatch.exe"
+VM-Assert-Path "${toolDir}\x32\plugins\dbgchild.dp32"
+VM-Assert-Path "${toolDir}\x64\plugins\dbgchild.dp64"
```
MalwareMechanic, line 62: Please put these in the try/catch

packages/x64dbg.dbgchild.vm/tools/chocolateyuninstall.ps1
```diff
@@ -0,0 +1,15 @@
+$ErrorActionPreference = 'Continue'
+
+$toolDir = Join-Path ${Env:RAW_TOOLS_DIR} 'x64dbg\release'
+foreach ($file in $("NewProcessWatcher.exe", "readme_dbgchild.txt", "x64_post.unicode.txt", "x64_pre.unicode.txt", "x86_post.unicode.txt", "x86_pre.unicode.txt", "dbgchildlogs")) {
+    Remove-Item "${toolDir}\${file}" -Recurse -Force -vb
```
MalwareMechanic, line 5: Not sure what the `-vb` do and if they're necessary

MalwareMechanic (dismissed):

packages/x64dbg.dbgchild.vm/tools/chocolateyinstall.ps1
thejoelpatrol: the `-Verbose` common parameter
https://learn.microsoft.com/en-us/powershell/module/microsoft.powershell.core/about/about_commonparameters?view=powershell-7.3
it's sometimes helpful for debugging, but I'll remove for production now

packages/x64dbg.dbgchild.vm/tools/chocolateyinstall.ps1
thejoelpatrol: For this directory that behavior is OK. It's where child PIDs are stored while debugging. If it exists on upgrade, we don't need to do anything. 

I guess `Move-Item` is probably not the clearest operation to use here, since the source directory is actually empty. I'll replace with `New-Item`.

packages/x64dbg.dbgchild.vm/tools/chocolateyuninstall.ps1
thejoelpatrol: removed

thejoelpatrol:
Comments addressed, please let me know if you see anything else :)

mr-tz (approved):
LGTM

packages/x64dbg.dbgchild.vm/tools/chocolateyinstall.ps1
```diff
@@ -0,0 +1,61 @@
+$ErrorActionPreference = 'Stop'
+Import-Module vm.common -Force -DisableNameChecking
+
+try {
+    $toolSrcDir = "$(Split-Path -parent $MyInvocation.MyCommand.Definition)"
```
Ana06, line 5: [nitpick] For other similar packages (for example [idr.vm](https://github.com/mandiant/VM-Packages/blob/main/packages/idr.vm/tools/chocolateyinstall.ps1#L15)), we create a temp directory for cases like this. @vm-packages any preference on how we should do this in general? I think I tend to prefer a temp directory as we don't plan to keep it as the tool directory.

packages/x64dbg.dbgchild.vm/tools/chocolateyinstall.ps1
```diff
@@ -0,0 +1,61 @@
...
+        unzipLocation = $toolSrcDir
+        url           = 'https://github.com/therealdreg/DbgChild/releases/download/beta10/DbgChild.Beta.10.zip'
+        checksum      = 'f17f588795d8f5f94d71335a8acfa58946bb03a94a5637be7f3e804c652ea2b4'
+        checksumType  = 'sha256'
+    }
+
+    VM-Remove-PreviousZipPackage ${Env:chocolateyPackageFolder}
+    Install-ChocolateyZipPackage @packageArgs
+    $toolSrcDir = Join-Path $toolSrcDir 'DbgChild Beta 10'
+    VM-Assert-Path $toolSrcDir
```
Ana06, line 18:
```suggestion
    $toolSrcDir = Join-Path $toolSrcDir 'DbgChild Beta 10' -Resolve
```

The assert is not needed if you use `-Resolve` in `Join-Path`.

packages/x64dbg.dbgchild.vm/tools/chocolateyinstall.ps1
```diff
@@ -0,0 +1,61 @@
...
+        packageName   = ${Env:ChocolateyPackageName}
+        unzipLocation = $toolSrcDir
+        url           = 'https://github.com/therealdreg/DbgChild/releases/download/beta10/DbgChild.Beta.10.zip'
+        checksum      = 'f17f588795d8f5f94d71335a8acfa58946bb03a94a5637be7f3e804c652ea2b4'
+        checksumType  = 'sha256'
+    }
+
+    VM-Remove-PreviousZipPackage ${Env:chocolateyPackageFolder}
+    Install-ChocolateyZipPackage @packageArgs
+    $toolSrcDir = Join-Path $toolSrcDir 'DbgChild Beta 10'
```
Ana06, line 17:
```suggestion
    $unzippedDir = (Get-ChildItem -Directory $tempDownloadDir | Where-Object {$_.PSIsContainer} | Select-Object -f 1).FullName
```
[niptick] I think I would use a different variable name for clarity and also make it non dependent on the version so that this package gets automatically updated

packages/x64dbg.dbgchild.vm/tools/chocolateyinstall.ps1
```diff
@@ -0,0 +1,61 @@
...
+    VM-Assert-Path $toolSrcDir
+
+    $archs = @("x32", "x64")
+    foreach ($arch in $archs) {
+        $archDstDir = Join-Path $toolDir "${arch}" -Resolve
+        $pluginDstDir = Join-Path $archDstDir 'plugins'
+        if (-Not (Test-Path $pluginDstDir -PathType Container)) {
+            New-Item -ItemType directory $pluginDstDir -Force -ea 0 | Out-Null
+        }
+        VM-Assert-Path $pluginDstDir
```
Ana06, line 27: [nitpick] I think this is not needed as you are using `-Force` before

packages/x64dbg.dbgchild.vm/x64dbg.dbgchild.vm.nuspec
```diff
@@ -0,0 +1,13 @@
+<?xml version="1.0" encoding="utf-8"?>
+<package xmlns="http://schemas.microsoft.com/packaging/2015/06/nuspec.xsd">
+  <metadata>
+    <id>x64dbg.dbgchild.vm</id>
+    <version>10.0.0</version>
```
Ana06, line 5: tricky though, as it is 10 beta. 
```suggestion
    <version>10</version>
```

[nitpick] I would change to 10, as I think the tool version does not use the second and third segments. This is

Ana06 (commented):
@thejoelpatrol this is great work! Sorry that we haven't merged yet. 

I have added some minor comments. If you don't have time to address them, I can take over.

It would also be great if you could squash you commits to help as keeping a clean commit history. :smile: 

@MalwareMechanic is there something else to address here?

packages/x64dbg.dbgchild.vm/tools/chocolateyinstall.ps1
thejoelpatrol: I may be wrong, as I am not a PowerShell expert. But I think there might still be scenarios that cause the directory to not be created? My reading of `-Force` when creating directories is that if the directory exists this will return the existing directory instead of erroring out, not that it guarantees the directory will then exist. It seems to help in cases like this: 

```
PS C:\Users\flare > New-Item -ItemType directory C:\Users\flare\
New-Item : An item with the specified name C:\Users\flare\ already exists.
At line:1 char:1
+ New-Item -ItemType directory C:\Users\flare\
+ ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
    + CategoryInfo          : ResourceExists: (C:\Users\flare\:String) [New-Item], IOException
    + FullyQualifiedErrorId : DirectoryExist,Microsoft.PowerShell.Commands.NewItemCommand

FLARE-VM 03/15/2024 20:33:15
PS C:\Users\flare > New-Item -ItemType directory C:\Users\flare\ -Force


    Directory: C:\Users


Mode                 LastWriteTime         Length Name
----                 -------------         ------ ----
d-----          3/7/2024   1:01 PM                flare
```

But it can't help if you give some totally bad path for whatever reason (here D: does not exist):

```
PS C:\Users\flare > New-Item -ItemType directory D:\bad -Force
New-Item : The path is not of a legal form.
At line:1 char:1
+ New-Item -ItemType directory D:\bad -Force
+ ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
    + CategoryInfo          : InvalidArgument: (D:\:String) [New-Item], ArgumentException
    + FullyQualifiedErrorId : CreateDirectoryArgumentError,Microsoft.PowerShell.Commands.NewItemCommand
```

packages/x64dbg.dbgchild.vm/tools/chocolateyinstall.ps1
thejoelpatrol: done!

packages/x64dbg.dbgchild.vm/tools/chocolateyinstall.ps1
thejoelpatrol: I think this is superseded by the following suggestion.

packages/x64dbg.dbgchild.vm/tools/chocolateyinstall.ps1
thejoelpatrol: I changed it to be similar to the `idr.vm` example you linked

thejoelpatrol:
@Ana06 I have submitted the CLA and addressed the comments. I hope that does the trick :)

naacbin:
> @Ana06 I have submitted the CLA and addressed the comments. I hope that does the trick :)

I think you will be asked to squash your commits into a single one. :)

Ana06:
it would be great if you could squash the commits and use the right email so that the CLA check stops complaining :smile:

Ana06:
The rebasing and CLA look good! But we have merged another PR that touches `lint.py` causing merge conflicts. Could you rebase and fix them? If you need help or don't have time, let me know and I'll do it. Sorry for all the rebasing, too many contributions lately (at least it because of a good thing)!

thejoelpatrol:
Conflict resolved. I hope no more PRs land before this one!

PR #239, opened 2023-02-13
fix version available check

mr-tz:
Noticed, because capa did not update. Should result in a dozen or so other updates.

### DIFF
--- a/scripts/utils/update_package.py
+++ b/scripts/utils/update_package.py
@@ -44,7 +44,7 @@ def format_version(version):
     # 1 -> 1
     match = re.match("v?(?P<version>\d+(.\d+){0,2})", version)
     if not match:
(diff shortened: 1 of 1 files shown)


scripts/utils/update_package.py
```diff
@@ -67,7 +67,7 @@ def update_github_url(package):
     for url, org, project, version in matches:
         latest_version_match = get_latest_version(org, project, version)
         # No newer version available
-        if (not latest_version) or (latest_version_match == version):
+        if (not latest_version_match) or (latest_version_match == version):
```
mr-tz, line 70: wrong variable name (this one is set to None before loop)

MalwareMechanic (approved):
LGTM

PR #241, opened 2023-02-13
add pdf-parser.py by Didier Stevens

thejoelpatrol:


mr-tz (approved):
LGTM

MalwareMechanic (changes requested):
Please add a Python dependency like `<dependency id="python3" version="[3.9.0, 3.10.0)" />` to the nuspec file.

Can you also update to use the a category like "PDF" and add some code to add a shortcut in the category folder like (below needs testing, just threw it together):

```
$targetPath = Join-Path $toolsDir "pdf-parser.py" -Resolve
$targetCmd  = Join-Path ${Env:WinDir} "system32\cmd.exe" -Resolve
$targetDir  = $toolDir
$targetArgs = '/K "' + "py -3 ^`"$targetPath^`""  + '" ' + "$(Join-Path ${Env:WinDir} 'System32\notepad.exe')"

$shortcut = Join-Path $path "$name.lnk"
Install-ChocolateyShortcut -shortcutFilePath $shortcut -targetPath $targetPath -Arguments $targetArgs -WorkingDirectory $targetDir
```

I believe you can also add that directory to the $PATH via:
```
Install-ChocolateyPath $toolDir
```

thejoelpatrol:
Good ideas, will do

MalwareMechanic (approved):
LGTM

packages/pdfparser.vm/pdfparser.nuspec
```diff
@@ -0,0 +1,13 @@
+<?xml version="1.0" encoding="utf-8"?>
+<package xmlns="http://schemas.microsoft.com/packaging/2015/06/nuspec.xsd">
+  <metadata>
+    <id>pdfparser.vm</id>
+    <version>0.7.7</version>
+    <description>pdf-parser.py: a solid script to extract objects from PDF files</description>
+    <authors>Didier Stevens</authors>
+    <dependencies>
+      <dependency id="common.vm" />
+      <dependency id="python3" version="(, 3.10.4]" />
```
mr-tz, line 10: Should this depend on `libraries.python3.vm` or a custom Python3 package? I can imagine errors around various Python versions (and keeping them in sync in packages).

Not sure if the CI error is related exactly to that:
```
WARNING: Generic MSI Error. This is a local environment error, not an issue with a package or the MSI itself - it could mean a pending reboot is necessary prior to install or something else (like the same version is already installed). Please see MSI log if available. If not, try again adding '--install-arguments="'/l*v c:\python3_msi_install.log'"'. Then search the MSI Log for "Return Value 3" and look above that for the error.
ERROR: Running ["C:\ProgramData\chocolatey\lib\python3\tools\python-3.10.4-amd64.exe" /quiet InstallAllUsers=1 PrependPath=1 TargetDir="C:\Python310" ] was not successful. Exit code was '1603'. Exit code indicates the following: Generic MSI Error. This is a local environment error, not an issue with a package or the MSI itself - it could mean a pending reboot is necessary prior to install or something else (like the same version is already installed). Please see MSI log if available. If not, try again adding '--install-arguments="'/l*v c:\python3_msi_install.log'"'. Then search the MSI Log for "Return Value 3" and look above that for the error..
The install of python3 was NOT successful.
```

MalwareMechanic: Good catch! I see two options:

1. When a package needs Python, depend on `libraries.pythonX.vm`. Downside is it also installs all the Python libraries which may or may not be wanted.
2. Create a new metapackage to install a specific version of PythonX to unify / simply the global VM Python version. Pro is that we can have `libraries.pythonX.vm` depend on this new metapackage as well as other packages that want to utilize Python as well. Con is if we need multiple versions of Python (e.g., 3.7 and 3.10) for whatever reason.

I'm in favor of option `2` above. Thoughts?

mr-tz: 2. sounds good! Ideally, we can focus on one core version and let users handle any special use-cases.

There's also pyenv, but I have no experience with it and it does not officially support Windows.

MalwareMechanic: Works for me, I'll get this up pretty soon with a PR, then we can circle back to this package

MalwareMechanic: Added with #263

MalwareMechanic: Welp, I added it but I need to fix it. I got too hasty. Will post back once fixed

Ana06:
@MalwareMechanic what is the state of this PR?

Ana06:
I think this tool is already included in [didier-stevens-suite.vm](https://github.com/mandiant/VM-Packages/tree/efc0978e5be3e67c41282ffa9e0f5303974614eb/packages/didier-stevens-suite.vm). @thejoelpatrol is the [didier-stevens-suite.vm](https://github.com/mandiant/VM-Packages/tree/efc0978e5be3e67c41282ffa9e0f5303974614eb/packages/didier-stevens-suite.vm) working for you? can we close this PR?

PR #242, opened 2023-02-13
add Linux `file` utility to read magic bytes and determine file type

thejoelpatrol:
Also add a right-click shortcut to quickly run it on any file. I find this useful when downloading a bunch of files by hash that are missing extensions and you don't even know what they are.

### DIFF
--- a/packages/file.vm/tools/chocolateyinstall.ps1
+++ b/packages/file.vm/tools/chocolateyinstall.ps1
@@ -9,7 +9,6 @@ try {
     $zipSha256 = "963147318f96d9345471e1a9a3943def4d95fcb3c1fe020e465ab910d0cda4a3"
 
     $executablePath = (VM-Install-From-Zip $toolName $category $zipUrl -zipSha256 $zipSha256 -consoleApp $true)[-1]
-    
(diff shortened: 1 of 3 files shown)


packages/file.vm/tools/chocolateyinstall.ps1
```diff
@@ -0,0 +1,20 @@
...
+
+try {
+    $toolName = 'file'
+    $category = 'Utilities'
+
+    $zipUrl = "https://github.com/nscaife/file-windows/releases/download/20170108/file-windows-20170108.zip"
+    $zipSha256 = "963147318f96d9345471e1a9a3943def4d95fcb3c1fe020e465ab910d0cda4a3"
+
+    $executablePath = (VM-Install-From-Zip $toolName $category $zipUrl -zipSha256 $zipSha256 -consoleApp $true)[-1]
+
```
mr-tz, line 12:
```suggestion

```

mr-tz: the linter complains

mr-tz (approved):

PR #243, opened 2023-02-13
Add LECmd by Eric Zimmerman

thejoelpatrol:
This tool parses Windows shortcut files (LNK), which can do all kinds of stuff besides simply point to local disk files.

This version requires .NET version 4, which is apparently already installed in FLARE VM. (there's also a version for .NET 6 which did not work for me in the VM). I'm not sure if there's a way to declare that dependency, or if we can just assume it's there.

### DIFF
--- a/packages/lecmd.vm/lecmd.vm.nuspec
+++ b/packages/lecmd.vm/lecmd.vm.nuspec
@@ -1,0 +1,12 @@
+<?xml version="1.0" encoding="utf-8"?>
+<package xmlns="http://schemas.microsoft.com/packaging/2015/06/nuspec.xsd">
+  <metadata>
+    <id>lecmd.vm</id>
+    <version>1.5.0.0</version>
+    <description>LECmd: parse Windows shortcut files with LNK Explorer Command Line edition</description>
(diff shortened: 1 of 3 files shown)


Ana06:
@thejoelpatrol sorry that we haven't reviewed this PR  before. 
> This version requires .NET version 4, which is apparently already installed in FLARE VM. (there's also a version for .NET 6 which did not work for me in the VM). I'm not sure if there's a way to declare that dependency, or if we can just assume it's there.

You should require the dependency explicitly. Likely the dependency is either installed by default in your windows version or required by another package installed by flare-vm. The tool should be able to install correctly in other environments. See other .net packages, for example: https://github.com/mandiant/VM-Packages/blob/efc0978e5be3e67c41282ffa9e0f5303974614eb/packages/garbageman.vm/garbageman.vm.nuspec#L10

Ana06:
@thejoelpatrol is it ok with you if we close this PR in favour of https://github.com/mandiant/VM-Packages/pull/778?

thejoelpatrol:
No problem, go ahead

PR #244, opened 2023-02-14
Add SQLRecon for CommandoVM

nos3curity:
SQLRecon is a C# toolkit for exploitation of MSSQL databases similar to PowerUpSQL.

https://github.com/skahwah/SQLRecon


mr-tz:
Thank you!

PR #246, opened 2023-02-14
update packages

mr-tz:
ran update script locally, not sure why this failed via the CI workflow (added a print on the update script to notice an error case)

### DIFF
--- a/packages/cyberchef.vm/cyberchef.vm.nuspec
+++ b/packages/cyberchef.vm/cyberchef.vm.nuspec
@@ -7,7 +7,7 @@
     <description>The Cyber Swiss Army Knife - a web app for encryption, encoding, compression, data analysis, and more.</description>
     <dependencies>
       <dependency id="common.vm" />
-      <dependency id="googlechrome" version="[110.0.5481.78]" />
+      <dependency id="googlechrome" version="[110.0.5481.100]" />
     </dependencies>
   </metadata>
 </package>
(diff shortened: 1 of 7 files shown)


MalwareMechanic:
Seems a hash mismatch for `googlechrome`. I wonder if they're constantly updating it?? Could we potentially lock the version?

```
Downloading googlechrome 64 bit
  from 'https://dl.google.com/tag/s/dl/chrome/install/googlechromestandaloneenterprise64.msi'

Download of googlechromestandaloneenterprise64.msi (92.33 MB) completed.
Error - hashes do not match. Actual value was '983287A6BEADBF368DD15[43](https://github.com/mandiant/VM-Packages/actions/runs/4176646219/jobs/7233245204#step:4:44)BCB3EDFA917B9827406FE760B79A609FEBA12AEA8'.
ERROR: Checksum for 'C:\Users\runneradmin\AppData\Local\Temp\chocolatey\GoogleChrome\110.0.5481.78\googlechromestandaloneenterprise64.msi' did not meet '568131a0375efd4dbf4f790031fec8131a40aa7b2cb85d9be58cf0c371ba3f35' for checksum type 'sha256'. Consider passing the actual checksums through with --checksum --checksum64 once you validate the checksums are appropriate. A less secure option is to pass --ignore-checksums if necessary.
The install of googlechrome was NOT successful.
```

mr-tz:
The Chrome installation URL is always the same so new versions require an updated package and we cannot lock the version.

Seems that this is handled fairly quickly by the community though.

mr-tz:
Thanks, @MalwareMechanic!

PR #253, opened 2023-02-15
Add Mozilla Thunderbird

thejoelpatrol:
Thunderbird is useful for viewing malicious emails (and their attachments) exported from the mail server, e.g. as .eml files. Built in Windows Mail requires you to add an account even before opening local files.

### DIFF
--- a/packages/thunderbird.vm/tools/chocolateyuninstall.ps1
+++ b/packages/thunderbird.vm/tools/chocolateyuninstall.ps1
@@ -1,0 +1,7 @@
+$ErrorActionPreference = 'Continue'
+Import-Module vm.common -Force -DisableNameChecking
+
+$toolName = 'Thunderbird'
+$category = 'Email'
+
+VM-Remove-Tool-Shortcut $toolName $category
--- a/packages/thunderbird.vm/thunderbird.vm.nuspec
+++ b/packages/thunderbird.vm/thunderbird.vm.nuspec
@@ -1,0 +1,13 @@
+<?xml version="1.0" encoding="utf-8"?>
+<package xmlns="http://schemas.microsoft.com/packaging/2015/06/nuspec.xsd">
+  <metadata>
+    <id>thunderbird.vm</id>
+    <version>102.7.2</version>
+    <authors>MZLA Technologies Corporation (Mozilla)</authors>
+    <description>Thunderbird is an open-source email client that can open saved email messages such as .eml files</description>
+    <dependencies>
+      <dependency id="common.vm" />
+      <dependency id="thunderbird" version="[102.7.2]" />
+    </dependencies>
+  </metadata>
+</package>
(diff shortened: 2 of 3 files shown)


packages/thunderbird.vm/tools/chocolateyuninstall.ps1
```diff
@@ -0,0 +1,7 @@
+$ErrorActionPreference = 'Continue'
+Import-Module vm.common -Force -DisableNameChecking
+
+$toolName = 'Thunderbird'
+$category = 'Email'
```
Ana06, line 5: The categories are fixed and there is no Email category. See discussion in https://github.com/mandiant/VM-Packages/issues/38

@mr-tz should we add a category check to the linter?

Ana06 (commented):
@thejoelpatrol thanks a lot for contributing to VM-Packages. The package looks good. But I am not sure if Mozilla Thunderbird should be part of VM-Packages, as it is not an analysis tool itself, even if you can use it to open malicious emails. A similar example are Web browsers that can be used to debug malicious JavaScript, although it is not their main application. We are at the moment installing browsers using directly the choco package and there is not `chrome.vm` package.

Any other ideas? @mandiant/flare-vm

packages/thunderbird.vm/tools/chocolateyuninstall.ps1
mr-tz: Isn't there one? Checks didn't run here yet.

I'm also unsure if we should add this here.

day1player:
@Ana06 it could be useful for Commando VM, for example, sending phishing email campaigns.

packages/thunderbird.vm/tools/chocolateyuninstall.ps1
mr-tz: There's not category check, so we could add that. Still confusing that no Actions run here.

mr-tz:
Could we add standard choco packages to our list/config? They won't show up in the Tools directory, but can still be installed as part of the default setup (if selected by an user).

packages/thunderbird.vm/tools/chocolateyuninstall.ps1
Ana06: This is what you meant https://github.com/mandiant/VM-Packages/issues/270 😅

thejoelpatrol:
> Could we add standard choco packages to our list/config? They won't show up in the Tools directory, but can still be installed as part of the default setup (if selected by an user).

This would work fine for my purpose. I've run into some other standard choco packages I would also like. They don't have to show up in Tools, that's fine. It would be best not to have to wrap them here, too. Is there a recommended way to install choco packages other than just doing it manually?

day1player:
@thejoelpatrol you can include the chocolatey package name in the config file, it will pull from chocolatey sources. Can be done by editing the `.xml` file and just adding a line in, it should be installed automatically if done before running the install script.

Ana06:
I am going to close this PR as I think we agree that we don't need a .vm package for this (if someone disagrees please open an issue for discussion). Thanks for all the work @thejoelpatrol!

PR #255, opened 2023-02-16
added telnet package

geo-lit:
added telnet package (source: https://community.chocolatey.org/packages/telnet). not consistent with output of create_package_template.py because tool is installed in System32 directory.


geo-lit:
adding versioning lock for the install script.

PR #256, opened 2023-02-21
Adding Mimikatz

Menn1s:
For extracting passwords and hashes. By gentilkiwi 

[https://github.com/gentilkiwi/mimikatz](https://github.com/gentilkiwi/mimikatz)

### DIFF
--- a/packages/mimikatz.vm/mimikatz.vm.nuspec
+++ b/packages/mimikatz.vm/mimikatz.vm.nuspec
@@ -2,7 +2,7 @@
 <package xmlns="http://schemas.microsoft.com/packaging/2015/06/nuspec.xsd">
   <metadata>
     <id>mimikatz.vm</id>
-    <version>2.2.0_20220919</version>
+    <version>2.2.0</version>
     <authors>Benjamin Delpy, gentilkiwi</authors>
     <description>Mimikatz is an open-source application that allows users to view and save authentication credentials such as Kerberos tickets</description>
     <dependencies>
--- a/packages/mimikatz.vm/mimikatz.vm.nuspec
+++ b/packages/mimikatz.vm/mimikatz.vm.nuspec
@@ -1,0 +1,12 @@
+<?xml version="1.0" encoding="utf-8"?>
+<package xmlns="http://schemas.microsoft.com/packaging/2015/06/nuspec.xsd">
+  <metadata>
+    <id>mimikatz.vm</id>
+    <version>2.2.0_20220919</version>
+    <authors>Benjamin Delpy, gentilkiwi</authors>
+    <description>Mimikatz is an open-source application that allows users to view and save authentication credentials such as Kerberos tickets</description>
+    <dependencies>
+      <dependency id="common.vm" />
+    </dependencies>
+  </metadata>
+</package>
--- a/packages/mimikatz.vm/tools/chocolateyinstall.ps1
+++ b/packages/mimikatz.vm/tools/chocolateyinstall.ps1
@@ -1,15 +1,13 @@
 $ErrorActionPreference = 'Stop'
 Import-Module vm.common -Force -DisableNameChecking
+$category = 'Password Attacks'
 
 try {
-
-  $category = 'Password Attacks'
-
 
   $ErrorActionPreference = 'Stop'
   Import-Module vm.common -Force -DisableNameChecking
   $name = "Mimikatz"
-  $path = Join-Path ${Env:TOOL_LIST_DIR} 'Password Attacks'
+  $path = Join-Path ${Env:TOOL_LIST_DIR} $category
 
   # Remove files from previous zips for upgrade
   VM-Remove-PreviousZipPackage ${Env:chocolateyPackageFolder}
--- a/packages/mimikatz.vm/tools/chocolateyinstall.ps1
+++ b/packages/mimikatz.vm/tools/chocolateyinstall.ps1
@@ -1,6 +1,6 @@
 $ErrorActionPreference = 'Stop'
 Import-Module vm.common -Force -DisableNameChecking
-$category = 'Password Attacks'
+$category = 'Credential Access'
 
 try {
 
--- a/packages/mimikatz.vm/tools/chocolateyinstall.ps1
+++ b/packages/mimikatz.vm/tools/chocolateyinstall.ps1
@@ -35,7 +35,7 @@ try {
   $target_cmd = Join-Path ${Env:WinDir} "system32\cmd.exe"
   $target_args = '/K "' + $target + '"'
   $target_icon = $target_cmd
-  $target_dir = Join-Path ${Env:UserProfile} "Desktop"
+  $target_dir = $toolsDir
   Install-ChocolateyShortcut -shortcutFilePath $shortcut -targetPath $target_cmd -Arguments $target_args -WorkingDirectory $target_dir -IconLocation $target_icon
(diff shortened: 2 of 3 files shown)


packages/mimikatz.vm/tools/chocolateyinstall.ps1
```diff
@@ -0,0 +1,49 @@
+$ErrorActionPreference = 'Stop'
+Import-Module vm.common -Force -DisableNameChecking
+
+$toolName = 'mimikatz'
```
MalwareMechanic, line 4: Be sure to enclose code in a try / catch. For an example see:
- https://github.com/mandiant/VM-Packages/blob/main/packages/ollydbg2.scyllahide.vm/tools/chocolateyinstall.ps1

packages/mimikatz.vm/tools/chocolateyinstall.ps1
```diff
@@ -0,0 +1,49 @@
...
+
+# Remove files from previous zips for upgrade
+VM-Remove-PreviousZipPackage ${Env:chocolateyPackageFolder}
+
+$toolsDir = Join-Path "${Env:RAW_TOOLS_DIR}" $name
+if (-Not (Test-Path $toolsDir)) {
+  New-Item -Path $toolsDir -ItemType Directory -Force | Out-Null
+}
+
+$url = "https://github.com/gentilkiwi/mimikatz/releases/download/2.2.0-20220919/mimikatz_trunk.7z"
```
MalwareMechanic, line 21: You'll likely want to use `VM-Install-From-Zip` (see: https://github.com/mandiant/VM-Packages/blob/main/packages/common.vm/tools/vm.common/vm.common.psm1#L269).

Here's an example using it: https://github.com/mandiant/VM-Packages/blob/main/packages/gobuster.vm/tools/chocolateyinstall.ps1

In general, I recommend reading over all the helper functions in https://github.com/mandiant/VM-Packages/blob/main/packages/common.vm/tools/vm.common/vm.common.psm to know what's available

packages/mimikatz.vm/tools/chocolateyuninstall.ps1
```diff
@@ -0,0 +1,15 @@
+$ErrorActionPreference = 'Stop'
+$name = "Mimikatz"
+$path = Join-Path ${Env:TOOL_LIST_DIR} 'Password Attacks'
+
+$toolsDir = Join-Path "${Env:RAW_TOOLS_DIR}" $name
```
MalwareMechanic, line 5: Like above, you'll likely want `VM-Uninstall` (see: https://github.com/mandiant/VM-Packages/blob/main/packages/common.vm/tools/vm.common/vm.common.psm1#L473).

Here's an example of it being used: https://github.com/mandiant/VM-Packages/blob/main/packages/gobuster.vm/tools/chocolateyinstall.ps1

packages/mimikatz.vm/mimikatz.vm.nuspec
```diff
@@ -0,0 +1,12 @@
+<?xml version="1.0" encoding="utf-8"?>
+<package xmlns="http://schemas.microsoft.com/packaging/2015/06/nuspec.xsd">
+  <metadata>
+    <id>mimikatz.vm</id>
+    <version>2.2.0_20220919</version>
```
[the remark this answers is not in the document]
MalwareMechanic: So I'd use `2.2.0` as the version. We reserve the last unit for our own tracking of the date of package modification (e.g., it we updated this package on 03/15/2023 we'd tack on `2.2.0.20230315`).

MalwareMechanic (changes requested):

packages/mimikatz.vm/mimikatz.vm.nuspec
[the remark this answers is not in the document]
Menn1s: Version changed to 2.2.0

packages/mimikatz.vm/tools/chocolateyinstall.ps1
Menn1s: Enclosed in try/catch

packages/mimikatz.vm/tools/chocolateyinstall.ps1
Menn1s: I looked over the help tool `VM-Install-From-Zip`, but found that its current options are unable to reach the nest files for Mimikatz zip file structure (there is only an option to go down one nested folder of the same name as the tool). 

I may look into modifying the helper tool for future installations, but it looks like Mimikatz is an outlier.

packages/mimikatz.vm/tools/chocolateyinstall.ps1
```diff
@@ -0,0 +1,56 @@
+$ErrorActionPreference = 'Stop'
+Import-Module vm.common -Force -DisableNameChecking
+
+try {
+
+  $category = 'Password Attacks'
```
mr-tz, line 6: lint complains that this also isn't used

day1player: `$category` needs to be outside of the try/catch block

packages/mimikatz.vm/tools/chocolateyinstall.ps1
```diff
@@ -0,0 +1,54 @@
...
+  }
+
+  Install-ChocolateyZipPackage @packageArgs
+
+  $shortcut = Join-Path $path "mimikatz.x86.lnk"
+  $target = Join-Path (Join-Path $toolsDir "Win32") "mimikatz.exe"
+  $target_cmd = Join-Path ${Env:WinDir} "system32\cmd.exe"
+  $target_args = '/K "' + $target + '"'
+  $target_icon = $target_cmd
+  $target_dir = Join-Path ${Env:UserProfile} "Desktop"
```
day1player, line 38: Keep in mind this variable when creating new packages.. For example, this will place the user in the desktop when they click the lnk file. I think for this tool it is better to launch in the $toolsDir (x86/x64 respectively) folder in case people want to do things like load the mimikatz SSP or mimidrv.sys file.

packages/mimikatz.vm/tools/chocolateyinstall.ps1
```diff
@@ -0,0 +1,54 @@
...
+  Install-ChocolateyShortcut -shortcutFilePath $shortcut -targetPath $target_cmd -Arguments $target_args -WorkingDirectory $target_dir -IconLocation $target_icon
+  Install-BinFile -Name "mimikatz.x86" -Path $target
+
+  $shortcut = Join-Path $path "mimikatz.x64.lnk"
+  $target = Join-Path (Join-Path $toolsDir "x64") "mimikatz.exe"
+  $target_cmd = Join-Path ${Env:WinDir} "system32\cmd.exe"
+  $target_args = '/K "' + $target + '"'
+  $target_icon = $target_cmd
+  $target_dir = Join-Path ${Env:UserProfile} "Desktop"
```
day1player, line 47: Same as above, but x64 working directory

packages/mimikatz.vm/tools/chocolateyinstall.ps1
Menn1s: Makes sense, fixing both!

nos3curity:
@Menn1s, need to fix the category in the uninstall script

nos3curity:
Tested locally. Looks good.

nos3curity (approved):

nos3curity (approved):

PR #257, opened 2023-02-22
Add telnet package

geo-lit:
Added telnet package.

Chocolatey installs package in system32, which is reflected by the changes in the install script compared to default template.


mr-tz:
I assume this is for CommandVM usage. Is this a valuable tool for red teams?

geo-lit:
Yes it is!

mr-tz (approved):
LGTM

Ana06 (changes requested):
Should the package version be 0.9.0 and the linter fail because of that? From the wiki:
```
When a metapackage is named after and installs/configures a specific tool (for instance cygwin.vm that installs/configures cygwin) lock the dependency version (e.g., version="[3.2.0]") and use the locked version as the metapackage's version.
```

geo-lit:
@Ana06  I locked the version to 0.9.0 in the nuspec file `<dependency id="telnet" version="[0.9.0]"/>`, is there another place I should be locking it?

packages/telnet.vm/telnet.vm.nuspec
```diff
@@ -0,0 +1,13 @@
+<?xml version="1.0" encoding="utf-8"?>
+<package xmlns="http://schemas.microsoft.com/packaging/2015/06/nuspec.xsd">
+  <metadata>
+    <id>telnet.vm</id>
+    <version>0.0.0.20230222</version>
```
mr-tz, line 5:
```suggestion
    <version>0.9.0</version>
```

mr-tz: `lint.py` fails to catch this... 😞

mr-tz:
The `windows-2022` CI run is still not happy :(

Ana06:
Documenting possible solutions discussed privately:
- Increase the Sleep to 20 seconds
- Do not use the telnet dependency and instead use:
  `Start-Process -FilePath $pkgmgr -Argumentlist '/iu:"TelnetClient"' -Wait`
Both options work, but the second is more elegant.

@geo-lit let me know if you need help to address the changes or if you don't have time (and then I'll address them).

Ana06:
Also, @geo-lit it would be very appreciated if you could squash the commits to keep a clean commit history (I can help you if you are not familiar with this). Alternatively, we can merge the PR with the squash and merge option (one commit for the whole PR).

Good reference: https://about.gitlab.com/blog/2018/06/07/keeping-git-commit-history-clean

geo-lit:
It seems like the checks are passing. Please let me know how you would like me to squash the commits

PR #263, opened 2023-03-14
Add Python3 metapackage and update library package

MalwareMechanic:
Will help with #241

### DIFF
--- a/packages/libraries.python3.vm/libraries.python3.vm.nuspec
+++ b/packages/libraries.python3.vm/libraries.python3.vm.nuspec
@@ -2,13 +2,13 @@
 <package xmlns="http://schemas.microsoft.com/packaging/2015/06/nuspec.xsd">
   <metadata>
     <id>libraries.python3.vm</id>
-    <version>0.0.0.20230118</version>
+    <version>0.0.0.20230314</version>
     <description>Metapackage to install common Python3 libraries</description>
     <authors>Mandiant, Microsoft, Python Software Foundation</authors>
     <dependencies>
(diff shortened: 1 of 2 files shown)


MalwareMechanic:
CI test failures due to Python3 version already being installed, okay to overlook
> Installing 64-bit python3...
WARNING: Generic MSI Error. This is a local environment error, not an issue with a package or the MSI itself - it could mean a pending reboot is necessary prior to install or something else (like the same version is already installed). Please see MSI log if available. If not, try again adding '--install-arguments="'/l*v c:\python3_msi_install.log'"'. Then search the MSI Log for "Return Value 3" and look above that for the error.
ERROR: Running ["C:\ProgramData\chocolatey\lib\python3\tools\python-3.10.0-amd64.exe" /quiet InstallAllUsers=1 PrependPath=1 TargetDir="C:\Python310" ] was not successful. Exit code was '1603'. Exit code indicates the following: Generic MSI Error. This is a local environment error, not an issue with a package or the MSI itself - it could mean a pending reboot is necessary prior to install or something else (like the same version is already installed). Please see MSI log if available. If not, try again adding '--install-arguments="'/l*v c:\python3_msi_install.log'"'. Then search the MSI Log for "Return Value 3" and look above that for the error..
The install of python3 was NOT successful.

MalwareMechanic:
Got hasty. I added it to the test exclusion list only to realize that since it's a dependency it'll still need to not fail for other packages that depend on it. Will look into and fix. In the mean time I'm reverting libraries.python3.vm to use it's old dependencies.

PR #264, opened 2023-03-14
Add python3.vm to test exclusion list

MalwareMechanic:


mr-tz:
Let's add a comment on why we don't want to test the excluded packages.

PR #269, opened 2023-03-19
Add the Cloud category

nos3curity:
Adding a new tool category for cloud-specific tools

### DIFF
--- a/.github/ISSUE_TEMPLATE/new_metapackage.yml
+++ b/.github/ISSUE_TEMPLATE/new_metapackage.yml
@@ -51,6 +51,7 @@ body:
       description: Which category should this tool be installed to?
       options:
         - Android
+        - Cloud
         - Debuggers
         - Delphi
         - Disassemblers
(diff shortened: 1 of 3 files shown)


mr-tz (approved):

mr-tz:
@Ana06, thoughts on this?

mr-tz:
This sounds like subcategories would be nice.

Ana06:
I like the idea of a Cloud category. Not sure if we need as many subcategories. Taking into account we don't even have a Cloud category at the moment, it may be enough with adding that. Any other opinions @vm-packages? Do you have any preference @nos3curity?

nos3curity:
@Ana06 We can keep it generic as "Cloud". I don't foresee us having an overwhelming amount of packages in that category to the point where we need to break it up. Exploitation will have a lot more. Ready to merge this if you are good with that.

day1player:
Agreed just one for Cloud, lots of tools popping up in that category for us lately

PR #271, opened 2023-03-20
Add Azure-CLI

nos3curity:
added a tool to manage azure resources from command-line.

this metapackage doesn't have a shim, so the shortcut calls cmd.exe with `az --help` instead.

this PR is blocked until #269 is merged

### DIFF
--- a/packages/azure-cli.vm/tools/chocolateyinstall.ps1
+++ b/packages/azure-cli.vm/tools/chocolateyinstall.ps1
@@ -1,0 +1,16 @@
+$ErrorActionPreference = 'Stop'
+Import-Module vm.common -Force -DisableNameChecking
+
+try {
+  $toolName = 'Azure-CLI'
+  $category = 'Cloud'
+
+  $shortcutDir = Join-Path ${Env:TOOL_LIST_DIR} $category
+  $shortcut = Join-Path $shortcutDir "$toolName.lnk"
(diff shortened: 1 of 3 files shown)


nos3curity:
Not sure why the checks are failing. This is a metapackage. The only thing different here from the one hosted on Chocolatey is the shortcut path, which isn't the component failing. Also confirmed to work on my Windows 11 install.

packages/azure-cli.vm/tools/chocolateyinstall.ps1
day1player, line 1: If there is no shim created is this the best way to create the shortcut? After clicking the shortcut users will simply see the help menu and then when they try to run `az` again will it not work? Perhaps we should look at installing a shim?

nos3curity: It will keep the terminal open for the user and they may run `az` again and it will work. Would that be a good approach to this?

day1player: Oh ok, thats probably fine then. What working directory are you dropped into with the shortcut? Is it system32?

day1player: If so, might consider the `-WorkingDirectory` option with `Install-ChocolateyShortcut`.. Probably user's Desktop?

nos3curity:
On second thought, the shortcut is not necessary for this package. We will add it directly from chocolatey instead of using a metapackage. Closing this.

PR #272, opened 2023-03-20
Add Azure PowerShell 

nos3curity:
Added the Azure PowerShell module.

The metapackage doesn't have a shim, therefore we are calling PowerShell and listing the current available Azure contexts.

This PR is blocked until #269 is resolved.

### DIFF
--- a/packages/azure-powershell.vm/tools/chocolateyinstall.ps1
+++ b/packages/azure-powershell.vm/tools/chocolateyinstall.ps1
@@ -1,0 +1,16 @@
+$ErrorActionPreference = 'Stop'
+Import-Module vm.common -Force -DisableNameChecking
+
+try {
+  $toolName = 'Azure PowerShell'
+  $category = 'Cloud'
+
+  $shortcutDir = Join-Path ${Env:TOOL_LIST_DIR} $category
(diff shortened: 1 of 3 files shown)


packages/azure-powershell.vm/tools/chocolateyinstall.ps1
```diff
@@ -0,0 +1,16 @@
...
+
+try {
+  $toolName = 'Azure PowerShell'
+  $category = 'Cloud'
+
+  $shortcutDir = Join-Path ${Env:TOOL_LIST_DIR} $category
+  $shortcut = Join-Path $shortcutDir "$toolName.lnk"
+  $executablePwsh  = Join-Path ${Env:WinDir} "System32\WindowsPowerShell\v1.0\powershell.exe" -Resolve
+  $executableArgs = "-NoExit -Command Get-AzContext -ListAvailable"
+  Install-ChocolateyShortcut -shortcutFilePath $shortcut -targetPath $executablePwsh -Arguments $executableArgs -RunAsAdmin
```
day1player, line 12: I think the shortcut should import the module first no?

nos3curity:
On second thought, the shortcut is not necessary for this package. We will add it directly from chocolatey instead of using a metapackage. Closing this.

PR #273, opened 2023-03-20
Add AzureHound

nos3curity:
azurehound is the azure collector for bloodhound

this PR is blocked until #269 is resolved

### DIFF
--- a/packages/azurehound.vm/azurehound.vm.nuspec
+++ b/packages/azurehound.vm/azurehound.vm.nuspec
@@ -1,0 +1,12 @@
+<?xml version="1.0" encoding="utf-8"?>
+<package xmlns="http://schemas.microsoft.com/packaging/2015/06/nuspec.xsd">
+  <metadata>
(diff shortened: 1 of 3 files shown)


day1player:
Version 2.0.1 was released 3 hours ago -_-

mr-tz:
If this is a standard GitHub install our auto-updates can fix this :)

Ana06 (commented):
I don't know much about Azure, but aren't we missing dependencies here as it happens with https://github.com/mandiant/VM-Packages/pull/277? 🤔 

As @mr-tz says, the [automatic updates GH](https://github.com/mandiant/VM-Packages/blob/main/.github/workflows/update_package.yml) will updates the GH release after the PR is merged. We run it automatically once a week. 😄

day1player:
Since this is a compiled release there should be no dependencies @Ana06

nos3curity:
@day1player @Ana06 confirmed, no dependencies here. It's a compiled binary, not a Python tool like #278

PR #277, opened 2023-03-20
Add AzureADRecon

nos3curity:
AzureADRecon is a tool which extracts and combines various reconnaissance artifacts out of an Azure AD environment with a valid credential.

This PR is blocked until @269 is resolved.

### DIFF
--- a/packages/azureadrecon.vm/tools/chocolateyinstall.ps1
+++ b/packages/azureadrecon.vm/tools/chocolateyinstall.ps1
@@ -1,10 +1,20 @@
 $ErrorActionPreference = 'Stop'
 Import-Module vm.common -Force -DisableNameChecking
 
+Try {
+    Install-Module -Name Microsoft.Graph -ErrorAction Stop
+}
+Catch {
+    Write-Host "Failed to install Microsoft.Graph module. Error: $_"
+    Exit
+}
+
 $toolName = 'AzureADRecon'
-$category = 'Cloud'
+$category = 'Reconnaissance'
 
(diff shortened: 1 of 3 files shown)


packages/azureadrecon.vm/tools/chocolateyinstall.ps1
```diff
@@ -0,0 +1,10 @@
+$ErrorActionPreference = 'Stop'
+Import-Module vm.common -Force -DisableNameChecking
+
+$toolName = 'AzureADRecon'
+$category = 'Cloud'
+
+$zipUrl = 'https://github.com/adrecon/AzureADRecon/archive/18410c1a630c085091ad15a768ac742fd2513902.zip'
+$zipSha256 = '51e0af14ae0e35300fe6fe8fa96ebbce86abb34a597fe81fd16b7358ebe13aae'
+
+VM-Install-Raw-GitHub-Repo $toolName $category $zipUrl $zipSha256
```
day1player, line 10: This project has some prerequisites that we can probably handle, such as these:
```
Install-Module -Name AzureAD
Install-Module -Name Microsoft.Graph
```
I would add these commands in a try catch here so that we can get the prereq's installed

day1player: Optionally we can look for chocolatey packages that do this already and add dependencies to the nuspec file

day1player:
@nos3curity bump

nos3curity:
this package depends on a deprecated powershell module. closing this.

PR #278, opened 2023-03-20
Add ADConnectDump

nos3curity:
This toolkit offers several ways to extract and decrypt stored Azure AD and Active Directory credentials from Azure AD Connect servers.

This PR is blocked until @269 is resolved.

### DIFF
--- a/packages/adconnectdump.vm/adconnectdump.vm.nuspec
+++ b/packages/adconnectdump.vm/adconnectdump.vm.nuspec
@@ -7,6 +7,7 @@
     <description>This toolkit offers several ways to extract and decrypt stored Azure AD and Active Directory credentials from Azure AD Connect servers.</description>
     <dependencies>
       <dependency id="common.vm" />
+      <dependency id="python2" />
     </dependencies>
   </metadata>
 </package>
--- a/packages/adconnectdump.vm/adconnectdump.vm.nuspec
+++ b/packages/adconnectdump.vm/adconnectdump.vm.nuspec
@@ -1,0 +1,12 @@
+<?xml version="1.0" encoding="utf-8"?>
+<package xmlns="http://schemas.microsoft.com/packaging/2015/06/nuspec.xsd">
+  <metadata>
+    <id>adconnectdump.vm</id>
+    <version>0.0.0.20230320</version>
+    <authors>fox-it</authors>
+    <description>This toolkit offers several ways to extract and decrypt stored Azure AD and Active Directory credentials from Azure AD Connect servers.</description>
+    <dependencies>
+      <dependency id="common.vm" />
+    </dependencies>
+  </metadata>
+</package>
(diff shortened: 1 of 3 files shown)


packages/adconnectdump.vm/tools/chocolateyinstall.ps1
```diff
@@ -0,0 +1,10 @@
+$ErrorActionPreference = 'Stop'
+Import-Module vm.common -Force -DisableNameChecking
+
+$toolName = 'ADConnectDump'
+$category = 'Cloud'
```
Ana06, line 5: As in https://github.com/mandiant/VM-Packages/pull/253, this is another category we don't have at the moment (see https://github.com/mandiant/VM-Packages/issues/38). We need a linter and further discussion on this.

Ana06 (commented):
This looks great, thank @nos3curity. We need some discussion about the categories before merging this. We can add new categories, but the problem is we can't change a package of category (as this break for example updates) so it needs to be a meditated/agreed decision. Let's discuss this in #269

day1player:
We need to discuss how we are going to handle Python tools. Since ADConnectDump is a python tool, should we have Python as a dependency?

mr-tz:
Yes, I'd add the needed Python version as dependency.

nos3curity:
Added dependency. Should be ready to merge.

packages/adconnectdump.vm/adconnectdump.vm.nuspec
```diff
@@ -0,0 +1,13 @@
+<?xml version="1.0" encoding="utf-8"?>
+<package xmlns="http://schemas.microsoft.com/packaging/2015/06/nuspec.xsd">
+  <metadata>
+    <id>adconnectdump.vm</id>
+    <version>0.0.0.20230320</version>
+    <authors>fox-it</authors>
+    <description>This toolkit offers several ways to extract and decrypt stored Azure AD and Active Directory credentials from Azure AD Connect servers.</description>
+    <dependencies>
+      <dependency id="common.vm" />
+      <dependency id="python2" />
```
mr-tz, line 10: Does this work with Python3? We're thinking about removing Py2 (see #301).

day1player: It does not! :(

Ana06: I think this is ok for tools that don't work on python 3.

Ana06 (approved):

packages/adconnectdump.vm/adconnectdump.vm.nuspec
Ana06: I have just realised that there is an issue with this as it mess up the path: https://github.com/mandiant/VM-Packages/issues/337

PR #288, opened 2023-04-14
Revert ":robot: Package update"

mr-tz:
Reverts mandiant/VM-Packages#287

CI failed on merge


Ana06:
The problem is that sysinternals (the dependency) uses a url that doesn't include the version: https://download.sysinternals.com/files/SysinternalsSuite.zip A new update to this url broke the package after vm-packages sent the PR. But this break both the new and the old package. The community package has already released a new version to fix this. So I have trigger the package update manually in https://github.com/mandiant/VM-Packages/actions/runs/4721610932. The resulting PR https://github.com/mandiant/VM-Packages/pull/289 fixes the issue.

mr-tz:
Great, thanks!

Issue #270, opened 2023-03-20
Lint category


mr-tz:
The categories are fixed and there is no Email category. See discussion in https://github.com/mandiant/VM-Packages/issues/38

@mr-tz should we add a category check to the linter?

_Originally posted by @Ana06 in https://github.com/mandiant/VM-Packages/pull/253#discussion_r1140315223_

Ana06:
We should also find a way to only have categories in one place 👍

Ana06:
I can move the categories to a file and use this file in the linter and the script to generate more packages. But I am not sure how to avoid the duplication with the issue templates. Any ideas @mr-tz ?

mr-tz:
Not really, we should document where the categories need to be updated. I hope it's less of a problem since we don't plan to change them often.

PR #291, opened 2023-04-18
Lint categories

Ana06:
Prevent use of custom (non defined) categories. Modifying the category a package belongs to is a breaking change (breaks updates and uninstalling), so we have fixed the categories and adding new categories needs to be discussed.

Some packages don't have a category (we don't create a link in the tools directory) and need to be ignored when linting.

Note we still need to keep the issue templates in sync manually, but we don't expect this to be a big issue as we don't plan to change them often.

I have added some documentation in https://github.com/mandiant/VM-Packages/wiki/Coding-Conventions/_compare/4fc1c6ac77f6d60d0639c81f8d23c31871fb50d4...9592e06e72738632d04c6a7be9cf0acbfa656f1f Is there any other place we need to document this?

Closes https://github.com/mandiant/VM-Packages/issues/270

### DIFF
--- a/categories.txt
+++ b/categories.txt
@@ -1,0 +1,28 @@
+Android
+Cloud
+Debuggers
+Delphi
+Disassemblers
+dotNet
+Forensic
+Hex Editors
+Java
+Javascript
+Networking
+Office
+PDF
+PE
+PowerShell
+Python
+Text Editors
+Utilities
+VB
+Active Directory
+Command & Control
+Evasion
+Exploitation
+Information Gathering
+Password Attacks
+Vulnerability Analysis
+Web Application
+Wordlists
--- a/packages/cmder.vm/cmder.vm.nuspec
+++ b/packages/cmder.vm/cmder.vm.nuspec
@@ -2,7 +2,7 @@
 <package xmlns="http://schemas.microsoft.com/packaging/2015/06/nuspec.xsd">
   <metadata>
     <id>cmder.vm</id>
-    <version>1.3.20.20221201</version>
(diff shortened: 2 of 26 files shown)


Ana06:
@mr-tz where should we document where the categories need to be updated? What need to be updated to change/add a category is the `categories.txt` file and the issue templates (as we can't read the file from the templates).

mr-tz (approved):

categories.txt
```diff
@@ -0,0 +1,28 @@
+Android
```
mr-tz, line 1: Can we add a comment here where else to update this?

Ana06: The problem with this option is that we have to take this into account when reading the file (it makes the code to read the file a bit more complicated). Or can you think on an easy way to do this?

Ana06: I was thinking about adding a check to the CI that ensure the issue templates and this file are in sync. But I was not sure if the effort is worth as we don't plan to update the categories often. This is not an issues this PR introduces as we have the problem already. What about merging this PR and discuss how to ensure categories are up-to-date in an issue @mr-tz ? (I can take care of opening the issue)

Ana06:
We should be able to automatically fix CyberChef after #295 has been merged, which should fix the CI here.

categories.txt
mr-tz: Sounds good! Let's keep it simple, but happy to discuss further separately.

Ana06: Issue created: https://github.com/mandiant/VM-Packages/issues/298

PR #295, opened 2023-04-19
Improve package updates automation

Ana06:
In case a package had GitHub release and dependencies updates, only the GitHub release was updated. The ideas was to avoid updating many things at once. But this is causing problems with packages (like CyberChef with GoogleChrome as dependency) whose dependencies use download links that break often (like the ones that don't include the version). In this case, any update to the package which doesn't update the failing dependency will make the test fail.

Update and test dependencies and GitHub url independently. This way we ensure we update as much as we can and prevent the described problem.

In addition, catch the `FileNotFoundError` exception caused by metapackages that don't have `chocolateyinstall.ps1` when trying to read this file to update it. This exception is not breaking our CI, but it displayed in the logs.

**After this PR, we will be able to update CyberChef and Python3**

### DIFF
--- a/.github/workflows/update_package.yml
+++ b/.github/workflows/update_package.yml
@@ -25,24 +25,31 @@ jobs:
           New-Item test_logs -itemType Directory | Out-Null
           foreach ($packagePath in (Get-ChildItem packages)){
             $package = $packagePath.Name
-            $version = python scripts\utils\update_package.py $package
-            $updated = $?
-            echo "$package $version"
-            if ($updated -and $version) {
-              # Test package before committing
-              scripts\test\test_install.ps1 $package *>> "test_logs/$package.txt"
-              $tested = $?
-              cd $root
-              if ($tested) {
-                git add "packages/$package/"
-                git commit -m "Update $package to $version"
-              } else {
-                echo "$package $version FAILED"
+            $newVersion = 0
+            # Test indepdendly every type of update and commit what works
+            foreach ($UPDATE_TYPE in ('DEPENDENCIES', 'GITHUB_URL')) {
+              $version = python scripts\utils\update_package.py $package --update_type $UPDATE_TYPE
+              $updated = $?
+              echo "$package $version"
+              if ($updated -and $version) {
+                # Test package before committing
+                scripts\test\test_install.ps1 $package *>> "test_logs/$package.txt"
+                $tested = $?
+                cd $root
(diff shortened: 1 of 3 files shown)


.github/workflows/update_package.yml
```diff
@@ -25,24 +25,31 @@ jobs:
...
-              cd $root
-              if ($tested) {
-                git add "packages/$package/"
-                git commit -m "Update $package to $version"
-              } else {
-                echo "$package $version FAILED"
+            $newVersion = 0
+            # Test indepdendly every type of update and commit what works
+            foreach ($UPDATE_TYPE in ('DEPENDENCIES', 'GITHUB_URL')) {
+              $version = python scripts\utils\update_package.py $package --update_type $UPDATE_TYPE
```
mr-tz, line 31: For a potential follow up, would it make sense to turn this into a Python script or add to `update_package`?

scripts/test/test_install.ps1
```diff
@@ -44,7 +44,9 @@ Set-Location $built_pkgs_dir
 $failed = 0
 $success = 0
 foreach ($package in $built_pkgs) {
-    choco install $package -y -r -s "'.;https://www.myget.org/F/vm-packages/api/v2;https://community.chocolatey.org/api/v2/'" --no-progress --force
+    # install looks for a nuspec with the same version as the installed one
+    # upgrade install the last found version (even if the package is not installed)
```
mr-tz, line 48:
```suggestion
    # upgrade installs the last found version (even if the package is not installed)
```

mr-tz (approved):
LGTM, thanks! One suggestion on grammar.

.github/workflows/update_package.yml
Ana06: not sure I understand what you mean. There is already a way to run `update_package.py` with both update types (note the definition of `ALL = DEPENDENCIES | GITHUB_URL`). But we need to run them separately here to ensure we can commit only the part which passes the tests.

mr-tz: Sorry, I mean to do the surrounding code in Python instead of PS.

Ana06: We could rewrite the surrounding code in Python but not sure if we win much. We still need to call the test powershell script. 🤔 Let's merge this (as this is merging the linter of categories) and continue the discussion afterwards.

Issue #299, opened 2023-04-20
Unify way we display errors in different GH actions


Ana06:
# Current approach
There are 3 actions that install packages and upload/display differently the errors. We should decide one way to do it and do it in the same way in the three actions. This are the different ways we currently use:

## 1. Upload choco & VM logs as artifacts in one zip ([`ci.yml`](https://github.com/mandiant/VM-Packages/blob/main/.github/workflows/ci.yml))

Upload the following files as artifacts (one zip):
```
C:\ProgramData\chocolatey\logs\chocolatey.log
C:\ProgramData\chocolatey\lib-bad\**\tools\install_log.txt
C:\ProgramData\_VM\log.txt
```

I like that it is simple, but it creates too many folders in the zip if all the files are present because the **least common ancestor of all the search paths is used as the root directory** .

## 2. Display VM log in the run ([`daily.yml`](https://github.com/mandiant/VM-Packages/actions/workflows/daily.yml))

In addition to upload the artifacts, [`daily.yml`](https://github.com/mandiant/VM-Packages/actions/workflows/daily.yml) also displays `C:\ProgramData\_VM\log.txt` during the run with:

```
if (Test-Path -Path C:\ProgramData\_VM\log.txt ) { Select-String "ERR|WARN" -CaseSensitive -Context 0,5 C:\ProgramData\_VM\log.txt }
```

I personally always download the log as artifact and don't use this view as I need more clicks to open it. @mr-tz is there a reason you prefer this vs just uploading the file (asking as you introduce this)?

## 3. Upload log per package ([update_package.yml](https://github.com/mandiant/VM-Packages/actions/workflows/update_package.yml)
In , as we are installing every package individually and I wanted to remove the noise output from the log, I created a log per package and upload all of them as zip.

```
scripts\test\test_install.ps1 $package *>> "test_logs/$package.txt"
```

I like that it is easy to see the run of a single package, but it includes more information that the failures. That has the advantage that you see all the information and the disadvantage that you have too much information. I think in general using `C:\ProgramData\_VM\log.txt` for debugging errors makes it easier.

# 4. Proposal: Upload choco & VM logs as artifacts in two zips

Upload one zip with the following files:
```
C:\ProgramData\chocolatey\logs\chocolatey.log
C:\ProgramData\chocolatey\lib-bad\**\tools\install_log.txt
```
Upload a second zip only with the file (note it is not possible to upload a file without zip):
```
C:\ProgramData\_VM\log.txt
````

[...]

PR #302, opened 2023-04-21
Unify way of displaying errors in GH actions

Ana06:
There are 3 workflows that install packages and upload/display differently the errors. Upload choco & VM logs as artifacts in two zips for all workflows. Use [composite actions](https://docs.github.com/en/actions/creating-actions/creating-a-composite-action) to avoid repeating code and ensuring consistency.

You can see it working in https://github.com/Ana06/VM-Packages/pull/29

Closes https://github.com/mandiant/VM-Packages/issues/299

### DIFF
--- a/.github/actions/upload-logs/action.yml
+++ b/.github/actions/upload-logs/action.yml
@@ -1,0 +1,20 @@
+name: Update logs
+description: Update test install logs to artifacts
+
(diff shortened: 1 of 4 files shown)


mr-tz (approved):
very cool! love the unification.

PR #304, opened 2023-04-24
Update pestudio.vm

mr-tz:


mr-tz:
Manually updating this is cumbersome. The page https://www.winitor.com/download2 lists the SHA256. We could add it to the update script and/or ask the developer to provide version specific links.

Ana06:
> We could add it to the update script and/or ask the developer to provide version specific links.

I was thinking about this, but I don't really like the idea of having code that we use for just one package. Or are there other packages where we could use something like this?

Two other ideas:
- Reaching out to the maintainer of the tool asking to include the version in the link (and keep the link for a bit longer)
- Automatically update hashes that fail, but if we go this way, I wonder if it would be better to just ignore the hash for this kind of packages.

PR #305, opened 2023-04-25
Add party emoji to daily run log

Ana06:
It is now realistic that we have 0 failures! Celebrate it with a :partying_face: in the log


mr-tz (approved):
lol

Issue #311, opened 2023-04-26
Should we retry installing in the daily run?


Ana06:
Yesterday regshot failed in one of the daily runs and installed correctly in the other. This is the error message:

```
2023/04/26 04:39:00 [regshot.vm] vm.common.psm1 [+] ERROR : [ERR] The remote file either doesn't exist, is unauthorized, or is forbidden for url 'https://sourceforge.net/projects/regshot/files/regshot/1.9.1-beta/Regshot-1.9.1-beta_r321.7z'. Exception calling "GetResponse" with "0" argument(s): "The request was aborted: Could not create SSL/TLS secure channel."
At C:\ProgramData\chocolatey\helpers\functions\Get-WebFile.ps1:330 char:8
+        throw "The remote file either doesn't exist, is unauthorized,  ...
+        ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
```

It seems the url was not available for some reason. Or is there other reason for this failure?

What can we do about this? We could try to install failing packages a second time and only mark them as fail if they fail twice. Any other ideas?

Ana06:
It seems Install-Choco may be an easy option: https://github.com/actions/runner-images/pull/721

PR #317, opened 2023-04-27
Reinstall packages to prevent transient failures

Ana06:
We often get in the daily run the error:
`The remote file either doesn't exist, is unauthorized, or is forbidden for url`

Reinstall a package when the installation fails to prevent transient failures in tests. Reinstall twice by default to avoid delaying test runs (locally and in PRs) too much. Run it three times in the daily runs as time is not an issue in this case and the transient failures are very annoying in the logs.

Similar approach as in: https://github.com/actions/runner-images/pull/721

Closes https://github.com/mandiant/VM-Packages/issues/311

### DIFF
--- a/.github/workflows/daily.yml
+++ b/.github/workflows/daily.yml
@@ -21,7 +21,7 @@ jobs:
         uses: actions/checkout@v3
       - name: Build and test all packages
         id: test
-        run: scripts/test/test_install.ps1 -all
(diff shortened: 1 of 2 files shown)


Ana06:
merging without review so that we can test over the weekend how well this work in the daily run. @mr-tz please review it when you have time and we can address any feedback you have in an extra PR. 😄

mr-tz:
Looks good!

PR #319, opened 2023-04-28
adding KeePass

skollr34p3r:


Ana06 (commented):
Can you please explain why this is a tool that you want to have as a `.vm` package vs just install it using the community package? It doesn't seem to be an analysis tool.

day1player:
> Can you please explain why this is a tool that you want to have as a `.vm` package vs just install it using the community package? It doesn't seem to be an analysis tool.

@Ana06 We are looking to add some of these packages with their dependencies mainly because we are looking at reading package metadata through the myget API with our installer. If we have a package like this for example, that's a metapackage, we can pull down the description and other properties from MyGet so that when people go to create a custom install they can decide whether they want to add it to their install or not a lot easier. Also in packages like these some of them we will want to so simply because we want to have an LNK created in the tools folder so that it is easier for users to find the tool

Ana06:
> @Ana06 We are looking to add some of these packages with their dependencies mainly because we are looking at reading package metadata through the myget API with our installer. If we have a package like this for example, that's a metapackage, we can pull down the description and other properties from MyGet so that when people go to create a custom install they can decide whether they want to add it to their install or not a lot easier.

You can directly use `keepass` (instead of `keepass.vm`) in the installer. 

> Also in packages like these some of them we will want to so simply because we want to have an LNK created in the tools folder so that it is easier for users to find the tool

This is installed in Utilities, which I think doesn't make it easier to find. 

---

Every package we add to the repository needs to be maintained and increase testing time. I think we should only add non analysis tools as vm packages if there is no community package or if there is a very good reason.

nos3curity:
@day1player @geo-lit @Menn1s We have dozens of tools on the CommandoVM roadmap that are available already in the Chocolatey community repository. For us to add these packages to the install directly from chocolatey would take no more than 5 minutes. The downside is obviously that their shortcuts will not be categorized. 

On the flipside, if we were to make a metapackage for each one of them, that will be dozens of additional PRs for us to go through and then maintain going forward. 

To me as a user, category shortcuts within CommandoVM are helpful because I may not be familiar with every tool that is available to me for a given phase of assessment. I personally wouldn't benefit from having Google Chrome, VS Code, golang, etc categorized. I would want to see confuserex2 categorized under evasion though (it also is an existing choco package).

day1player:
@nos3curity completely agreed, I think we should use our own discretion for creating the shortcuts. How do you feel about KeePass? I dont think this one needs a shortcut.

nos3curity:
@day1player I agree that a categorized shortcut isn't necessary for KeePass

nos3curity:
Closing per Commando team's conversation today and the earlier comments.

Issue #301, opened 2023-04-21
why do we need a python3 package?


Ana06:
python3.vm is just installing the community package, there is no even chocolateyinstall.ps1 file. Why do we need this package?

mr-tz:
Good question :) In another thread we talked about adding other default packages. If we come up with a good solution to also install existing choco packages we could save a lot of overhead duplicating them.

Ana06:
There is already a way to do this, adding them to the/your [config file](https://github.com/mandiant/flare-vm/blob/main/config.xml). In the case of python3, it would be a dependency for `libraries.python3.vm`. Or am I missing something?

Ana06:
I understand we can get ride of this, is that right @mr-tz ?

mr-tz:
Yeah, do we have to add it to the profile or something so Py3 gets installed then?

Ana06:
I think python 3.9 comes installed by default in Windows 10, but the package libraries.python3.vm installs it anyway.

PR #331, opened 2023-05-16
Remove python3.vm

Ana06:
This packages only install the dependency, then we can remove this package and use the dependency instead. In fact, we are already using the dependency directly in libraries.python3.vm. After merging, we should remove this package from MyGet.

Closes https://github.com/mandiant/VM-Packages/issues/301


mr-tz (approved):
all good, never mind my other comment in the issue

MalwareMechanic:
Ah, I made this package because I was testing why the "python3" dependency was failing in the CI pipeline, see: https://github.com/mandiant/VM-Packages/pull/263

Since the GH VMs already have Python3 installed, the package was failing. I was trying to find a way around this... but didn't have enough time to get the issue ironed out. Sorry for the confusion.

Ana06:
@MalwareMechanic can we just add python 3.9 as dependency in the packages that require python? This is what we currently do in libraries.python3.vm and it works in the GH VMs.

We could document somewhere to use Python 3.9 when possible.

Issue #308, opened 2023-04-25
Add python dependencies


Ana06:
We need the following python 3 libraries for FLARE-EDU:
- pathlib
- psutil
- art

Is there something agains adding them to the current libraries.python3.vm package?

Do we want to continue trying to keep python3 and 2 in sync? Or do we even want to deprecate/remove python2? Is there any reason to keep it?

mr-tz:
Python2 was useful for FakeNet-NG but that has now also been updated to Py3. So I'm fine removing it.
Adding those libraries sounds good.

Ana06:
@mr-tz fakenet is not part of the libraries.python2.vm package https://github.com/mandiant/VM-Packages/blob/main/packages/libraries.python2.vm/tools/modules.xml The fakenet package uses the .exe from the GH release: https://github.com/mandiant/VM-Packages/blob/main/packages/fakenet-ng.vm/tools/chocolateyinstall.ps1#L14

Ana06:
pathlib is included in python 3.9, so we don't need to add it.

PR #332, opened 2023-05-16
Remove libraries.python2.vm

Ana06:
I think it is time to completely get rid of Python 2. Having only one libraries.python package saves as some work as we don't have to try to keep them in sync. After merging, we should remove this package from MyGet.

This has been discussed in https://github.com/mandiant/VM-Packages/issues/308


Ana06:
This is not related to #278. I want to get rid of our python2 libraries package as we hav the same package for python3. Installing libraries that require python2 is still possible the python2 metapackage as #278 does. Although this should be only done if the tool does not work on python3.

Issue #250, opened 2023-02-15
Add right-click for any file extension: CFFExplorer/HxD/010Editor


mr-tz:

Ana06:
This has been annoying me for some time as well 👍

mr-tz:
What does "if 010 Editor can open the file" mean? Can that open files without extensions or .exe_ etc.?

PR #351, opened 2023-05-24
Improve/fix right click menu

Ana06:
- Support icons in the right click menu by adding an argument to the `VM-Add-To-Right-Click-Menu` helper function
- Add icon to the right click menu entries of DIE and HashMyFiles
- Fix HashMyFiles right click menu command.
- Add CFF explorer (for all extensions) and HxD to the right click menu.

Screenshot 2023-05-24 at 22 33 22

Closes https://github.com/mandiant/VM-Packages/issues/250

@mr-tz @emtuls some extra manual testing would be good if you have time to ensure this works as everybody expects. 😉


mr-tz (approved):
looks good, but I did not test this

PR #347, opened 2023-05-24
:robot: Add net-reactor-slayer.vm

vm-packages:
Automated PR from https://github.com/mandiant/VM-Packages/issues/342 :rocket:

PR #353, opened 2023-05-25
Ensure template follows version documentation

Ana06:
If the dependency/tool's version uses the 4th segment, update the package's version to use the current date (YYYYMMDD) in the 4th segment. We had this problem with https://github.com/mandiant/VM-Packages/pull/347 This PR should allows us to use the integration next time.


mr-tz (approved):

mr-tz:
cool! thanks!

emtuls (approved):

Issue #371, opened 2023-05-29
Package proposal: situational-awareness-bofs.vm


nos3curity:
### Package Name

situational-awareness-bofs

### Tool Name

Situational Awareness BOF

### Package type

GITHUB_REPO

### Tool's version number

0.0.0.20230529

### Category

Command & Control

### Tool's authors

trustedsec

### Tool's description

Provides a set of basic situational awareness commands implemented in a Beacon Object File (BOF). This allows you to perform some checks on a host before you begin executing commands that may be more invasive.

### Download URL

https://github.com/trustedsec/CS-Situational-Awareness-BOF/archive/refs/heads/master.zip

### Download SHA256 Hash

e3673d7e41ad6d36ca7d6d44821f68238aae9968e062acb6d96fc7663c87bbdb

### Why is this tool a good addition?

This is an incredibly useful set of BOFs for CommandoVM

PR #372, opened 2023-05-29
:robot: Add situational-awareness-bofs.vm

vm-packages:
Automated PR from https://github.com/mandiant/VM-Packages/issues/371 :rocket:


mr-tz (approved):

day1player:
similar to #383 should we name this with the tool author? ie "trustedsec-situational-awareness-bof"

nos3curity:
@day1player changed the tool category. I think #383 being "c2-collection" is too generic, needing us to specify whose collection it is, but in this case, it is specific and known enough under this name that it isn't necessary in my opinion. Thoughts?

day1player:
Yeah I think you're right. I noticed the nuspec file is still titled with an "s", I think after that is fixed we are good to merge

nos3curity:
@day1player fixed. Ready to merge.

PR #328, opened 2023-05-16
Adding Sliver client

Menn1s:
The Sliver client is used as part of an adversary emulation framework. The framework is being used in red team training classes and is a viable alternative to the closed-source Cobalt Strike adversary emulation framework.

Source:
https://github.com/BishopFox/sliver

PR #373, opened 2023-05-29
Fixed linter regex

nos3curity:
per PRs #328 and #372, the linter had faulty regex that didn't approve of the "Command & Control" category.

thanks to @mr-tz for the fix suggested in #328


nos3curity:
tested the fix suggested by mr-tz. merging myself, since it's a minor one character fix to unblock the PRs

Issue #412, opened 2023-06-05
dnspyex.vm: community package does not install both 64 and 32 versions


Ana06:
@emtuls
> I believe that in order to properly debug a 32 bit .NET file, you need to use the 32 bit version, and for a 64 bit .NET, you would need the 64 bit version.

The community package we are using does not install both 64 and 32 versions. Thanks for noticing this @emtuls! I propose we install dnspyex directly to fix this.

Ana06:
I have tried to install this manually using the zips from https://github.com/dnSpyEx/dnSpy/releases/tag/v6.3.0, the log says that the zips are unzipped sucessfully, but the 32 folder is empty:

```
Downloading dnspyex.vm
  from 'https://github.com/dnSpyEx/dnSpy/releases/download/v6.3.0/dnSpy-net-win32.zip'

Download of dnSpy-net-win32.zip (81.9 MB) completed.
Hashes match.
Extracting C:\Users\flare\AppData\Local\ChocoCache\dnspyex.vm\6.3.0.20230605\dnSpy-net-win32.zip to C:\Tools\dnSpyEx32...
C:\Tools\dnSpyEx32
Added C:\ProgramData\chocolatey\bin\dnSpyEx32.exe shim pointed to 'c:\tools\dnspyex32\dnspy.exe'.
C:\Tools\dnSpyEx32\dnSpy.exe
Downloading dnspyex.vm
  from 'https://github.com/dnSpyEx/dnSpy/releases/download/v6.3.0/dnSpy-net-win64.zip'

Download of dnSpy-net-win64.zip (87.46 MB) completed.
Hashes match.
Extracting C:\Users\flare\AppData\Local\ChocoCache\dnspyex.vm\6.3.0.20230605\dnSpy-net-win64.zip to C:\Tools\dnSpyEx64...
C:\Tools\dnSpyEx64
Added C:\ProgramData\chocolatey\bin\dnSpyEx64.exe shim pointed to 'c:\tools\dnspyex64\dnspy.exe'.
C:\Tools\dnSpyEx64\dnSpy.exe
```

I am able to unzip `C:\Users\flare\AppData\Local\ChocoCache\dnspyex.vm\6.3.0.20230605\dnSpy-net-win32.zip` manually. Any idea what's going on?

mr-tz:
`dnSpy-netframework.zip` contains 32- and 64-bit EXEs. Maybe we can use that?

Ana06:
[...]

Issue #370, opened 2023-05-26
dnspyex.vm: shortcut opens console app


Ana06:
In dnspyex.vm the shortcut should open the GUI application, not the console one.

PR #413, opened 2023-06-05
dnspyex.vm: Fix dnSpy

Ana06:
Install dnSpyEx directly instead of using the community package as it does not install both 64 and 32 versions. Add shortcuts for both GUI applications (in addition to the console app that was already created).

Closes https://github.com/mandiant/VM-Packages/issues/412
Closes https://github.com/mandiant/VM-Packages/issues/370

### DIFF
--- a/packages/dnspyex.vm/tools/chocolateyinstall.ps1
+++ b/packages/dnspyex.vm/tools/chocolateyinstall.ps1
@@ -2,12 +2,22 @@ $ErrorActionPreference = 'Stop'
 Import-Module vm.common -Force -DisableNameChecking
 
 try {
-  $toolName = 'dnSpyEx'
+  $toolName = 'dnSpy'
   $category = 'dotNet'
-  $shimPath = 'bin\dnSpy.Console.exe'
 
-  $executablePath = Join-Path ${Env:ChocolateyInstall} $shimPath -Resolve
-  VM-Install-Shortcut $toolName $category $executablePath -consoleApp $true -arguments $null
+  $zipUrl = "https://github.com/dnSpyEx/dnSpy/releases/download/v6.3.0/dnSpy-netframework.zip"
+  $zipSha256 = "122df37b8668eb38be1c139cb244185824ff0f2ab4b4c81862a8397c6c2e7f1f"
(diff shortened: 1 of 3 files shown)


packages/dnspyex.vm/tools/chocolateyinstall.ps1
```diff
@@ -2,12 +2,22 @@ $ErrorActionPreference = 'Stop'
 Import-Module vm.common -Force -DisableNameChecking
 
 try {
-  $toolName = 'dnSpyEx'
+  $toolName = 'dnSpy'
```
mr-tz, line 5: I think this should stay `dnSpyEx` to indicate we're using the updated tool.

mr-tz, line 5:
```suggestion
  $toolName = 'dnSpyEx'
```

packages/dnspyex.vm/tools/chocolateyuninstall.ps1
```diff
@@ -1,7 +1,15 @@
 $ErrorActionPreference = 'Continue'
 Import-Module vm.common -Force -DisableNameChecking
 
-$toolName = 'dnSpyEx'
+$toolName = 'dnSpy'
```
mr-tz, line 4:
```suggestion
$toolName = 'dnSpyEx'
```

mr-tz (approved):
looks good besides the tool name

packages/dnspyex.vm/tools/chocolateyinstall.ps1
Ana06: Not really, the tool name (the exe) is called dnSpy.

The only problem is if we want to install dnSpy (the not Ex version), but it is not easy as we need to compile it. I was thinking maybe it is enough with dnSpyEx and in this case we can call the tool `dnSpy`.

In case you think we should call the tool `dnSpyEx`, how do you propose we implement it?

mr-tz: Ah, nevermind. Confused this with the package name I guess :)

Issue #268, opened 2023-03-17
Package proposal: diaphora.vm


Ana06:
### Package Name

diaphora

### Tool Name

diaphora

### Package type

OTHER/UNKNOWN

### Tool's version number

2.1.0

### Category

Utilities

### Tool's authors

joxeankoret

### Tool's description

Diaphora is a program diffing tool that works as an IDA plugin.

### Download URL

https://github.com/joxeankoret/diaphora/archive/refs/tags/2.1.0.zip

### Download SHA256 Hash

TODO

### Why is this tool a good addition?

Diaphora is very useful for comparing samples.

Ana06:
@mandiant/flare-vm where do you think we should put this tool? Is Utility the right place?

mr-tz:
Out of the current categories, I'd say yes. Utility is good.

Ana06:
What about adding an `IDA plugin` category? 🤔

Ana06:
It think it could be cool to install IDA plugins using VM-Packages. But I am thinking now that we don't need a category, but to link them from the IDA plugins directory. So probably a category doesn't make sense in any sense. Diaphora is a bit special because it is not really a plugin, but a script you need to open with IDA. So I think it is ok to add it to Utilities.

PR #417, opened 2023-06-06
ida.diaphora.vm: Add diaphora

Ana06:
Note diaphora is a bit different than the existent packages, as doesn't contain an executable, but a Python script that needs to be run from IDA. We just unzip the tool files to the Tools directory. Adjust the helper `VM-Install-From-Zip` to support this case.

I think we shouldn't add diaphora as default package for flare-vm, but I would like to install it myself.

Closes https://github.com/mandiant/VM-Packages/issues/268

### DIFF
--- a/packages/common.vm/tools/vm.common/vm.common.psm1
+++ b/packages/common.vm/tools/vm.common/vm.common.psm1
@@ -323,9 +323,13 @@ function VM-Install-From-Zip {
         [Parameter(Mandatory=$false)]
         [bool] $consoleApp=$false,
         [Parameter(Mandatory=$false)]
-        [bool] $innerFolder=$false, # subfolder in zip with the app files
+        [bool] $innerFolder=$false, # Subfolder in zip with the app files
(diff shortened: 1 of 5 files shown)


packages/common.vm/tools/vm.common/vm.common.psm1
```diff
@@ -325,7 +325,11 @@ function VM-Install-From-Zip {
         [Parameter(Mandatory=$false)]
         [bool] $innerFolder=$false, # subfolder in zip with the app files
         [Parameter(Mandatory=$false)]
-        [string] $arguments = "--help"
+        [string] $arguments = "--help",
+        [Parameter(Mandatory=$false)]
+        [string] $executableName,
+        [Parameter(Mandatory=$false)]
+        [switch] $withoutBinFile
```
mr-tz, line 332: can we document these with a bit more details?

mr-tz (approved):
LGTM, one comment on doc though

mr-tz (approved):
cool, thanks!

Issue #409, opened 2023-06-03
Unable to install boxstarter packages through chocolatey 2.0.0


Rahulvarma-13:
### What's the problem?

After running Unblock-File .\install.ps1 command and Set-ExecutionPolicy Unrestricted command, I ran the script using this command for the flare installation:
.\install.ps1 -customConfig https://raw.githubusercontent.com/HuskyHacks/PMAT-labs/main/config.xml

Once the password for VM is entered, the chocolatey was installed and when the boxstarter packages are installing, the flare installation was stopped and I got the following error:
Invoke-Expression: The term 'cinst' is not recognized as the name of a cmdlet, function, script file, or operable program. Check the spelling of the name, or if a path was included, verify that the path is correct and try again.

Checked the Chocolatey version and it was 2.0.0   Downgraded the Chocolatey version to 1.4.0 and ran the script again and now the flare installation was successful without any errors.

### Steps to Reproduce

1. In PowerShell Admin prompt, run: (New-Object net.webclient).DownloadFile('https://raw.githubusercontent.com/mandiant/flare-vm/main/install.ps1',"$([Environment]::GetFolderPath("Desktop"))\install.ps1")

2. Change directories to the Desktop

3. Run: Unblock-File .\install.ps1

4. Run: Set-ExecutionPolicy Unrestricted

5. Run: .\install.ps1 -customConfig https://raw.githubusercontent.com/HuskyHacks/PMAT-labs/main/config.xml

6. Enter VM password and click Y to continue the installation process.

### Environment

HostOS : Windows 11 Home
Virtualization software : Virtualbox 7.0

Version                 : 10.0.19045
BuildNumber             : 19045
OSArchitecture          : 64-bit
ServicePackMajorVersion : 0
Caption                 : Microsoft Windows 10 Enterprise Evaluation

VM OS HDD Space / Usage
-----

DeviceID DriveType ProviderName VolumeName     Size        FreeSpace
-------- --------- ------------ ----------     ----        ---------
C:       3                                     85286490112 50864492544
D:       5                      VBox_GAs_7.0.8 53485568    0

VM PowerShell Version
-----
5.1.19041.1682

Chocolatey version was now changed to 1.4.0 as i downgraded it to make it work through the installation process. Previously it was 2.0.0

VM Boxstarter Version
-----

Boxstarter|3.0.0
Boxstarter.Bootstrapper|3.0.0
Boxstarter.Chocolatey|3.0.0
[...]

PR #429, opened 2023-06-07
Chocolatey 2: Do not used removed shims

Ana06:
Some of our packages do not work with Chocolatey 2. See https://github.com/chocolatey/choco/issues/2642

Closes https://github.com/mandiant/VM-Packages/issues/409
Related https://github.com/mandiant/flare-vm/issues/458
Related https://github.com/mandiant/flare-vm/issues/456

### DIFF
--- a/packages/common.vm/tools/vm.common/vm.common.psm1
+++ b/packages/common.vm/tools/vm.common/vm.common.psm1
@@ -840,8 +840,8 @@ Common Environment Variables
     $diskInfo = Get-CimInstance -ClassName Win32_LogicalDisk | Out-String
     $psInfo = $PSVersionTable.PSVersion
     $psInfoClr = $PSVersionTable.CLRVersion
-    $chocoInfo = chocolatey --version
-    $installedPackages = chocolatey list -l -r -all
+    $chocoInfo = choco --version
+    $installedPackages = choco list -r
     $boxstarerInfo = $installedPackages | Select-String -Pattern "Boxstarter" | Out-String
     $installedPackages = $installedPackages | Out-String
 
--- a/packages/flarevm.installer.vm/tools/chocolateyinstall.ps1
+++ b/packages/flarevm.installer.vm/tools/chocolateyinstall.ps1
@@ -3,8 +3,8 @@ $global:VerbosePreference = "SilentlyContinue"
 Import-Module vm.common -Force -DisableNameChecking
 
 function Get-InstalledPackages {
-    if (Get-Command clist -ErrorAction:SilentlyContinue) {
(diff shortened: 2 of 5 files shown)


mr-tz (approved):
good catch!

packages/common.vm/tools/vm.common/vm.common.psm1
```diff
@@ -840,8 +840,8 @@ Common Environment Variables
     $diskInfo = Get-CimInstance -ClassName Win32_LogicalDisk | Out-String
     $psInfo = $PSVersionTable.PSVersion
     $psInfoClr = $PSVersionTable.CLRVersion
-    $chocoInfo = chocolatey --version
-    $installedPackages = chocolatey list -l -r -all
+    $chocoInfo = choco --version
+    $installedPackages = choco list -l -r -all
```
gep13, line 844:
```suggestion
    $installedPackages = choco list -r -all
```

There was another change in Chocolatey v2.0.0 where the list command has been made to only list local packages, and the `-l` option is no longer required, or supported.  If this VM is going to be using v2.0.0 of Chocolatey, I would suggest that you remove the `-l` option.

packages/flarevm.installer.vm/tools/chocolateyinstall.ps1
```diff
@@ -60,7 +60,7 @@ try {
         }
     }
 
-    $installedPackages = chocolatey list -l -r -all | Out-String
+    $installedPackages = choco list -l -r -all | Out-String
```
gep13, line 63:
```suggestion
    $installedPackages = choco list -r -all | Out-String
```

There was another change in Chocolatey v2.0.0 where the list command has been made to only list local packages, and the `-l` option is no longer required, or supported.  If this VM is going to be using v2.0.0 of Chocolatey, I would suggest that you remove the `-l` option.

packages/common.vm/tools/vm.common/vm.common.psm1
mr-tz: Neat, thanks for flagging this here!

Ana06:
Good catch @gep13! The same applies for the `-all` flag 👍

gep13:
> @Ana06 said...
Good catch @gep13! The same applies for the -all flag 👍

Doh! Missed that one, was focusing just on the `-l`. You are absolutely correct though, since the support for side-by-side has been removed, the `--all` option is not required, needed either.

gep13 (approved):
LGTM!

Ana06:
Thanks for the review @gep13

bakhtiyarjatoi:
Can you please update the script ? I'm having a problem to understand.

Issue #415, opened 2023-06-05
Add consoleApp argument to issue template


Ana06:
This would allow to create tools like the ones in https://github.com/mandiant/VM-Packages/pull/414 without manual changes.

PR #435, opened 2023-06-14
ci: Add consoleApp argument to issue template

Ana06:
Add consoleApp argument to issue template to allow us to create common package types without any manual modification.

You can see it working in https://github.com/Ana06/VM-Packages/pull/34.

Closes https://github.com/mandiant/VM-Packages/issues/415.


mr-tz (approved):
cool, thanks!

PR #439, opened 2023-06-16
ida.diaphora.vm: Fix common.vm version requirement

Ana06:
- Require common.vm version >= 0.0.0.20230615, as this package uses new arguments of `VM-Install-From-Zip` and installation fails in previous versions.
- Remove unneeded try-catch.

### DIFF
--- a/packages/ida.diaphora.vm/ida.diaphora.vm.nuspec
+++ b/packages/ida.diaphora.vm/ida.diaphora.vm.nuspec
@@ -2,11 +2,12 @@
 <package xmlns="http://schemas.microsoft.com/packaging/2015/06/nuspec.xsd">
   <metadata>
     <id>ida.diaphora.vm</id>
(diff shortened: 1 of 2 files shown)


mr-tz (approved):

packages/ida.diaphora.vm/ida.diaphora.vm.nuspec
```diff
@@ -2,11 +2,12 @@
 <package xmlns="http://schemas.microsoft.com/packaging/2015/06/nuspec.xsd">
   <metadata>
     <id>ida.diaphora.vm</id>
-    <version>2.1.0</version>
+    <version>2.1.0.20230617</version>
```
Ana06, line 5: @mr-tz we are in the future! 👀  today is 2023-06-16! Well, as long as we don't update this package again, it doesn't matter. 😆

PR #441, opened 2023-06-16
Add malware-jail

Ana06:
Add malware-jail, Ignore warnings while installing dependencies with npm to avoid failing the package. There may be a better way to do it, please let me know if you know how. 😄 

Update the common.vm documentation in this PR as well as it was not accurate and was getting me confused while adding malware-jail.

### DIFF
--- a/packages/malware-jail.vm/tools/chocolateyinstall.ps1
+++ b/packages/malware-jail.vm/tools/chocolateyinstall.ps1
@@ -1,0 +1,24 @@
+$ErrorActionPreference = 'Stop'
+Import-Module vm.common -Force -DisableNameChecking
+
+try {
+    $toolName = 'malware-jail'
+    $category = 'Javascript'
+
+    $zipUrl = 'https://github.com/HynekPetrak/malware-jail/archive/52d580bd76e0e0fc3ff2543758bb8fc45355c668.zip'
+    $zipSha256 = '24a9312635b95e4ccc41d5719a67f0db23cd577a648f33c7dab5d47c249531fd'
+    # Install dependencies with npm when running shortcut as we ignore errors below
+    $powershellCommand = "npm install; node jailme.js -h -b list"
+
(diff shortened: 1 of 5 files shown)


packages/malware-jail.vm/tools/chocolateyinstall.ps1
```diff
@@ -0,0 +1,23 @@
+Import-Module vm.common -Force -DisableNameChecking
+
+try {
+    $toolName = 'malware-jail'
+    $category = 'Javascript'
+
+    $zipUrl = 'https://github.com/HynekPetrak/malware-jail/archive/52d580bd76e0e0fc3ff2543758bb8fc45355c668.zip'
+    $zipSha256 = '24a9312635b95e4ccc41d5719a67f0db23cd577a648f33c7dab5d47c249531fd'
+    # Install dependencies with npm when running shortcut as we ignore errors below
+    $powershellCommand = "npm install; node jailme.js -h -b list"
```
mr-tz, line 10: does `npm install` need to run on every invocation?

mr-tz: or am I reading this wrong?

Ana06: We don't really need to, but I didn't find a way of running `npm install` without having to change to `$ErrorActionPreference = 'Continue'`. If `npm install` has been already run, it is very fast. 

If someone knows how we can run `npm install`, ensure it works and don't fail the package, I would really like to know how to do it. Otherwise, I see two options (and I have no strong opinion on what we should do):
1. We remove the `npm install` from the shortcut and hope that it is installed in line 24. But if it fails, it may the difficult for users to know what the problem is.
2. We run `npm install` during the installation (ignoring errors) and retry every time the shortcut is open. If it worked during installation it should be fast. If it failed, the user will see the error again.

mr-tz: Ah, I see. Then I think it's better we run `npm install` as you have right now to not mask any errors and confuse people.

Issue #430, opened 2023-06-07
idr.vm: downloading files from Google Drive fails


Ana06:
We have added the [Interactive Delphi Reconstructor (IDR)](https://github.com/crypto2011/IDR) to VM-Packages in https://github.com/mandiant/VM-Packages/pull/428 and as a default FLARE-VM tool in https://github.com/mandiant/flare-vm/pull/460

It seems the only place where we can download knowledge bases from is Google Drive
> Knowledge bases for various Delphi versions can be found at:
> https://drive.google.com/drive/folders/1IBsO1_O9xGHkPPV9Pnqkq2uoFmuak5C7?usp=sharing

But Google drive is not meant to be used in scripts and downloading the files from Google Drive doesn't seem to always work:

```
The remote file either doesn't exist, is unauthorized, or is forbidden for url 'https://drive.google.com/u/0/uc?id=16PREPxK6gJINAFC7TvS4k6qtXiZ71aEr&export=download'. Exception calling "GetResponse" with "0" argument(s): "The remote server returned an error: (403) Forbidden."
File appears to be downloaded already. Verifying with package checksum to determine if it needs to be redownloaded.
Error - hashes do not match. Actual value was '95800F4EAD2C1BCB71017DBCDFF0046D41F6E6E19CEBCB594E0010F7E7E5DE90'.
```
![Screenshot 2023-06-06 at 18 15 36](https://github.com/mandiant/VM-Packages/assets/16052290/a0b18136-8d76-44d3-813d-6276a6c4db7e)

As a workaround we continue if downloading any knowledge base fails without raising an error, hoping to download some of them.

It would be ideal if the knowledge bases were added to https://github.com/crypto2011/IDR

Ana06:
The maintainer of IDR has added the files to the GitHub repository 🎉

PR #467, opened 2023-06-27
idr.vm: Install knowledges bases from GitHub

Ana06:
The knowledge bases were only available in Google Drive. But Google drive is not meant to be used in scripts and downloading the files from Google Drive doesn't seem to always work. The maintainer of IDR has added the files to the GitHub repository and we can now take them from the downloaded zip file.

Closes https://github.com/mandiant/VM-Packages/issues/430


mr-tz (approved):
this very neat!

mr-tz:
Also big thanks to @crypto2011!

PR #471, opened 2023-06-29
Use new pestudio URL & update automatically version URLs

Ana06:
pestudio uses now a link that includes the version and will exist for two months, giving us time to update without breaking the package. Thanks to marc ochsenmeier for providing the new URL! 💘 

Try to update generic URLs by increasing the version sequentially in scripts/utils/update_package.py. Now the script is able to update tools like pestudio, and 010editor. You can see it working here: https://github.com/Ana06/VM-Packages/pull/7

This means **pestudio won't break anymore and we can add it back to the default flare-vm installation**. I'll send a PR in a moment. 😄

### DIFF
--- a/scripts/utils/update_package.py
+++ b/scripts/utils/update_package.py
@@ -111,6 +111,8 @@ def update_github_url(package):
         sha256 = get_sha256(url)
         latest_sha256 = get_sha256(latest_url)
         # Hash can be uppercase or downcase
+        if not latest_sha256:
+            return None
         content = content.replace(sha256, latest_sha256).replace(sha256.upper(), latest_sha256)
 
     content = content.replace(version, latest_version)
--- a/scripts/utils/update_package.py
+++ b/scripts/utils/update_package.py
@@ -30,6 +30,7 @@ def replace_version(latest_version, nuspec_content):
     return latest_version, re.sub("<version>[^<]+</version>", f"<version>{latest_version}</version>", nuspec_content)
 
 
+# Get latest version from GitHub releases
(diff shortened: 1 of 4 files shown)


scripts/utils/update_package.py
```diff
@@ -93,15 +114,68 @@ def update_github_url(package):
...
+    version_list_original = version.split(".")
+    # Try all possible increased versions, for example for 12.0.1
+    # ['13.0.1', '12.1.1', '12.0.2', '12.0.1.1']
+    versions = []
+    for i in range(len(version_list_original)):
+        version_list = version_list_original.copy()
+        version_list[i] = str(int(version_list[i]) + 1)
+        versions.append(".".join(version_list))
+    # New possible segment
+    versions.append(version + ".1")
```
mr-tz, line 135: what about .0, .2, .3, etc., i.e., does it suffice to assume it would be always .1?

mr-tz (approved):
nice, thanks!

scripts/utils/update_package.py
Ana06: There was a bug in this code, I have just updated it. I am assuming there will always be an update by 1. So for example for `12.0.1`, I am trying the possible version updates: `'12.0.1.1', '13', '13.0', '13.0.0', '13.0.0.0', '12.1', '12.1.0', '12.0.2'`. If a tool is updated twice, it will take several iterations to update to the newest version. But as we update every week, I think this is fine.

mr-tz (approved):

PR #474, opened 2023-07-03
Remove python3.vm (Previous PR deleted by accident)

Ana06:
This packages only install the dependency, then we can remove this package and use the dependency instead. In fact, we are already using the dependency directly in libraries.python3.vm

This was before in https://github.com/mandiant/VM-Packages/pull/331, I closed it by accident.

Closes https://github.com/mandiant/VM-Packages/issues/301


Ana06:
@mr-tz has already approved #331 and it had been open for a long time. So I think we can remove this package. I'll leave the package by now in myget, so in case someone is using it (unlikely) nothing will break. Just the package won't receive updates or be tested anymore.

PR #475, opened 2023-07-03
Remove libraries.python2.vm (Previous PR deleted by accident)

Ana06:
I think it is time to completely get rid of Python 2. Having only one libraries.python package saves as some work as we don't have to try to keep them in sync.

This was before in https://github.com/mandiant/VM-Packages/pull/332, I closed it by accident.

This has been discussed in
https://github.com/mandiant/VM-Packages/issues/308


Ana06:
@mr-tz has already approved #332 and it had been open for a long time. So I think we can remove this package. I'll leave the package by now in myget, so in case someone is using it (unlikely) nothing will break. Just the package won't receive updates and be tested anymore.

Issue #472, opened 2023-06-29
Add InnoSetup tools


Ana06:
I would like to add the following two tools/packages:
- [IFPSTools](https://github.com/Wack0/IFPSTools) 
  - depends on the community package `php`
  - php has to be configured to use mbstring extension: remove the `;` of the line `;extension=mbstring` in the php.ini file (inside the php installation folder)
  - run it with `php DisIFPS.php <your_compiled_script_file> <output file>`
- [isdsetup.1.5.exe.zip](https://github.com/dscharrer/innoextract/files/5507287/isdsetup.1.5.exe.zip) which I found in https://github.com/dscharrer/innoextract/issues/106

We have already 2 InnoSetup tools (innoextract and innounp). With these two extra it is already 4. I propose creating a `InnoSetup` category. I think it makes sense as Utilities is getting big. Moving the two existent packages may be a breaking change that may break updates. But we are not testing updates anyway and we recommend to do new installs vs updating packages.

mr-tz:
The site says: Superseded by [IFPSTools.NET](https://github.com/Wack0/IFPSTools.NET).

Let's use that instead?

Ana06:
It seems to work also fine. We can use the zip in the releases. Do we want to create a shortcurt only for `ifpsdasm` or also for any of the other tools?

PR #481, opened 2023-07-10
Add InnoSetup tools & category

Ana06:
- Create a new InnoSetup category for the four InnoSetup tools we have now to it easier to find them and make `Utilities` smaller.
- Add Inno Setup Decompiler
- Add ifpsdasm from IFPSTools.NET

Closes https://github.com/mandiant/VM-Packages/issues/472

### DIFF
--- a/.github/ISSUE_TEMPLATE/new_metapackage.yml
+++ b/.github/ISSUE_TEMPLATE/new_metapackage.yml
@@ -50,40 +50,41 @@ body:
       label: Category
       description: Which category should this tool be installed to?
       options:
+        - Active Directory
         - Android
         - Cloud
+        - Command & Control
+        - Credential Access
         - Debuggers
         - Delphi
         - Disassemblers
         - dotNet
+        - Evasion
+        - Exploitation
         - Forensic
         - Hex Editors
+        - Information Gathering
+        - InnoSetup
         - Java
         - Javascript
+        - Lateral Movement
         - Networking
         - Office
+        - Password Attacks
+        - Payload Development
         - PDF
         - PE
+        - Persistence
         - PowerShell
+        - Privilege Escalation
         - Python
+        - Reconnaissance
         - Text Editors
         - Utilities
         - VB
-        - Active Directory
-        - Command & Control
-        - Evasion
-        - Exploitation
-        - Information Gathering
-        - Password Attacks
         - Vulnerability Analysis
         - Web Application
         - Wordlists
-        - Payload Development
-        - Reconnaissance
-        - Persistence
-        - Privilege Escalation
-        - Credential Access
-        - Lateral Movement
   - type: input
     id: authors
     attributes:
--- a/.github/ISSUE_TEMPLATE/new_package.yml
+++ b/.github/ISSUE_TEMPLATE/new_package.yml
@@ -69,40 +69,41 @@ body:
       label: Category
       description: Which category should this tool be installed to?
       options:
+        - Active Directory
         - Android
         - Cloud
+        - Command & Control
+        - Credential Access
         - Debuggers
         - Delphi
         - Disassemblers
         - dotNet
+        - Evasion
+        - Exploitation
         - Forensic
         - Hex Editors
+        - Information Gathering
+        - InnoSetup
         - Java
         - Javascript
+        - Lateral Movement
         - Networking
         - Office
+        - Password Attacks
+        - Payload Development
         - PDF
         - PE
+        - Persistence
         - PowerShell
+        - Privilege Escalation
         - Python
+        - Reconnaissance
         - Text Editors
         - Utilities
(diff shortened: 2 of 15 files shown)


packages/ifpstools.vm/ifpstools.vm.nuspec
```diff
@@ -0,0 +1,12 @@
+<?xml version="1.0" encoding="utf-8"?>
+<package xmlns="http://schemas.microsoft.com/packaging/2015/06/nuspec.xsd">
+  <metadata>
+    <id>ifpstools.vm</id>
+    <version>2.0.2</version>
+    <authors>Wack0, zc, Jonson Tan</authors>
```
Ana06, line 6: @Wack0 we are adding IFPSTools.NET to flare-vm (`ifpsdasm` is very using useful to analyze InnoSetup!). I am not sure who are the tool's authors. Is `Wack0, zc, Jonson Tan`ok or should we use something different?

.github/ISSUE_TEMPLATE/new_package.yml
```diff
@@ -69,34 +69,35 @@ body:
...
         - Office
         - PDF
         - PE
+        - Password Attacks
         - PowerShell
         - Python
         - Text Editors
         - Utilities
         - VB
-        - Active Directory
```
mr-tz, line 91: I kind of liked that we separated the FLARE VM and Commando VM categories. Although separate lists would probably make more sense.

.github/ISSUE_TEMPLATE/new_package.yml
```diff
@@ -69,34 +69,35 @@ body:
...
-        - Command & Control
-        - Evasion
-        - Exploitation
-        - Information Gathering
-        - Password Attacks
         - Vulnerability Analysis
         - Web Application
         - Wordlists
+        - dotNet
```
mr-tz, line 100: if A-Z why is dotNet at the end?

packages/ifpstools.vm/ifpstools.vm.nuspec
Wack0: Main author is myself, `ifpscc` is a fork of a .NET C compiler by Jonson Tan, hence the additional credit.

Wack0: btw, I didn't know about Inno Setup Decompiler before, looks like an interesting tool (although it doesn't support `push` opcode and expects every script to have an entrypoint function, so anything hand-optimised or compiled by `ifpscc` won't work with it).

Decompiler and debugger is on my to-do list for IFPSTools.NET eventually (will probably end up doing debugger first, and I would prefer to emit C for any hypothetical decompiler rather than pascalscript...)

.github/ISSUE_TEMPLATE/new_package.yml
Ana06: Because it is not capitalized as the rest. I used `:sort` in vim. Should I put in in the D (ignoring capitalization)?

packages/ifpstools.vm/ifpstools.vm.nuspec
Ana06: @Wack0 thanks for the answer! 😄  Should we then use `Wack0, Jonson Tan` to give you credit to both of you? Should we use `Wack0` for you? or `Rairii`? we can use whatever you prefer 😅

.github/ISSUE_TEMPLATE/new_package.yml
Ana06: This is displayed to the user. I was thinking alphabetically would make it easier for users. I don't see how we could have separate lists here, can you elaborate?

.github/ISSUE_TEMPLATE/new_metapackage.yml
```diff
@@ -50,34 +50,35 @@ body:
...
         - Debuggers
         - Delphi
         - Disassemblers
-        - dotNet
+        - Evasion
+        - Exploitation
         - Forensic
         - Hex Editors
+        - Information Gathering
+        - InnoSetup
```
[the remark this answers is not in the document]
Ana06: We have 4 InnoSetup tools and I find it difficult to find them in Utilities. InnoSetup is a installer, not a packer. So I think there is no other category would it would work and I would like to add an InnoSetup category (as proposed in https://github.com/mandiant/VM-Packages/issues/472) @stevemk14ebr @tarterp @mandiant/flare-vm any opinions about in which category (folder name in Tools) we should use for the InnoSetup tools?

.github/ISSUE_TEMPLATE/new_package.yml
Ana06: I have updated it using `:sort i` (case insensitive)

Issue #518, opened 2023-07-11
Metapackage proposal: burp-free.vm


skollr34p3r:
### Package Name

burp-free

### Tool Name

BurpSuiteCommunity

### Package type

METAPACKAGE

### Tool's version number

0.0.0.20230711

### Category

Utilities

### Tool's authors

PortSwigger Ltd.

### Tool's description

Burp Suite Community Edition is PortSwigger's free integrated platform for performing security testing of web applications.

### Dependency

burp-suite-free-edition

### Shim path

"C:\Program Files\BurpSuiteCommunity\BurpSuiteCommunity.exe"

### Why is this tool a good addition?

Burp Suite Community Edition is PortSwigger's free integrated platform for performing security testing of web applications. Burp Suite is a tool used by almost all pentesters when assessing the security of web applications.

PR #519, opened 2023-07-11
:robot: Add burp-free.vm

vm-packages:
Automated PR from https://github.com/mandiant/VM-Packages/issues/518 :rocket:


Menn1s:
Try the version specified in Chocolatey? Seems to be looking for that 2022.12.4 version

skollr34p3r:
> Try the version specified in Chocolatey? Seems to be looking for that 2022.12.4 version

I noticed that as well. I thought we were supposed to specify the 0.0.0.date format, which is why I did so. I think stating that version for the package, but then listing the one expected in choco for the dependency was what I was missing. Thanks for pointing that out and getting me looking in the right direction!

Menn1s (approved):
Works awesome!

PR #447, opened 2023-06-21
:robot: Add gowitness.vm

vm-packages:
Automated PR from https://github.com/mandiant/VM-Packages/issues/446 :rocket:

PR #520, opened 2023-07-11
Missing Shortcut Working Directory Change

nos3curity:
@Menn1s noticed that a shortcut for a tool installed with the `common.vm` function `VM-Install-Single-Exe` was not changing the working directory into the tool folder, but instead defaulting to the user desktop #447 

Upon investigation, I found that this occurs because the `common.vm` function `VM-Install-Shortcut` does the following check:
```
if (!$executableDir) {
            $executableDir  = Join-Path ${Env:UserProfile} "Desktop"
}
```

Yet `VM-Install-Single-Exe` does not pass it the `executableDir` argument, as seen here:
```
VM-Install-Shortcut -toolName $toolName -category $category -executablePath $executablePath -consoleApp $consoleApp -arguments $arguments
```

Therefore, because we never pass the executable directory, it will always default to the user desktop no matter what. The fix is to pass the directory to the function:
```
VM-Install-Shortcut -toolName $toolName -category $category -executableDir $toolDir -executablePath $executablePath -consoleApp $consoleApp -arguments $arguments
```

### DIFF
--- a/packages/common.vm/common.vm.nuspec
+++ b/packages/common.vm/common.vm.nuspec
@@ -2,7 +2,7 @@
 <package xmlns="http://schemas.microsoft.com/packaging/2015/06/nuspec.xsd">
   <metadata>
     <id>common.vm</id>
-    <version>0.0.0.20230616</version>
+    <version>0.0.0.20230711</version>
     <description>Common libraries for VM-packages</description>
     <authors>Mandiant</authors>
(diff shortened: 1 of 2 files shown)


day1player (approved):
I dont think we would ever _not_ want the executableDir to be the toolDir, so I think this is a good fix.

Issue #584, opened 2023-07-14
Package proposal: hashcat.vm


Menn1s:
### Package Name

hashcat

### Tool Name

hashcat

### Package type

ZIP_EXE

### Is the tool a console application?

true

### Tool's version number

6.2.6

### Category

Credential Access

### Tool's authors

Jens Steube (jsteube)

### Tool's description

Hashcat is a fast password recovery utility.

### Download URL

https://github.com/hashcat/hashcat/releases/download/v6.2.6/hashcat-6.2.6.7z

### Download SHA256 Hash

96697e9ef6a795d45863c91d61be85a9f138596e3151e7c2cd63ccf48aaa8783

### Why is this tool a good addition?

Hashcat is used to crack hashes during proactive assessments and is the go-to open-source hash cracking engine.

PR #585, opened 2023-07-14
:robot: Add hashcat.vm

vm-packages:
Automated PR from https://github.com/mandiant/VM-Packages/issues/584 :rocket:


Menn1s:
In progress; Need to clean up vars and make the cmd window that pops up not exit so the user can use the working directory to run commands.

skollr34p3r:
Leaving comment here for visibility: I am unable to complete the installation on an empty (no packages) install of commandovm since it requires 7zip. 7zip could be added as a dependency to the nuspec file, but that would install a different version of 7zip than the 7zip package itself that we have installs. 

My suggestion is to remake the 7zip-15-05.vm package by using a version in choco, then making sure that the dependency for 7zip used by other packages is using the same version for consistency.

skollr34p3r (approved):
runs well on intel mac finally lol

PR #594, opened 2023-07-15
Standardize PowerShell Shortcuts

nos3curity:
`common.vm` function `VM-Install-Raw-GitHub-Repo` used for a lot of the PowerShell packages defines the shortcut like this:
```
```

However, `VM-Install-Single-Ps1` defines it differently, adding a double file extension with `ps1`:
```
$shortcut = Join-Path $shortcutDir "$toolName.ps1.lnk"
```

This creates an inconsistency, where some PowerShell tools have a `ps1` in their shortcut and some don't. This PR removes the `ps1` extension from the shortcut.


mr-tz (approved):

Ana06 (approved):
Thanks @nos3curity!

PR #597, opened 2023-07-16
added unzip location to VM-Install-From-Zip

emtuls:
This allows us to choose a specific location that we want to unzip our binary to. This is required for the `psnotify` package to work, as `psnotify` requires being located in the "C:\" directory, not the default "C:\Tools" which is our current default.

### DIFF
--- a/packages/common.vm/common.vm.nuspec
+++ b/packages/common.vm/common.vm.nuspec
@@ -2,7 +2,7 @@
 <package xmlns="http://schemas.microsoft.com/packaging/2015/06/nuspec.xsd">
   <metadata>
     <id>common.vm</id>
-    <version>0.0.0.20231013</version>
+    <version>0.0.0.20231020</version>
     <description>Common libraries for VM-packages</description>
     <authors>Mandiant</authors>
   </metadata>
--- a/packages/common.vm/tools/vm.common/vm.common.psm1
+++ b/packages/common.vm/tools/vm.common/vm.common.psm1
@@ -329,10 +329,17 @@ function VM-Install-From-Zip {
         [Parameter(Mandatory=$false)]
         [string] $executableName, # Executable name, needed if different from "$toolName.exe"
         [Parameter(Mandatory=$false)]
-        [switch] $withoutBinFile # Tool should not be installed as a bin file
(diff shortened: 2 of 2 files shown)


packages/common.vm/tools/vm.common/vm.common.psm1
```diff
@@ -329,10 +329,16 @@ function VM-Install-From-Zip {
...
         [Parameter(Mandatory=$false)]
-        [switch] $withoutBinFile # Tool should not be installed as a bin file
+        [switch] $withoutBinFile, # Tool should not be installed as a bin file
+        [Parameter(Mandatory=$false)]
+        [string] $unzipLocation
     )
     try {
-        $toolDir = Join-Path ${Env:RAW_TOOLS_DIR} $toolName
+        if ($unzipLocation) {
+            $toolDir = Join-Path $unzipLocation $toolName
```
Ana06, line 338: In the case of `psnotify` the folder has the toolName. But I wonder if it would be better to leave this more generic in case another tool requires a path with a different name. 🤔

Ana06: I think this comment still applies, or am I missing something?

emtuls: I am inclined to leave it as this for the moment, because this is how we treat other tools already (each tool is unzipped into a default directory + the tool name), so this change only adds the ability to change the default directory.

If we require modifying the name of the directory that the tool gets unzipped into as well, then I think that we can consider that when the case arises. :)

packages/common.vm/common.vm.nuspec
```diff
@@ -2,7 +2,7 @@
 <package xmlns="http://schemas.microsoft.com/packaging/2015/06/nuspec.xsd">
   <metadata>
     <id>common.vm</id>
-    <version>0.0.0.20230714</version>
+    <version>0.0.0.20230718</version>
```
Ana06, line 5: We need increase this version to today's date as this is now not increasing the version.

Ana06 (approved):
Please rebase (`git rebase main`) and update the common version and we are good to go! 😉

PR #601, opened 2023-07-27
new versions of Malcode Analyst Pack & scdbg

dzzie:


dzzie:
I am not going to sign the google CLA thing, because I cant be sure what email address they will attach to it and they have access to a private one. 

Anyway the old git repos will go away in a week or so as I move completely off the platform. They are only up right now so I dont break your install scripts

mr-tz:
thanks, for the heads-up and new links!

mr-tz:
is it ok if we use #621?

dzzie:
I just saw one of the auto checks was blocking because version was not updated in the nuspec's so added that as well

Ana06:
@dzzie sorry for the late reply and thank you so much for keeping the old version to avoid breaking our package.

I am very sorry that the signing of the CLA discourages you to contribute to VM-Packages. I could ask about the email attached to the CLA (or any other questions you have) if this would make it any easier.

I am closing this PR in favour of https://github.com/mandiant/VM-Packages/pull/621, there is a small problem with the version: https://github.com/mandiant/VM-Packages/pull/621/files#r1318383221 After we have fixed that, we can get it merged.

PR #621, opened 2023-08-07
update MAP and scdbg

mr-tz:
supersedes #601

### DIFF
--- a/packages/map.vm/map.vm.nuspec
+++ b/packages/map.vm/map.vm.nuspec
@@ -2,7 +2,7 @@
 <package xmlns="http://schemas.microsoft.com/packaging/2015/06/nuspec.xsd">
   <metadata>
     <id>map.vm</id>
-    <version>12.20.21</version>
+    <version>0.0.0.20230723</version>
     <authors>David Zimmer</authors>
     <description>Handful of small utility type applications useful for analyzing malicious code.</description>
     <dependencies>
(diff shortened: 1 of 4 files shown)


packages/map.vm/map.vm.nuspec
```diff
@@ -2,7 +2,7 @@
 <package xmlns="http://schemas.microsoft.com/packaging/2015/06/nuspec.xsd">
   <metadata>
     <id>map.vm</id>
-    <version>12.20.21</version>
+    <version>7.26.23</version>
```
Ana06, line 5: We should always increase the version, otherwise the old version will be installed (and tested). We need to add this to the linter. We should use the tool version here (and not the date). If there is no tool version, the version should be 0.0.0.DATE, using the 4th segment for the date. This way we ensure we always increase the version.

As the new version will be probably lower after using the correct one, before we can merge this PR, we need to change the version, delete the package in myget and run the CI to ensure this new version is working. We should try to merge this PR fast after deleting the package in myget as it will fail to install during that time.

Ana06 (changes requested):

mr-tz:
If you want, please go ahead and take this.

Ana06 (approved):
I have deleted the packages in myget and updated the version. The test suite is now testing the correct version. I am going to merge this quick, as the packages can't be installed at the moment.

PR #640, opened 2023-08-22
Adding package dnfile

emtuls:
This PR is to add the Python library `dnfile`, a tool that helps parse .NET executables, to the list of Python modules installed via PIP.

The package repository can be found here: https://github.com/malwarefrank/dnfile


mr-tz:
I'd suggest to disable it for now and create an issue to fix and reenable.

mr-tz (approved):

PR #642, opened 2023-08-23
remove chrome dependency from cyberchef

### LINKED ISSUE #296: Re-add Chrome as a dependency for CyberChef
The Google Chrome community package uses a link that doesn't include the version and breaks every time a new version of Chrome is released (which happens often). This break the cyberchef package until the community package is updated and we update the dependency (which happens automatically once a week). We could remove this dependency as use the default browser instead.

mr-tz:
closes #296


day1player (approved):

day1player:
Should also resolve https://github.com/mandiant/commando-vm/issues/299

PR #649, opened 2023-08-28
Added dnlib library

emtuls:


packages/dnlib.vm/tools/chocolateyinstall.ps1
```diff
@@ -0,0 +1,29 @@
...
+      unzipLocation  = $toolDir
+      url            = $zipUrl
+      checksum       = $zipSha256
+      checksumType   = 'sha256'
+      SpecificFolder = "lib"
+  }
+  Install-ChocolateyZipPackage @packageArgs
+  VM-Assert-Path $toolDir
+
+  $executablePath = Join-Path $toolDir "lib"
```
Ana06, line 24: Is this how we want to install dnlib? This creates a link int the FLARE-VM Tools folder to a folder with the following libraries:
Screenshot 2023-09-05 at 08 06 11

I am not sure if we need the link in the Tools folder as this is not an executable we can run. I am also not sure if copying the libraries somewhere could be more useful. @mike-hunhoff can you help us here? how are you using dnlib and where would you like to have it installed?

Ana06 (commented):
Thanks for adding dnlib @emtuls!

[nitpick] Good practice: Use imperative mood in the subject line of the commit message. See for example https://www.freecodecamp.org/news/how-to-write-better-git-commit-messages/

packages/dnlib.vm/tools/chocolateyinstall.ps1
mike-hunhoff: This looks fine to me! My uses cases for `dnlib` installed in this manner are to compile smaller projects directly via the C# compiler (`csc.exe`) and interactive use via the C# REPL. The important part here is that the `dnlib` DLL is easy to find.

Ana06: Good, then let's merge this! 👍  Thanks @mike-hunhoff

Issue #480, opened 2023-07-10
💡 Idea: Right click option to unzip with password infected


Ana06:
Many malware analysis related zips have the password `infected`. What about adding the option to unzip with this password to the right menu to speed unzipping this kind of files to the 7-zip package?

mr-tz:
To unzip, set:
    HKEY_CLASSES_ROOT*\shell\unzip "infected"\command
        "C:\Program Files\7-Zip\7z.exe" e -pinfected "%1"

https://moritzraabe.de/2023/06/05/windows-right-click-menu-shortcuts-open-with-ida/

PR #663, opened 2023-09-26
Add 'unzip "infected"' and 'Open with IDA" to right click menu

Ana06:
- Add 7z unzip with password "infected" option to the right menu for the most common extensions.
Screenshot 2023-09-26 at 19 12 15

- Add "Open with IDA" and "Open with IDA (x64)" to the right click menu. The command executes a Powershell script that looks for the latest IDA Pro version and uses it to open the file. If no IDA Pro version is found, it opens the file with IDA free.
  Screenshot 2023-09-26 at 19 11 38

- Simplify calls to `VM-Remove-From-Right-Click-Menu` by giving `type` as default value as it is `file` in most of the cases.

Closes https://github.com/mandiant/VM-Packages/issues/480

### DIFF
--- a/packages/7zip-15-05.vm/tools/chocolateyinstall.ps1
+++ b/packages/7zip-15-05.vm/tools/chocolateyinstall.ps1
@@ -31,8 +31,15 @@ try {
   Install-ChocolateyShortcut -shortcutFilePath $shortcut -targetPath $executablePath
   VM-Assert-Path $shortcut
 
-  $executablePath = Join-Path $toolDir "7z.exe" -Resolve
-  Install-BinFile -Name $toolName -Path $executablePath
+  $7zExecutablePath = Join-Path $toolDir "7z.exe" -Resolve
+  Install-BinFile -Name $toolName -Path $7zExecutablePath
+
+  # Add 7z unzip with password "infected" to the right menu for the most common extensions.
+  # 7z can unzip other file extensions like .docx but these don't likely use the infected password.
+  $extensions = @(".7z", ".bzip2", ".gzip", ".tar", ".wim", ".xz", ".txz", ".zip", ".rar")
+  foreach ($extension in $extensions) {
+    VM-Add-To-Right-Click-Menu $toolName 'unzip "infected"' "`"$7zExecutablePath`" e -pinfected `"%1`"" "$executablePath" -extension $extension
+  }
 } catch {
   VM-Write-Log-Exception $_
 }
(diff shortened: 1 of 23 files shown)


packages/7zip-15-05.vm/tools/chocolateyinstall.ps1
```diff
@@ -31,8 +31,15 @@ try {
...
   VM-Assert-Path $shortcut
 
-  $executablePath = Join-Path $toolDir "7z.exe" -Resolve
-  Install-BinFile -Name $toolName -Path $executablePath
+  $7zExecutablePath = Join-Path $toolDir "7z.exe" -Resolve
+  Install-BinFile -Name $toolName -Path $7zExecutablePath
+
+  # Add 7z unzip with password "infected" to the right menu for the most common extensions.
+  # 7z can unzip other file extensions like .docx but these don't likely use the infected password.
+  $extensions = @(".7z", ".bzip2", ".gzip", ".tar", ".wim", ".xz", ".txz", ".zip", ".rar")
```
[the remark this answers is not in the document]
mr-tz: looks good

packages/die.vm/die.vm.nuspec
```diff
@@ -2,11 +2,11 @@
 <package xmlns="http://schemas.microsoft.com/packaging/2015/06/nuspec.xsd">
   <metadata>
     <id>die.vm</id>
-    <version>3.07.20230523</version>
+    <version>3.07.20230925</version>
     <authors>Hellsp@wn, horsicq</authors>
     <description>Detect It Easy, or abbreviated "DIE" is a program for determining types of files.</description>
     <dependencies>
-      <dependency id="common.vm" />
+      <dependency id="common.vm" version="0.0.0.20230925" />
```
mr-tz, line 9: if we pin the dependency do we have to update manually?

mr-tz (approved):
one question on dependency pinning, otherwise looks good

packages/die.vm/die.vm.nuspec
Ana06: nuspec version ranges are a bit confusing:
Screenshot 2023-09-27 at 13 29 50
So this is requesting version >= 0.0.0.20230925 to ensure that the argument format is compatible. So that if an older `common.vm` is installed it is upgraded when installing this version of `die.vm`.

We don't need to update it manually when we update `common.vm`.

mr-tz: ok, perfect, that should work nicely then!

PR #670, opened 2023-09-29
Add pkg-unpacker & introduce Packers category 

Ana06:
Add pkg-unpacker, a nodejs tool installed similarly to malware-jail. Introduce a `Packers` category as the Utilities folder is getting big and to make easier finding tools. Add upx and pkg-unpacker to the packers category. Moving upx to the Packers category  likely breaks updates. But out documentation explains that updates are best-effort and mentions this case. I think it is better to introduce this category now that we only have to move one tool.


Ana06:
Merging... 🚢   If someone has any feedback on this PR, please open an issue/PR for further discussion.

PR #683, opened 2023-10-10
Installer improvements

Ana06:
- FLARE-VM does not use an icon for the tools folder at the moment. Do not make the installer fail if the icon file does not exist.
- be consistent with variable names: use `configPath` (and not `configFile`) for configuration file paths to be
consistent with the configuration file few lines up.
-  Use the WallpaperStyle setting `center` (0) if images' width/height is bigger or equal than 16:9 ht and `fit` (6) otherwise. Using the width/height ratio ensure that different images (including the ones used at the moment in Commando VM and FLARE-VM) are displayed nicely.
- Set the lock screen image if `%VM_COMMON_DIR%\lockscreen.png` is present.
Screenshot 2023-10-10 at 15 25 49

### DIFF
--- a/packages/installer.vm/tools/chocolateyinstall.ps1
+++ b/packages/installer.vm/tools/chocolateyinstall.ps1
@@ -122,8 +122,17 @@ try {
         Write-Host "`t[-] %LOCALAPPDATA%\Boxstarter\boxstarter.log" -ForegroundColor Yellow
     }
 
-    # Let users know installation is complete by setting background, playing win sound, and display message box
-    # Set background
+    # Let users know installation is complete by setting lock screen & wallpaper background, playing win sound, and display message box
+
+    # Set lock screen image
+    $lockScreenImage = "${Env:VM_COMMON_DIR}\lockscreen.png"
+    if ((Test-Path $lockScreenImage)) {
+        New-Item "HKLM:\SOFTWARE\Microsoft\Windows\CurrentVersion\PersonalizationCSP" -Force | Out-Null
+        New-ItemProperty -Path "HKLM:\SOFTWARE\Microsoft\Windows\CurrentVersion\PersonalizationCSP" -Name LockScreenImagePath -PropertyType String -Value $lockScreenImage -Force | Out-Null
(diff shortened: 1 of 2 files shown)


packages/installer.vm/tools/chocolateyinstall.ps1
```diff
@@ -121,13 +122,18 @@ try {
...
     $backgroundImage = "${Env:VM_COMMON_DIR}\background.png"
     if ((Test-Path $backgroundImage)) {
+        # Set lock screen image
+        New-Item "HKLM:\SOFTWARE\Microsoft\Windows\CurrentVersion\PersonalizationCSP" -Force | Out-Null
+        New-ItemProperty -Path "HKLM:\SOFTWARE\Microsoft\Windows\CurrentVersion\PersonalizationCSP" -Name LockScreenImagePath -PropertyType String -Value "C:\ProgramData\_VM\background.png" -Force | Out-Null
+        New-ItemProperty -Path "HKLM:\SOFTWARE\Microsoft\Windows\CurrentVersion\PersonalizationCSP" -Name LockScreenImageStatus -PropertyType DWord -Value 1 -Force | Out-Null
+        # Set desktop wallpaper image
         # Center: 0, Stretch: 2, Fit:6, Fill: 10, Span: 22
-        New-ItemProperty -Path "HKCU:\Control Panel\Desktop" -Name WallpaperStyle -PropertyType String -Value 0 -Force | Out-Null
+        New-ItemProperty -Path "HKCU:\Control Panel\Desktop" -Name WallpaperStyle -PropertyType String -Value 6 -Force | Out-Null
```
Ana06, line 136: @mandiant/flare-vm I think this means that Commando-VM needs to use the old version of the background image with the black background that was replaced in https://github.com/mandiant/commando-vm/pull/306. Is this ok for you? This allows us to use our current image in FLARE-VM and to use the same image for wallpaper and lockscreen (in both FLARE-VM and Commando VM)

Ana06: I discussed it with Commando and this doesn't work for them. I'll push a new proposal in a moment.

day1player (approved):

PR #690, opened 2023-10-13
installer.vm: Simplify dialog code

Ana06:
Simplify the dialog code using `System.Windows.MessageBox` to make it easier to read and maintain. The only advantage of using a `FlowLayoutPanel` is to be able to specify an icon, but the icon is so small that I can't really recognise commando-vm logo in it (and flare-vm doesn't have an icon). IMO the simplified version also looks more elegant.

### Before
Screenshot 2023-10-13 at 15 29 29

### After

Screenshot 2023-10-13 at 15 49 16

### DIFF
--- a/packages/installer.vm/installer.vm.nuspec
+++ b/packages/installer.vm/installer.vm.nuspec
@@ -2,7 +2,7 @@
 <package xmlns="http://schemas.microsoft.com/packaging/2015/06/nuspec.xsd">
   <metadata>
     <id>installer.vm</id>
-    <version>0.0.0.20231009</version>
+    <version>0.0.0.20231014</version>
     <authors>Mandiant</authors>
     <description>Generic installer for custom virtual machines.</description>
     <dependencies>
--- a/packages/installer.vm/tools/chocolateyinstall.ps1
+++ b/packages/installer.vm/tools/chocolateyinstall.ps1
@@ -166,55 +166,13 @@ public class VMBackground
         VM-Write-Log-Exception $_
     }
 
-    # Show dialog that install has been complete
-    Add-Type -AssemblyName System.Windows.Forms
-    Add-Type -AssemblyName System.Drawing
-    # Create form
-    $form = New-Object System.Windows.Forms.Form
-    $form.Text = "$Env:VMname Installation Complete"
-    $form.TopMost = $true
-    $form.StartPosition = [System.Windows.Forms.FormStartPosition]::CenterScreen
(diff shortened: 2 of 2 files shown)


mr-tz (approved):

packages/installer.vm/installer.vm.nuspec
```diff
@@ -2,7 +2,7 @@
 <package xmlns="http://schemas.microsoft.com/packaging/2015/06/nuspec.xsd">
   <metadata>
     <id>installer.vm</id>
-    <version>0.0.0.20231009</version>
+    <version>0.0.0.20231014</version>
```
Ana06, line 5: This version may need to be updated if other PR related to the installer is merged first

packages/installer.vm/tools/chocolateyinstall.ps1
```diff
@@ -166,55 +166,13 @@ public class VMBackground
         VM-Write-Log-Exception $_
     }
 
-    # Show dialog that install has been complete
-    Add-Type -AssemblyName System.Windows.Forms
-    Add-Type -AssemblyName System.Drawing
-    # Create form
-    $form = New-Object System.Windows.Forms.Form
-    $form.Text = "$Env:VMname Installation Complete"
-    $form.TopMost = $true
```
day1player, line 175: @Ana06 the reason I went with this version of the dialog window is because it will stay on top of all other Windows (line 175). I think this is better for the user because when they come back to the VM they will see the dialog box before they see anything else.

I'm all for making it smaller, we could even throw it into a function, but I think we need to include some way of making the dialog box sit on top of all the other windows.

packages/installer.vm/tools/chocolateyinstall.ps1
```diff
@@ -166,55 +166,13 @@ public class VMBackground
         VM-Write-Log-Exception $_
     }
 
-    # Show dialog that install has been complete
-    Add-Type -AssemblyName System.Windows.Forms
-    Add-Type -AssemblyName System.Drawing
-    # Create form
-    $form = New-Object System.Windows.Forms.Form
-    $form.Text = "$Env:VMname Installation Complete"
```
day1player, line 174: Also to note the inclusion of the `$Env:VMname` variable in the title bar

Ana06:
I think after the discussion in https://github.com/mandiant/VM-Packages/pull/956, we can close this PR. @day1player will send a PR to move this popup to under an env variable.

PR #700, opened 2023-10-18
Add the ImHex.vm package

d35ha:


packages/imhex.vm/imhex.vm.nuspec
```diff
@@ -0,0 +1,13 @@
+<?xml version="1.0" encoding="utf-8"?>
+<package xmlns="http://schemas.microsoft.com/packaging/2015/06/nuspec.xsd">
+  <metadata>
+    <id>imhex.vm</id>
+    <version>1.31.0.20231018</version>
+    <authors>WerWolv</authors>
+    <description>A Hex Editor for Reverse Engineers, Programmers and people who value their retinas when working at 3 AM.</description>
+    <dependencies>
+      <dependency id="common.vm" />
+      <dependency id="imhex" version="1.31.0" />
```
Ana06, line 10:
```suggestion
      <dependency id="imhex" version="[1.31.0]" />
```

`version="1.31.0"` requires >=1.31.0, while `version="[1.31.0]"` requires exactly version `1.31`. See nuspec documentation for more details: https://learn.microsoft.com/en-us/nuget/concepts/package-versioning#version-ranges

We want to require a concrete version to ensure that we have tested that the version installs correctly and we update the version automatically once a week.  This is documented in the [Wiki](https://github.com/mandiant/VM-Packages/wiki/Package-Structure#dependencies).

This is the way it is done if you use the metapackage template (by using `--type METAPACKAGE`): https://github.com/mandiant/VM-Packages/blob/main/scripts/utils/create_package_template.py#L66 😉

packages/imhex.vm/imhex.vm.nuspec
```diff
@@ -0,0 +1,13 @@
+<?xml version="1.0" encoding="utf-8"?>
+<package xmlns="http://schemas.microsoft.com/packaging/2015/06/nuspec.xsd">
+  <metadata>
+    <id>imhex.vm</id>
+    <version>1.31.0.20231018</version>
```
Ana06, line 5: [nitpick] The version is the tool version, so `1.31.0` is enough. The forth segment only needs to be used if we update the package without increasing the tool version, as the package version always needs to be increased. See some the [documentation in the Wiki](https://github.com/mandiant/VM-Packages/wiki/Package-Structure#version) and let me know if you think it is not clear.

Ana06 (changes requested):
This is a great addition! Thanks for your PR @d35ha 😄 You need to require a exact version before we can merge it, see my comment.

packages/imhex.vm/imhex.vm.nuspec
d35ha: Resolved at https://github.com/mandiant/VM-Packages/pull/700/commits/a0a79e9ca926a7158a5e9f988ecc35b328e43dab.

packages/imhex.vm/imhex.vm.nuspec
d35ha: Solved at https://github.com/mandiant/VM-Packages/pull/700/commits/a0a79e9ca926a7158a5e9f988ecc35b328e43dab.

Ana06 (changes requested):
@d35ha the code looks great! can you please squash your commits into just one `Add imhex.vm` commit in order to maintain a clean commit history? I think this will also fix the CLA issue. I think this is caused by my personal email being co-author of one of the commits (asking around to see if there is a solution for this in general).

Please let me know if you need any help squashing your commits (it can be tricky if you are doing it for the first time). Some documentation that may help: https://www.freecodecamp.org/news/git-squash-commits

packages/common.vm/tools/vm.common/vm.common.psm1
```diff
@@ -157,7 +157,7 @@ function VM-Check-Reboot {
     )
     try {
         if (Test-PendingReboot){
-            VM-Write-Log "ERROR" "Host must be rebooted before continuing install of $package.`n"
+            VM-Write-Log "ERROR" "Host must be rebooted before continuing installation of $package.`n"
```
Ana06, line 160: Something weird happen while rebasing as I am the author of the commit and you the commiter (you should get authorship) and thisthis change is already in main. But I think merging shouldn't break anything

Ana06 (approved):
great work, thanks @d35ha!

PR #703, opened 2023-10-18
Ignore success_failure.json

d35ha:


Ana06 (commented):
Thanks for the PR! I wonder how you are doing the testing so that this file needs to be added here. The packages should only be added in a VM and I recommend to not do the development (git commitment) in the same VM. But is ok for me to add this change if you think is helpful (even if I use a different workflow) as it doesn't really hurt.

Ana06:
@mandiant/vms should we merge this PR?

Ana06 (approved):

Issue #686, opened 2023-10-12
python3 modifies the path in an unexpected way


Ana06:
### What's the problem?

After installing libraries.python3.vm `C:\Python310\Scripts\` is at the top of PATH, instead of `C:\ProgramData\chocolatey\bin`, which means that the python capa library is found instead of the capa executable (that includes the rules). I think we should read the chocolatey path at the top after installing Python. This issue probably have to be solved in combination with https://github.com/mandiant/VM-Packages/issues/673

### Steps to Reproduce

Install libraries.python3.vm and execute capa from the command line.

### Environment

Environment independent.

### Detected by test suite

No

PR #710, opened 2023-10-19
Add python3.vm

Ana06:
Add python3.vm, a metapackage that install python3 and re-adds chocolate shim path to the top of the path. This package is to be requested by all packages needing python3 to ensure they all install the same version and that the paths is being updated.

Closes https://github.com/mandiant/VM-Packages/issues/686
Partially addresses https://github.com/mandiant/VM-Packages/issues/686

### DIFF
--- a/packages/python3.vm/python3.vm.nuspec
+++ b/packages/python3.vm/python3.vm.nuspec
@@ -1,13 +1,13 @@
 <?xml version="1.0" encoding="utf-8"?>
 <package xmlns="http://schemas.microsoft.com/packaging/2015/06/nuspec.xsd">
   <metadata>
-    <id>didier-stevens-suite.vm</id>
+    <id>python3.vm</id>
     <version>0.0.0.20231019</version>
-    <authors>Didier Stevens</authors>
-    <description>Tools collection by Didier Stevens</description>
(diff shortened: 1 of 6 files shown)


packages/python3.vm/python3.vm.nuspec
```diff
@@ -0,0 +1,12 @@
+<?xml version="1.0" encoding="utf-8"?>
+<package xmlns="http://schemas.microsoft.com/packaging/2015/06/nuspec.xsd">
+  <metadata>
+    <id>python3.vm</id>
+    <version>3.10.0.20231019</version>
+    <description>Metapackage to install common Python to ensure all packages use the same Python version.</description>
+    <authors>Mandiant</authors>
+    <dependencies>
+      <dependency id="python3" version="[3.10.0, 3.11.0)" />
```
binjo, line 9: `[3.10.0, 3.11.0)` is the ending `)` a typo or doesn't matter?

Ana06: It is on purpose to install 3.10.x, see nuspec versioning: https://learn.microsoft.com/en-us/nuget/concepts/package-versioning#version-ranges

PR #718, opened 2023-10-21
Clean up psnotify chocoinstall

emtuls:


packages/psnotify.vm/tools/chocolateyinstall.ps1
```diff
@@ -10,38 +10,31 @@ try {
 
     $unzipLocation = 'C:\'  # psnotify has a requirement of being located in 'C:\psnotify'
```
Ana06, line 11: why can't you use the helper function with this unzip location instead of duplicating the code? what else are you changing to need this duplication?

packages/psnotify.vm/tools/chocolateyuninstall.ps1
```diff
@@ -9,7 +9,4 @@ $toolDir = 'C:\psnotify'
 Remove-Item $toolDir -Recurse -Force -ea 0 | Out-Null
 
 # Remove tool shortcut
-VM-Remove-Tool-Shortcut $toolName $category
-
-# Uninstall binary
-Uninstall-BinFile -Name $toolName
```
Ana06, line 15: why shouldn't we install the tool as a bin file?

Ana06 (commented):
@emtuls please do not merge PRs without a review (specially if there are discussions going on) and do not approve your own changes. The idea is that someone else reviews the changes as merging automatically makes packages available in myget. This helps avoiding cleanup PRs like this one.

[nitpick] Use the imperative mood in the subject line of commit messages. See for example https://www.freecodecamp.org/news/how-to-write-better-git-commit-messages/

packages/psnotify.vm/tools/chocolateyinstall.ps1
emtuls: I needed to add `-WorkingDirectory $executableDir` to creating the shortcut, which was done in the other PR. 

It didn't seem like a change that needed to be made for every package going forward at the moment, but if it is, we can make another PR for that as well.

packages/psnotify.vm/tools/chocolateyinstall.ps1
```diff
@@ -10,38 +10,31 @@ try {
...
+    Install-ChocolateyZipPackage @packageArgs
     VM-Assert-Path $toolDir
+    $executablePath = Join-Path $toolDir "$toolName.exe" -Resolve
+    $shortcut = Join-Path $shortcutDir "$toolName.lnk"
+
+    $executableDir  = $toolDir
+    Install-ChocolateyShortcut -shortcutFilePath $shortcut -targetPath $executablePath -WorkingDirectory $executableDir
+    VM-Assert-Path $shortcut
+
+    return $executablePath
```
Ana06, line 37: This return should be removed. It was in the helper to use from the calling code (outside the helper).

Ana06 (changes requested):

Ana06 (approved):
Thanks @emtuls!

PR #721, opened 2023-10-25
category: Rename 'Android/Java' to fix bug

Ana06:
The `Java/Android` category is creating a subfolder `Android` inside of the `Java` category and this is not the desired result. Rename to `Java & Android` to fix the issue.


mr-tz (approved):
oh good catch

Issue #672, opened 2023-09-29
Zip downloads


Ana06:
Some of the helper functions in common include some code to handle updates. But we are not using this code everywhere, such us for example in the packages that download ZIPs directly such us idr.vm and didier-stevens-suite.vm,  I am not sure if this code is needed. We should either remove it or create a helper function to download zips that is used everywhere else.

I think @MalwareMechanic wrote this code and may have some insight in why it is needed.

Ana06:
Documenting discussion with @MalwareMechanic: 
> Code mainly assists "upgrades", but since it's not a priority or something we want to maintain, consensus is to **remove code to handle upgrades**

PR #727, opened 2023-10-27
Improve zip download code

Ana06:
- simplify code that handles upgrades: By unzipping directly in the tools directory, chocolatey takes care of
maintaining the files to delete on an update. 
- We introduced unziplocation as a parameter for VM-Install-From-Zip in https://github.com/mandiant/VM-Packages/pull/597 as we were planing to use it for psnotify.vm, but we realized that it is better not to use the helper for this package. So this argument is unused. Remove it to make it easier to maintain the function.

If we merge this PR, the code of this two functions is very similar, which is not surprising as a GitHub-Raw install is the same as installing a Zip. Any thoughts about merging this two functions?

Closes https://github.com/mandiant/VM-Packages/issues/672

### DIFF
--- a/packages/common.vm/common.vm.nuspec
+++ b/packages/common.vm/common.vm.nuspec
@@ -2,7 +2,7 @@
 <package xmlns="http://schemas.microsoft.com/packaging/2015/06/nuspec.xsd">
   <metadata>
     <id>common.vm</id>
-    <version>0.0.0.20231020</version>
+    <version>0.0.0.20231027</version>
     <description>Common libraries for VM-packages</description>
     <authors>Mandiant</authors>
   </metadata>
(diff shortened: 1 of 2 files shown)


MalwareMechanic:
@Ana06 have you tested that the `unzipLocation` works as expected when calling `Install-ChocolateyZipPackage`? If so, it looks good to me. Given that the two functions are so similar, I'd recommend combining them to reduce confusion on which to use when.

Ana06:
> have you tested that the unzipLocation works as expected when calling Install-ChocolateyZipPackage?

I think so, packages continue to install correctly. Is there something else I need to test?

MalwareMechanic (approved):
LGTM

PR #738, opened 2023-11-09
Update new_metapackage.yml

MalwareMechanic:
Updated the description for the shim path to better describe what path to provide.


Ana06 (approved):
👍

PR #747, opened 2023-11-16
Clean up Desktop and Temporary Files

emtuls:
This resolves #517 from flare-vm (https://github.com/mandiant/flare-vm/issues/517) by adding each command to a separate function. These functions will need to be called from somewhere in order for them to work.

The desktop function is run against both the Current User and 'Public' desktops due to some cases where desktop icons showing on Current User desktop that are only located in Public/Desktop.


packages/common.vm/tools/vm.common/vm.common.psm1
```diff
@@ -1244,3 +1244,62 @@ public class Shell {
...
+# Usage example:
+# VM-Remove-DesktopFilesExcept -excludeFolders "PS_Transcripts", "Tools", "fakenet_logs" -excludeFiles "example.txt", "important.doc"
+function VM-Remove-DesktopFilesExcept {
+    param (
+        [Parameter(Mandatory=$false)]
+        [string[]]$excludeFolders,
+        [Parameter(Mandatory=$false)]
+        [string[]]$excludeFiles
+    )
+    $excludeFolders = $excludeFolders
```
Ana06, line 1258: I think we should ensure the following folders are never deleted by adding them here to the provided argument:
- the tools directory (use env variable instead of hardcoded value)
- `fakenet_logs`

Ana06 (changes requested):
This is great work @emtuls!

We need a single wrapper function (called something like `VM-Clean-Up`) that executes the 4 steps (the 4 functions you have created) as we may need to call this function manually often. I would keep the functions that have more logic, but I would get rid of the ones that just execute just one  command and just call the command directly.

I like to be able to provide items to ignore for deletion (as you are doing) and I think we should have the ignore arguments in the wrapper function as well.

packages/common.vm/tools/vm.common/vm.common.psm1
emtuls: I will add "PS_Transcripts" to this as well due an issue that can occur based on timing of execution of this function. 

If a restart happens and a PS log in this folder is generated and then this function is run, it will fail out due to a file being in use. 

I wasn't able to get it to cleanly bypass it for the moment and the circumstances are likely to be rare, but I think adding this for now won't be an issue. We can look into other ways to mitigate this in the future if needed.

packages/common.vm/tools/vm.common/vm.common.psm1
```diff
@@ -1244,3 +1244,91 @@ public class Shell {
...
+    param (
+        [Parameter(Mandatory=$false)]
+        [string[]]$excludeFolders,
+        [Parameter(Mandatory=$false)]
+        [string[]]$excludeFiles
+    )
+    # Ensure that "PS_Transcripts", "Desktop\Tools", and "fakenet_logs" folders are not to be deleted.
+    $defaultExcludedFolders = @("PS_Transcripts", ${Env:TOOL_LIST_DIR}, "fakenet_logs")
+    $excludeFolders = $excludeFolders + $defaultExcludedFolders
+    $excludeFiles = $excludeFiles   # required for lint.ps1 to pass
```
Ana06, line 1263: why is this exactly needed? What does the linter complain about?

packages/common.vm/tools/vm.common/vm.common.psm1
```diff
@@ -1244,3 +1244,91 @@ public class Shell {
...
+# The function is run against both the Current User and 'Public' desktops due to some cases where desktop icons showing on
+# Current user Desktop that are only located in Public/Desktop.
+function VM-Remove-DesktopFiles {
+    param (
+        [Parameter(Mandatory=$false)]
+        [string[]]$excludeFolders,
+        [Parameter(Mandatory=$false)]
+        [string[]]$excludeFiles
+    )
+    # Ensure that "PS_Transcripts", "Desktop\Tools", and "fakenet_logs" folders are not to be deleted.
```
Ana06, line 1260: [nitpick] The tools dirs is not always in `Desktop\Tools`. Although maybe it is clear what is meant, I purpose:
```suggestion
    # Ensure that the "PS_Transcripts" and "fakenet_logs" folders as well as the tools directory are not to be deleted.
```

packages/common.vm/tools/vm.common/vm.common.psm1
```diff
@@ -1244,3 +1244,91 @@ public class Shell {
...
+
+# SDelete can take a bit of time (~2+ mins) and requires sysinternals to be installed
+function VM-Clear-FreeSpace {
+    VM-Write-Log "INFO" "Performing SDelete to optimize disk."
+    $sdeletePath = Get-Command -Name "sdelete.exe" -ErrorAction SilentlyContinue
+    if ($sdeletePath) {
+        Invoke-Expression 'cmd /c sdelete -accepteula -nobanner -z C:'
+    }
+    else {
+        VM-Write-Log "WARN" "SDelete not found. Please make sure sysinternals.vm is installed and SDelete is in the system's PATH."
```
Ana06, line 1311: [nitpick] Mention what we are trying to do (free space) in the warning
```suggestion
        VM-Write-Log "WARN" "SDelete not found. Ensure sysinternals.vm is installed and SDelete is in the system's PATH before running VM-Clear-FreeSpace to free space."
```

Ana06 (commented):
This looks great! We need to discuss what to do with the linter complain that seems to not improve the code and then we are ready to go 😃

packages/common.vm/tools/vm.common/vm.common.psm1
emtuls: I agree, it isn't always on the desktop. However, because this function only removes folders that -are- on the desktop, if the Tools directory isn't located there, it will not be touched. :)

I can reword the comment to be a little more clear about that if need be.

packages/common.vm/tools/vm.common/vm.common.psm1
emtuls: The linter says that the `$excludeFiles` variable isn't used, even though it is being used in the `if statement`. I think this might be a bug in how they are parsing things.

It said the same for `$excludeFolders` as well, but now that we add `+ $defaultExcludedFolders`, this is a little less obvious that I was doing the same thing for that variable as well.

PR #748, opened 2023-11-21
installer.vm: Add custom layout for taskbar pinning.

emtuls:
This is a first stab at resolving https://github.com/mandiant/flare-vm/issues/512. This works in tandem with https://github.com/mandiant/flare-vm/pull/531

It adds the creation of an `Admin Command Prompt` shortcut into the `Tools/Utilities` directory so that we are able pin that, and also sets a custom Start Layout template which pins the items from the list to the taskbar.

The main current shortfalls are:
- File Explorer does not point to Tools directory
  - This requires access to a known shortcut location. It may be possible to use `Get-ChildItem -Path ${Env:TOOL_LIST_DIR}`. Further testing is needed.
- IDA shortcut currently points to IDA Free and not automatically set for any IDA Pro installation

One notable thing to mention: If an executable does not exist when the custom template is set, it **_does not_** create a broken shortcut in its place. It simply leaves it out.
- This is useful if we want to add an `IDA Pro` shortcut in our default template, because it will not create any issues if `IDA Pro` does not exist when the template is set, but it will pin it when it is installed and either a restart or logout/login occurs (or just a restart of `explorer.exe`)

Closes https://github.com/mandiant/flare-vm/issues/512

### DIFF
--- a/packages/installer.vm/tools/chocolateyinstall.ps1
+++ b/packages/installer.vm/tools/chocolateyinstall.ps1
@@ -32,6 +32,25 @@ try {
     }
     VM-Write-Log "INFO" "Packages installation complete"
 
+    ## Configure taskbar with custom Start Layout if it exists.
+    $customLayout = Join-Path ${Env:VM_COMMON_DIR} "CustomStartLayout.xml"
+    if (Test-Path $customLayout) {
+        # Create an Admin Command Prompt shortcut and it to pin to taskbar (analogous to how the Windows Dev VM does this).
+        $toolName = 'Admin Command Prompt'
+
+        $executablePath = Join-Path ${Env:SystemRoot} 'system32\cmd.exe'
+        $shortcutDir = ${Env:RAW_TOOLS_DIR}
+        $shortcut = Join-Path $shortcutDir "$toolName.lnk"
+
+        Install-ChocolateyShortcut -shortcutFilePath $shortcut -targetPath $executablePath -RunAsAdmin
+        VM-Assert-Path $shortcut
+
+        Import-StartLayout -LayoutPath $customLayout -MountPath "C:\"
+        Stop-Process -Name explorer -Force  # This restarts the explorer process so that the new taskbar is displayed.
+    } else {
+        VM-Write-Log "WARN" "CustomStartLayout.xml missing. No items will be pinned to the taskbar."
+    }
+
     # Set Profile/Version specific configurations
     VM-Write-Log "INFO" "Beginning Windows OS VM profile configuration changes"
     $configPath = Join-Path $Env:VM_COMMON_DIR "config.xml" -Resolve
--- a/packages/installer.vm/installer.vm.nuspec
+++ b/packages/installer.vm/installer.vm.nuspec
@@ -2,7 +2,7 @@
 <package xmlns="http://schemas.microsoft.com/packaging/2015/06/nuspec.xsd">
   <metadata>
     <id>installer.vm</id>
-    <version>0.0.0.20231020</version>
+    <version>0.0.0.20231121</version>
     <authors>Mandiant</authors>
     <description>Generic installer for custom virtual machines.</description>
     <dependencies>


packages/installer.vm/tools/chocolateyinstall.ps1
```diff
@@ -32,6 +32,22 @@ try {
...
+
+    $executablePath = Join-Path ${Env:SystemRoot} 'system32\cmd.exe'
+    $shortcutDir = Join-Path ${Env:TOOL_LIST_DIR} $category -Resolve
+    $shortcut = Join-Path $shortcutDir "$toolName.lnk"
+
+    Install-ChocolateyShortcut -shortcutFilePath $shortcut -targetPath $executablePath -RunAsAdmin
+    VM-Assert-Path $shortcut
+
+    ## Configure taskbar with custom Start Layout
+    $customLayout = Join-Path ${Env:USERPROFILE} 'Desktop\CustomStartLayout.xml'
```
Ana06, line 47: I think this file should be stored in `${Env:VM_COMMON_DIR}` (as I also pointed out in https://github.com/mandiant/flare-vm/pull/531/files#r1400438797)

packages/installer.vm/tools/chocolateyinstall.ps1
```diff
@@ -32,6 +32,22 @@ try {
     }
     VM-Write-Log "INFO" "Packages installation complete"
 
+    # Create an Admin Command Prompt shortcut to pin to taskbar
+    $toolName = 'Admin Command Prompt'
+    $category = 'Utilities'
```
Ana06, line 37: I am not sure about installing this inside a category, see my comment in https://github.com/mandiant/flare-vm/pull/531/files#r1400442141

packages/installer.vm/tools/chocolateyinstall.ps1
Ana06: What if the file doesn't exists? I think we should then render a warning and do not apply the layout.

Ana06 (changes requested):
I think this is a great approach to solve https://github.com/mandiant/flare-vm/issues/512, thanks for all the research and testing @emtuls!

@mandiant/commando-vm I think this is also you could benefit from 😉

packages/installer.vm/tools/chocolateyinstall.ps1
emtuls: Adjusted the location and applied a WARN if the file doesn't exist. Thank you for the review!

packages/installer.vm/tools/chocolateyinstall.ps1
emtuls: Moved it over to `Env{RAW_TOOLS_DIR}`

Ana06 (changes requested):
Thanks for the changes @emtuls, some minor things to address 😉

packages/installer.vm/tools/chocolateyinstall.ps1
```diff
@@ -32,6 +32,25 @@ try {
     }
     VM-Write-Log "INFO" "Packages installation complete"
 
+    # Create an Admin Command Prompt shortcut to pin to taskbar
+    $toolName = 'Admin Command Prompt'
```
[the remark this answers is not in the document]
emtuls: Resolved. Staying with 'Admin Command Prompt' to match Windows Dev VM's implementation. :)

packages/installer.vm/tools/chocolateyinstall.ps1
```diff
@@ -32,6 +32,25 @@ try {
...
+
+    $executablePath = Join-Path ${Env:SystemRoot} 'system32\cmd.exe'
+    $shortcutDir = ${Env:RAW_TOOLS_DIR}
+    $shortcut = Join-Path $shortcutDir "$toolName.lnk"
+
+    Install-ChocolateyShortcut -shortcutFilePath $shortcut -targetPath $executablePath -RunAsAdmin
+    VM-Assert-Path $shortcut
```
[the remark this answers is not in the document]
emtuls: Complete!

packages/installer.vm/tools/chocolateyinstall.ps1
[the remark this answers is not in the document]
Ana06: We discussed this outside GH, for documentation proposes: We agreed  to leave it like it is to match Windows Dev VM implementation and add a comment to explain why we chose that name.

Ana06 (approved):
Great work @emtuls, looking forward to build the new VM! 🥳 🍾 🎉 

I have extracted the opining of File Explorer into the Tools folder to https://github.com/mandiant/flare-vm/issues/535 so that merging this PR will completely solve https://github.com/mandiant/flare-vm/issues/512. I have added the `Close 512` to this PR description so that the closing of that issue happens automatically. 😉

PR #751, opened 2023-11-23
Improve files deleted in VM-Remove-DesktopFiles/VM-Clean-Up

Ana06:
Improve files deleted in `VM-Remove-DesktopFiles` (called from `VM-Clean-Up`):
- Use `-Force` in `Get-ChildItem` to delete hidden files (such as `desktop.ini`) in `VM-Remove-DesktopFiles`. 
- We originally decided to prevent that `PS_Transcripts` is deleted in https://github.com/mandiant/VM-Packages/pull/747. But after installing FLARE-VM I would like to delete the files generated by the installation in the `PS_Transcripts` directory. Remove the protection, allowing to delete this folder. It is still possible to keep it using the `excludedFolders` function argument.

@mandiant/commando-vm do you know if there is a reason not to delete `desktop.ini` files? I think they have appeared in FLARE-VM since we introduce the debloat and new installer (probably because we now show hidden files).

### DIFF
--- a/packages/common.vm/common.vm.nuspec
+++ b/packages/common.vm/common.vm.nuspec
@@ -2,7 +2,7 @@
 <package xmlns="http://schemas.microsoft.com/packaging/2015/06/nuspec.xsd">
   <metadata>
     <id>common.vm</id>
-    <version>0.0.0.20231116</version>
+    <version>0.0.0.20231123</version>
     <description>Common libraries for VM-packages</description>
     <authors>Mandiant</authors>
   </metadata>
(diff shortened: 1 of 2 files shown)


mr-tz:
I guess it could store some config? It may also be recreated automatically again at some point.

Ana06:
> I guess it could store some config? It may also be recreated automatically again at some point.

I am not sure I know what you mean. Can you please elaborate? Note that his function is thought to be called after performing a fresh install of FLARE-VM. I think we want to also get rid of any config file and the `desktop.ini` files. Note this function is not called from the installer and need to be called manually. The function also supports arguments to exclude any extra files.

Issue #694, opened 2023-10-16
Improve "Open with IDA" to right click menu implementation


Ana06:
At the moment we are using a PowerShell script to look for the latest version of IDA (potentially the Pro version installed after the ida.vm package). But there is no way to not show the PowerShell console (without a wrapper program). I think there are better ways to implement this.

Idea: Implement a binary launcher that starts the latest version of IDA (as the PowerShell script does). This has the advantage that we can also ping the binary to the taskbar and use it in the Tools folder.

Opinions? Other ideas?

williballenthin:
https://github.com/namazso/IDAShell

Ana06:
This can also be used to pin the correct ida to the taskbar

Ana06:
I wonder if we can ping IDAShell to the taskbar, the we dont need to create a binary (assuming IDAShell is working as described in the first comment, I haven't tried).

emtuls:
Initial PR for this is up: https://github.com/mandiant/VM-Packages/pull/758

PR #758, opened 2023-11-29
Add ida_launcher.exe to idafree.vm and VM-Packages

emtuls:
This resolves https://github.com/mandiant/VM-Packages/issues/694


ida_launcher/src/ida_launcher.c
```diff
@@ -0,0 +1,88 @@
...
+    return found;
+}
+
+// WinMain entry point
+int WINAPI WinMain(HINSTANCE hInstance, HINSTANCE hPrevInstance, LPSTR lpCmdLine, int nCmdShow) {
+    char idaPath[260] = {0};
+    DWORD dwSize = sizeof(idaPath);
+
+    // Search for the highest version of IDA Pro first
+    if (!findHighestIdaVersion(HKEY_LOCAL_MACHINE, "SOFTWARE\\Wow6432Node\\Microsoft\\Windows\\CurrentVersion\\Uninstall", "IDA Pro", idaPath, dwSize) &&
```
Ana06, line 65: I would prefer to look for the IDA executables in `%programfiles%` instead of using the Uninstall registry key (similar to the current PowerShell script). I think the code would be shorter, simplier and easier to understand.

As `FindFirstFile` supports wildcharacters, I think `FindFirstFile("%programfiles%\\IDA Pro *\\ida64.exe", &FindFileData)` returns the highest IDA Pro version and `FindFirstFile("%programfiles%\\IDA Freeware *\\ida64.exe", &FindFileData)` returns the highest IDA free version.

packages/idafree.vm/tools/chocolateyinstall.ps1
```diff
@@ -31,15 +31,22 @@ try {
...
+  $launcherPath = Join-Path ${Env:RAW_TOOLS_DIR} $launcherName
+  Write-Host "[+] Downloading '$launcherName' from '$launcherSource'"
+  try {
+    (New-Object System.Net.WebClient).DownloadFile($launcherSource, $launcherPath)
+  } catch {
+    Write-Host "`t[!] Failed to download '$launcherSource'"
+    Write-Host "`t[!] $_"
+  }
+
+  # We only need a single option now that ida64 supports both 32 bit and 64 bit as of IDA 8.2
```
Ana06, line 48: I think this is a nice ida

[nitpick]
```suggestion
  # ida64.exe supports both 32 bit and 64 bit in IDA >= 8.2
```

packages/idafree.vm/tools/chocolateyinstall.ps1
```diff
@@ -31,15 +31,22 @@ try {
...
-  # The "-WindowStyle hidden" still shows the Powershell Window briefly: https://github.com/PowerShell/PowerShell/issues/3028
-  # We could use the run-hidden wrapper, which won't display the Window but is likely slightly slower.
-  $script = "`$idaExecutable = Get-Item '$Env:programfiles\IDA Pro *\ida.exe' | Select-Object -Last 1; if (!`$idaExecutable) { `$idaExecutable = '$executablePath' }; & `$idaExecutable '%1'"
-  VM-Add-To-Right-Click-Menu $toolName 'Open with IDA' "powershell.exe -WindowStyle hidden `"$script`"" "$menuIcon"
-  # Repeat for x64
-  $script = "`$idaExecutable = Get-Item '$Env:programfiles\IDA Pro *\ida64.exe' | Select-Object -Last 1; if (!`$idaExecutable) { `$idaExecutable = '$executablePath' }; & `$idaExecutable '%1'"
-  VM-Add-To-Right-Click-Menu $toolName-64 'Open with IDA (x64)' "powershell.exe -WindowStyle hidden `"$script`"" "$executablePath"
+
+  # Download ida_launcher.exe to assist with taskbar and right click option and store it in %RAW_TOOLS_DIR%
+  # ida_launcher is a custom binary that searches for the latest IDA installation and executes it
```
Ana06, line 36: [nitpick]
```suggestion
  # Download ida_launcher.exe to assist with taskbar and right click option and store it in %RAW_TOOLS_DIR%
  # ida_launcher.exe is a custom binary that searches for the latest ida64.exe and executes it
```

packages/idafree.vm/tools/chocolateyinstall.ps1
```diff
@@ -31,15 +31,22 @@ try {
...
+  $launcherName = 'ida_launcher.exe'
+  $launcherSource = 'https://raw.githubusercontent.com/mandiant/vm-packages/main/ida_launcher/ida_launcher.exe'
+  $launcherPath = Join-Path ${Env:RAW_TOOLS_DIR} $launcherName
+  Write-Host "[+] Downloading '$launcherName' from '$launcherSource'"
```
Ana06, line 40: [nitpick] I think we don't need a variable for the name. We can remove it from the log message to make it shorter
```suggestion
  $launcherSource = 'https://raw.githubusercontent.com/mandiant/vm-packages/main/ida_launcher/ida_launcher.exe'
  $launcherPath = Join-Path ${Env:RAW_TOOLS_DIR} 'ida_launcher.exe'
  Write-Host "[+] Downloading '$launcherSource'"
```

packages/idafree.vm/tools/chocolateyinstall.ps1
```diff
@@ -31,15 +31,22 @@ try {
...
+  try {
+    (New-Object System.Net.WebClient).DownloadFile($launcherSource, $launcherPath)
+  } catch {
```
Ana06, line 43: I think we shouldn't catch this exception, as the package should fail if we are not able to download the launcher.

I would in addition check that the binary exists after downloading it and before adding it to the registry with `VM-Assert-Path $launcherPath`

Ana06 (changes requested):

packages/idafree.vm/idafree.vm.nuspec
```diff
@@ -2,7 +2,7 @@
 <package xmlns="http://schemas.microsoft.com/packaging/2015/06/nuspec.xsd">
   <metadata>
     <id>idafree.vm</id>
-    <version>8.3</version>
+    <version>8.3.20231128</version>
```
Ana06, line 5:
```suggestion
    <version>8.3.0.20231128</version>
```

We only use the forth segment for the date (not the third one): https://github.com/mandiant/VM-Packages/wiki/Package-Structure#non-metapackages

Ana06 (changes requested):

packages/idafree.vm/tools/chocolateyinstall.ps1
```diff
@@ -31,15 +31,22 @@ try {
...
+    Write-Host "`t[!] Failed to download '$launcherSource'"
+    Write-Host "`t[!] $_"
+  }
+
+  # We only need a single option now that ida64 supports both 32 bit and 64 bit as of IDA 8.2
+  VM-Add-To-Right-Click-Menu $launcherName 'Open with IDA' "`"$launcherPath`" `"%1`"" "$menuIcon"
```
Ana06, line 49: What about the icon for the taskbar?

ida_launcher/src/ida_launcher.c
emtuls: The issue I found with going the `%programfiles%` route was that a user can install IDA anywhere on the system, so this isn't guaranteed to work. Searching just within `%programfiles%` is an assumption we are making, whereas if IDA is installed, it should always be located in the registry (as far as I can tell).

If we're willing to stick with that assumption that IDA should -always- be installed inside of the default location (even IDA Pro for other users down the road), then I will make that change. I just want to confirm that we want to accept that tradeoff before I make the change. :)

emtuls: For this. I opted to stick with the registry, as a test with going the `FindFirstFile` route did not seem to be much cleaner. It also leads to issues with `strcmp` for versioning, where we need to deal with strings like `10.0` being less than `8.3` according to `strcmp` for example. 

While there are clearly ways we can make this work, I don't think the effort is worth it when the `registry` route is working currently, and now that it's been adjusted to look at versioning via the registry (Major and Minor) which uses numbers, the comparison is much more straightforward.

emtuls:
Hmmm. I'm not sure if there is a way to test the file download from github component until things have been merged? Maybe I am wrong, but that's why it is currently failing.

emtuls:
> The testing of idafree.vm fails because https://raw.githubusercontent.com/mandiant/vm-packages/main/ida_launcher/ida_launcher.exe does not exists until we merge this PR. What about we split this PR (leave here only the addition of ida_launcher and in another PR the modifciation to idafree.vm so that this update can be tested).

Complete! New PR for modifying idafree.vm is here: https://github.com/mandiant/VM-Packages/pull/761

Ana06 (approved):
The source code looks good to me. I have checked that this is the code that is in `ida_launcher/ida_launcher.exe`. If `ida_launcher/ida_launcher.exe` is the result of compiling the current source code, please go ahead and merge this PR (and re-trigger the test suite in the new PR)

emtuls:
> The source code looks good to me. I have checked that this is the code that is in `ida_launcher/ida_launcher.exe`. If `ida_launcher/ida_launcher.exe` is the result of compiling the current source code, please go ahead and merge this PR (and re-trigger the test suite in the new PR)

Completed!

PR #761, opened 2023-11-30
Update idafree to use ida_launcher

emtuls:
Split PR from https://github.com/mandiant/VM-Packages/pull/758 so that this PR only includes updates for modifying idafree.vm to add and utilize `ida_launcher.exe` as suggested.

### DIFF
--- a/packages/idafree.vm/tools/chocolateyinstall.ps1
+++ b/packages/idafree.vm/tools/chocolateyinstall.ps1
@@ -18,9 +18,6 @@ try {
 
   $toolDir = Join-Path ${Env:ProgramFiles} "IDA Freeware 8.3" -Resolve
   $executablePath = Join-Path $toolDir "ida64.exe" -Resolve
-  $shortcut = Join-Path $shortcutDir "$toolname.lnk"
-  Install-ChocolateyShortcut -shortcutFilePath $shortcut -targetPath $executablePath
-  VM-Assert-Path $shortcut
 
   Install-BinFile -Name $toolname -Path $executablePath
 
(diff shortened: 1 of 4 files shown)


packages/idafree.vm/tools/chocolateyinstall.ps1
```diff
@@ -30,16 +30,23 @@ try {
     Remove-Item $desktopShortcut -Force -ea 0
   }
 
+  # Download ida_launcher.exe to assist with taskbar and right click option and store it in %RAW_TOOLS_DIR%
+  # ida_launcher.exe is a custom binary that searches for the latest ida64.exe and executes it
+  $launcherName = 'ida_launcher'
+  $launcherSource = 'https://raw.githubusercontent.com/mandiant/vm-packages/main/ida_launcher/ida_launcher.exe'
```
Ana06, line 36: This is what I meant with using a hash commit:
```suggestion
  $launcherSource = 'https://github.com/mandiant/VM-Packages/blob/6416814e4c538c79066c7c830b5c7f9ebcea8be8/ida_launcher/ida_launcher.exe'
```

packages/idafree.vm/tools/chocolateyinstall.ps1
```diff
@@ -30,16 +30,23 @@ try {
...
+  $launcherPath = Join-Path ${Env:RAW_TOOLS_DIR} "$launcherName.exe"
+  Write-Host "[+] Downloading '$launcherSource'"
+  (New-Object System.Net.WebClient).DownloadFile($launcherSource, $launcherPath)
```
Ana06, line 39: I think it would be better to use Chocolatey function to download files. I think they would allow us to check the hash of the binary.

packages/idafree.vm/tools/chocolateyinstall.ps1
```diff
@@ -30,16 +30,23 @@ try {
...
-  $script = "`$idaExecutable = Get-Item '$Env:programfiles\IDA Pro *\ida.exe' | Select-Object -Last 1; if (!`$idaExecutable) { `$idaExecutable = '$executablePath' }; & `$idaExecutable '%1'"
-  VM-Add-To-Right-Click-Menu $toolName 'Open with IDA' "powershell.exe -WindowStyle hidden `"$script`"" "$menuIcon"
-  # Repeat for x64
-  $script = "`$idaExecutable = Get-Item '$Env:programfiles\IDA Pro *\ida64.exe' | Select-Object -Last 1; if (!`$idaExecutable) { `$idaExecutable = '$executablePath' }; & `$idaExecutable '%1'"
-  VM-Add-To-Right-Click-Menu $toolName-64 'Open with IDA (x64)' "powershell.exe -WindowStyle hidden `"$script`"" "$executablePath"
+
+  Install-ChocolateyShortcut -shortcutFilePath $launcherShortcut -targetPath $launcherPath -IconLocation $menuIcon
+
+  # ida64.exe supports both 32 bit and 64 bit in IDA >= 8.2
+  VM-Add-To-Right-Click-Menu $launcherName 'Open with IDA' "`"$launcherPath`" `"%1`"" "$menuIcon"
```
Ana06, line 49: We need to remove these two lines from `packages/idafree.vm/tools/chocolateyuninstall.ps1`:
```
VM-Remove-From-Right-Click-Menu $toolName
VM-Remove-From-Right-Click-Menu $toolName-64
```
I think we should keep the launcher binary and the right-click option as they work without IDAFree (so no changes in this regard).

packages/idafree.vm/tools/chocolateyinstall.ps1
emtuls: I had to keep the `raw.githubusercontent` piece, but I was able to convert it to a specific commit style.

I'm not sure how that might effect things later down the road if the launcher ever gets updated, since it will always pull from this commit instead of from the main branch.

Ana06: The will have to update it using two PR like we did this time. But that's the only way to ensure we are testing the correct version. So I think it is perfect like this 😉

packages/idafree.vm/tools/chocolateyinstall.ps1
```diff
@@ -30,16 +30,24 @@ try {
...
+  $launcherName = 'ida_launcher'
+  $launcherSource = 'https://raw.githubusercontent.com/mandiant/VM-Packages/773649e54aaea62f270c8416cb480020d6475065/ida_launcher/ida_launcher.exe'
+  $launcherPath = Join-Path ${Env:RAW_TOOLS_DIR} "$launcherName.exe"
+  $launcherChecksum = "ebebf8cf01253aab0562b42128f08fdb9b4452f1f49847c51673e5610063a8b5"
+  Write-Host "[+] Downloading '$launcherSource'"
+  Get-ChocolateyWebFile -PackageName $launcherName -FileFullPath $launcherPath -Url $launcherSource -Checksum $launcherChecksum -ChecksumType "sha256"
+
+  VM-Assert-Path $launcherPath
+
+  $launcherShortcut = Join-Path ${Env:RAW_TOOLS_DIR} "$launcherName.lnk"
```
Ana06, line 44: I have just tested this PR and I think there is an important improvement we can make: what about adding the launcher link to the Tools directory as `ida` and do not create the `idafree` one?

Ana06 (changes requested):
Thanks for the changes @emtuls

Ana06 (approved):
I have just tested and it is very cool! 🥳  Thanks for all the work @emtuls! 😄

PR #762, opened 2023-12-01
Remove default --help args

naacbin:
It appears (at least to me) that, by default, arguments should be left empty. Several programs display help by default upon execution, and some lack a "--help" option. Additionally, for certain console applications, executing them directly without arguments is necessary. Therefore, providing `-argument ""` as a parameter seems strange in such cases.
 
N.B. I'm currently working on contributing to VM-Packages by integrating various forensic tools. So you'll probably be receiving several PRs soon.


mr-tz:
Thanks, you don't need Gmail, but a Google account.

Can I sign the Google CLA without a Google account?
A Google account is required to sign the CLA. Google accounts are used to link individual and corporate CLAs to covered contributors and provide authentication for CLA management.

naacbin:
Indeed, didn't know that Gmail and Google account could be detach. It's still a pain to have to provide a phone number.

mr-tz (approved):
would be good to add `--help` again for programs that support it and where it's useful

naacbin:
How can I update the file version as it will not match the executable version ?

naacbin:
For package `Full help isn't displayed but a smaller one` and `Run ...`. I think it's okay as it is.
For package `Doesn't display help...`. I think change is required.
If your good with that I can make the change

mr-tz:
Wow, awesome, thanks a lot for investigating these!!
Regarding the version string, see https://github.com/mandiant/VM-Packages/wiki/Package-Structure#version

> The version needs to be increased every time you modify an existing package. For example, if you fix a bug in a package, you need to update the nuspec version string. Our guidance is to add the current date to the 4th segment using the format YYYYMMDD. For example, 1.6.2 becomes 1.6.2.20220113 and 3.02 becomes 3.02.0.20220113. If the date is already in the 4th segment, you can increment it to the next day (e.g., 20220113 -> 20220114) and when the PR is merged it should be good to go.

naacbin:
- `hashcat.vm`, I test the binary and it's the same situation as `Full help isn't displayed but a smaller one.`
- `inveigh.vm`, What should we do ? Keep the same or remove custom helper argument ?
- `sfextract.vm`, I can't test it.

Ana06:
@naacbin thanks for the PR and happy to hear you are planning to contribute to VM-Packages! Looking forward to see the PRs!

> Several programs display help by default upon execution, and some lack a "--help" option. Additionally, for certain console applications, executing them directly without arguments is necessary. Therefore, providing -argument "" as a parameter seems strange in such cases.

I agree with this. This is a big change (as it affects many packages), so we should test it throughfully. I'll do some testing as soon as I get some time. 😉

naacbin:
> @naacbin thanks for the PR and happy to hear you are planning to contribute to VM-Packages! Looking forward to see the PRs!
> 
> > Several programs display help by default upon execution, and some lack a "--help" option. Additionally, for certain console applications, executing them directly without arguments is necessary. Therefore, providing -argument "" as a parameter seems strange in such cases.
> 
> I agree with this. This is a big change (as it affects many packages), so we should test it throughfully. I'll do some testing as soon as I get some time. 😉

The job has already been done (even if I may have miss something). To extract this list I use `grep -ri 'consoleApp $true'`. The only one that I haven't check is `sfextract.vm`.

I have extract more package employing custom args with `grep -ri 'executableArgs'`
- `apktool.vm`
- `nmap.vm`
- `fakenet-ng.vm`
- `cygwin.vm`
- `windbg.vm`
- `powersploit.vm`
- `yara.vm`

Ana06 (changes requested):
@naacbin I have had finally time to do some testing, sorry for the delay.

- `sfextract` does not need the `--help` argument. So it is fine as it is.

- > `capa.vm` -> Full help isn't displayed but a smaller one.
  
  It displays an error message that includes some help usage:
  Screenshot 2023-12-12 at 10 41 22
  I think we should keep the `--help` argument in capa. The same applies to FLOSS and to innoextract. I would prefer to keep the `--help` argument for the 3 tools. @mr-tz what do you think?

Apart from that, the code looks good to me.  @naacbin thank you for the the work and explanations! Could you please squash the commits in order to keep a clean commit history? 😄

naacbin:
Done ! Indeed, I didn't

Ana06 (commented):
Thanks @naacbin! Could you rebase and fix the conflict in packages/common.vm/common.vm.nuspec? Then we are ready to merge 😉

Ana06 (approved):

Ana06:
Thanks for the contribution, really good job testing all the tools! 🥳

naacbin:
I haven't got the time to squash the last commit... Sorry if this isn't the cleanest history

PR #764, opened 2023-12-04
Refactor 010editor

naacbin:
Concerning EXE installer it's possible to use `/MERGETASKS="!desktopicon"` to avoid creating shortcut on Desktop.

In addition, I was wondering why did you set you own package instead of employing [010editor](https://community.chocolatey.org/packages/010editor) with METAPACKAGE ?

> You may consider to add `https://jrsoftware.org/ishelp/topic_setupcmdline.htm` to documentation it's very useful to create package with EXE installer.

### DIFF
--- a/packages/010editor.vm/010editor.vm.nuspec
+++ b/packages/010editor.vm/010editor.vm.nuspec
@@ -2,7 +2,7 @@
 <package xmlns="http://schemas.microsoft.com/packaging/2011/08/nuspec.xsd">
   <metadata>
     <id>010editor.vm</id>
-    <version>14.0</version>
+    <version>14.0.0.20231204</version>
     <description>Professional text and hex editor with Binary Templates technology.</description>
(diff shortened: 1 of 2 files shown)


Ana06 (commented):
This is a cool change! 😍 I'll test it early next week.

naacbin:
What I meant was that I would prefer to have the installation chain controlled like here, rather than relying on external chocolatey package (except for common packages like npm, python, etc.).

Ana06:
> What I meant was that I would prefer to have the installation chain controlled like here, rather than relying on external chocolatey package (except for common packages like npm, python, etc.).

It is less work to use the metapackages than reimplementing them. So we are using the metapackages as long as there is not a good reason not to. We rely anyway in many URLs outside of our control, so I think trusting the chocolatey community is a good idea. :wink:

Ana06 (approved):
Tested! Thanks @naacbin!

PR #765, opened 2023-12-04
Support innerFolder in create_package_template

naacbin:
Add support of innerFolder in `create_package_template.py`


Ana06 (approved):
Nice addition! 🥳

Issue #161, opened 2022-12-07
Package proposal: systeminformer.vm


seanthegeek:
### Package Name

processhacker

### Tool Name

Process Hacker

### Package type

ZIP_EXE

### Tool's version number

2.39

### Category

Utilities

### Tool's authors

Steven G (dmex), Others

### Tool's description

A multi-purpose tool for monitoring process activity

### Download URL

https://sourceforge.net/projects/processhacker/files/processhacker2/processhacker-2.39-bin.zip/download

### Download SHA256 Hash

2afb5303e191dde688c5626c3ee545e32e52f09da3b35b20f5e0d29a418432f5

### Why is this tool a good addition?

- Discover which processes are using a file
- See what programs have active network connections, and close them if necessary
- View detailed stack traces with kernel-mode, WOW64 and .NET support
- Create, edit and control services

mwilliams31:
[System Informer](https://github.com/winsiderss/systeminformer/) is the successor to Process Hacker. Downloads page: https://systeminformer.sourceforge.io/nightly.php

mr-tz:
@seanthegeek should we add System Informer instead?

MalwareMechanic:
I vote for the System Informer. We *could* also add another shortcut simply named "processhacker" that links to this tool for those that aren't as familiar with this tool's name. Thoughts?

mr-tz:
👍 I don't think it's necessary to add the extra shortcut.

Ana06:
I am not sure how to install the tool. I have tried the downloads from https://systeminformer.sourceforge.io/nightly.php
- I think it is not possible to use the setup download from the command line (it needs user interaction)
- The binary download appear to not include the correct binary (there is no intel 64 only intel 32 and it complains when running it).

@mandiant/flare-vm any ideas?

mr-tz:
Hm, then we may have to defer until they provide proper binaries.

For the interaction part, we could use Autoit (like for npcap) if we really wanted.

Ana06:
https://github.com/winsiderss/si-builds/releases/

Ana06:
https://systeminformer.sourceforge.io/nightly.php includes a binary that may be easy to install

PR #767, opened 2023-12-07
Add systeminformer.vm

d35ha:
Closes https://github.com/mandiant/VM-Packages/issues/161


Ana06 (commented):
Thanks for the PR @d35ha an for the testing @naacbin! 

Shouldn't we run system informer as admin?

- We would need to provide add a `RunAsAdmin` argument to `VM-Install-From-Zip` that we can pass to `VM-Install-Shortcut` which already has this argument.
- Another alternative may be to enable it in System Informer: https://github.com/winsiderss/systeminformer/issues/546

Ana06 (approved):
It seems I checked it wrong when I first tested it. It is running as admin.

PR #771, opened 2023-12-07
Add FTK Imager

naacbin:
Add of FTK Imager, a tool to capture memory capture and mount disk image.

### DIFF
--- a/packages/ftk-imager.vm/ftk-imager.vm.nuspec
+++ b/packages/ftk-imager.vm/ftk-imager.vm.nuspec
@@ -1,0 +1,12 @@
+<?xml version="1.0" encoding="utf-8"?>
+<package xmlns="http://schemas.microsoft.com/packaging/2015/06/nuspec.xsd">
+  <metadata>
+    <id>ftk-imager.vm</id>
+    <version>4.7.1.20231207</version>
+    <authors>AccessData</authors>
+    <description>The tool provides full disk imaging capabilities, mount images, preview data from forensic images, and seamlessly export files in common forensic formats, with an option for RAM capture.</description>
+    <dependencies>
+      <dependency id="common.vm" />
+    </dependencies>
+  </metadata>
+</package>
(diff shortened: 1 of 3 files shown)


packages/ftk-imager.vm/ftk-imager.vm.nuspec
```diff
@@ -0,0 +1,12 @@
+<?xml version="1.0" encoding="utf-8"?>
+<package xmlns="http://schemas.microsoft.com/packaging/2015/06/nuspec.xsd">
+  <metadata>
+    <id>ftk-imager.vm</id>
+    <version>4.7.1.20231207</version>
```
Ana06, line 5: [nitpick] fourth segment is not needed

Ana06 (approved):
Thanks for the PR @naacbin, LGTM

packages/ftk-imager.vm/ftk-imager.vm.nuspec
Ana06: I though the version was 4.7.1 as that is what is in the url. But you are right the tool version is 4.7.1.2, so this version is correct, sorry for the confusion.

packages/ftk-imager.vm/tools/chocolateyinstall.ps1
```diff
@@ -0,0 +1,21 @@
+$ErrorActionPreference = 'Stop'
+Import-Module vm.common -Force -DisableNameChecking
+
+$toolName = 'FTK Imager'
+$category = 'Forensic'
+
+$exeUrl = 'https://d1kpmuwb7gvu1i.cloudfront.net/AccessData_FTK_Imager_4.7.1.exe'
```
Ana06, line 7: Will this URL be used when releasing a new version? I am thinking this because it doesn't include the version forth segment. If this is the case, an update will break the hash and consequently the package.

naacbin: I think the URL will change (even for minor version) but I'm not sure. However, I haven't any other solution and last update of this tools was 2 years ago (2022-01-19), so I don't think we have to worry about like other packages such as Chrome.

Ana06 (approved):

Issue #673, opened 2023-09-29
Create metapackages for common dependencies


Ana06:
Installing several versions of the same software can be an issue (see for example: https://github.com/mandiant/VM-Packages/issues/622). That's why I think we should have metapackages for at least the several tools. This will make it easier in the future to keep them in sync.
## **python3.vm** 
We are currently using python `[3.10.0, 3.11.0)` in libraries.python3.vm, ida.plugin.capa.vm and didier-stevens-suite.vm.

## **openjdk.vm**
We are currently using the open source openjdk in ghidra.vm and dex2jar.vm while we use javaruntime (last version of Java Oracle with a commercial license) in bytecodeviewer.vm. I propose we use a openjdk metapackage for all of them.

I think the only issue is that we need a workaround to open jar files by default.

At the moment we are using openjdk version `[21.0.0]`, but I think we should update it to `[21.0.0, 21.1.0)` to prevent automatic updates. Only the metapackage will be tested when the dependency is updated. But even without the metapackage we are not testing that tools keep working in the CI only that they install successfully.

## **nodejs.vm**
malware-jail.vm and pkg-unpacker (in https://github.com/mandiant/VM-Packages/pull/670) use nodejs version `[20.7.0]`. I propose to update to `[20.7.0, 20.8.0)` to prevent automatic updates for the reasons mention above.

@mandiant/flare-vm opinions? any other tools we should move to a metapackage?

Ana06:
I am thinking we should require libraries.python3.vm in  ida.plugin.capa.vm and didier-stevens-suite.vm (and in general in other python packages) so that there are no conflicts with the basic libraries we install. I am not sure if we then want a python3.vm package that commando-vm can also use to keep the python version in sync as I assume they don't want to install our libraries. Or does it make sense to use the same libraries.python3.vm for both projects?

Ana06:
Documenting discussion with @MalwareMechanic:
>We agree that dependencies should be contained in metapackages to prevent version issues

Ana06:
vcredist140, used by nmap and rpcview is also a good candidate for a metapackage. I suggest to use version `[14.36.32532, 14.37)`.

Ana06:
dotnet-6.vm is another candidate with version `[6.0.400, 6.1)`

PR #778, opened 2023-12-08
Add Zimmermantools

naacbin:
Add Zimmerman forensic tools in .NET6 to improve performance (except hasher which is only available in .NET4).

As JLECmd is included in this PR, #243 can be closed.


Ana06 (commented):
I am not sure if `Forensic` is the best category. I tend to `Utilities` for most of them. I think we may want to use some of these tools in FLARE-VM as well and we don't have a `Forensic` category at the moment (and I think it doesn't make sense for FLARE-VM). @vm-packages opinions?

Ana06:
@mandiant/flare-vm opinions on adding any of these tools to FLARE-VM?

naacbin:
> I am not sure if `Forensic` is the best category. I tend to `Utilities` for most of them. I think we may want to use some of these tools in FLARE-VM as well and we don't have a `Forensic` category at the moment (and I think it doesn't make sense for FLARE-VM). @vm-packages opinions?

All of the tools are part of ZimmemanTools suite which are dedicated forensic tools, so it makes sense to be in `Forensic` category (which exist `yara` is part of it).

I don't think it should be included in a FLARE-VM or COMMANDO-VM by default but still need to be in VM-Packages as you can create your custom VM with the [config.xml](https://github.com/mandiant/flare-vm/blob/main/config.xml).

Ana06:
@naacbin 
> which exist yara is part of it

you are right, we already have the `Forensinc` folder in FLARE-VM.

> I don't think it should be included in a FLARE-VM or COMMANDO-VM by default but still need to be in VM-Packages as you can create your custom VM with the [config.xml](https://github.com/mandiant/flare-vm/blob/main/config.xml).

But it people think these are important/useful tools we should add them in order to provide a good default set of tools. JFYI, the default config file in FLARE-VM is what the FLARE team uses for its internal builds, the set of tools that we think a reverse engineer needs to do their job efficiently. 😉

Ana06 (approved):

naacbin:
3 tools could be of interest to Flare-VM :
- bstrings -> strings with regex
- EZViewer -> Application to open office document (xlsx, docx, etc.)
- Registry Explorer -> Better explorer of registry (without edition option)

packages/amcacheparser.vm/amcacheparser.vm.nuspec
```diff
@@ -0,0 +1,13 @@
+<?xml version="1.0" encoding="utf-8"?>
+<package xmlns="http://schemas.microsoft.com/packaging/2015/06/nuspec.xsd">
+  <metadata>
+    <id>amcacheparser.vm</id>
+    <version>1.5.1.20231208</version>
+    <authors>Eric Zimmerman</authors>
+    <description>Amcache.hve parser with lots of extra features. Handles locked files</description>
+    <dependencies>
+      <dependency id="common.vm" />
+      <dependency id="dotnet-6.0-runtime" version="[6.0.25]" />
```
Ana06, line 10: I think we should introduce a dotnet-6.vm package (see https://github.com/mandiant/VM-Packages/issues/673#issuecomment-1859977796) to ensure we don't require several version of the same tool and to avoid constant dependency updates with our test suite. We should use this package in all the tools that require dotnet version 6 (the ones in this PR and [sfextract.vm](https://github.com/mandiant/VM-Packages/blob/e4cf21e70e9a8dce58449e74f49184910232648c/packages/sfextract.vm/sfextract.vm.nuspec#L10))

packages/hasher.vm/hasher.vm.nuspec
```diff
@@ -0,0 +1,12 @@
+<?xml version="1.0" encoding="utf-8"?>
+<package xmlns="http://schemas.microsoft.com/packaging/2015/06/nuspec.xsd">
+  <metadata>
+    <id>hasher.vm</id>
+    <version>2.0.0.20231207</version>
+    <authors>Eric Zimmerman</authors>
+    <description>Hash all the things</description>
+    <dependencies>
+      <dependency id="common.vm" />
```
Ana06, line 9: isn't this package missing the .NET dependency?

packages/bstrings.vm/tools/chocolateyinstall.ps1
```diff
@@ -0,0 +1,10 @@
+$ErrorActionPreference = 'Stop'
+Import-Module vm.common -Force -DisableNameChecking
+
+$toolName = 'bstrings'
+$category = 'Forensic'
```
Ana06, line 5: We have similar tools like floss in `Utilities` instead of `Forensic`. I would add this tool in `Utilities`. Alternatively, we could consider introducing a `Strings` category as we have a few tools related with extracting strings. @vm-packages opinions?

packages/ezviewer.vm/tools/chocolateyinstall.ps1
```diff
@@ -0,0 +1,10 @@
+$ErrorActionPreference = 'Stop'
+Import-Module vm.common -Force -DisableNameChecking
+
+$toolName = 'EZViewer'
+$category = 'Forensic'
```
Ana06, line 5:
```suggestion
$category = 'Office'
```

packages/hasher.vm/tools/chocolateyuninstall.ps1
```diff
@@ -0,0 +1,7 @@
+$ErrorActionPreference = 'Continue'
+Import-Module vm.common -Force -DisableNameChecking
+
+$toolName = 'Hasher'
+$category = 'Forensic'
```
Ana06, line 5: We have similar tools in `Utilities` instead of `Forensic`.

Ana06 (changes requested):
I have the feeling some of the tools in this PR could be in a more concrete category than `Forensic`. @vm-packages opinions?

@mandiant/flare-vm opinions on adding any of this tools to the default FLARE-VM config?

packages/amcacheparser.vm/amcacheparser.vm.nuspec
naacbin: #783 require dotnet 6 too

emtuls:
Looking at the 3 tools suggested as potential for Flare-VM, these are my thoughts:
- `bstrings` - I'm indifferent on this. Possibly useful, but I don't personally think I'd use it.
- `EZViewer` - This looks potentially useful for document analysis, though I don't do enough document analysis to be able to determine how well it might handle actual malicious docs like how Didier Stevens tools work. I'm okay with adding it though.
- `Registry Explorer` - I think this is a very useful tool and I am for it.


On another note @Ana06, we have an `Office` category, but no tools added for it. Would changing the category name to something like `Documents` be better? I think it might encompass more and wouldn't silo it into just "MS Office" things, while still being inclusive of them.

KJaeYoung:
> Thanks for the input @emtuls! Apart from EZViewer, we are planing to add some more tools that may go in the `Office`/`Documents` category:
> 
>     * [New package: onedump.vm #788](https://github.com/mandiant/VM-Packages/issues/788)
> 
>     * [New package: xmlmacrodeobfuscator #790](https://github.com/mandiant/VM-Packages/issues/790)
> 
>     * [New package: offvis.vm #791](https://github.com/mandiant/VM-Packages/issues/791)
> 
> 
> I like the idea to change the `Office` category to `Documents`.
> 
> Adding @KJaeYoung as he has a lot of malicious document experience. @KJaeYoung do you have a name preference for these kind of tools (`Office` vs `Documents`)? What do you think about adding EZViewer to the default installation of FLARE-VM?
> 
> The current categories are: https://github.com/mandiant/VM-Packages/blob/main/categories.txt

All three tools going into the Office category seems fine to me. I have not used EZViewer before. I will give it a shot and evaluate it and get back to you.

packages/bstrings.vm/tools/chocolateyinstall.ps1
naacbin: I change the category to `Utilities`

packages/ezviewer.vm/tools/chocolateyinstall.ps1
naacbin: I change the category to `Office`

packages/hasher.vm/tools/chocolateyuninstall.ps1
naacbin: I change the category to `Utilities`

packages/amcacheparser.vm/amcacheparser.vm.nuspec
naacbin: Done in #812

emtuls:
Following up from https://github.com/mandiant/VM-Packages/pull/903

A few issues to address which are all just category mismatches:
1. `bstrings` is listed as `Utilities` in the `chocolateyinstall` and `Forensic` in `chocolatelyuninstall`.
2. `EZViewer` is listed as `Office` in the `chocolateyinstall` and `Forensic` in `chocolatelyuninstall`.
3. `Hasher` is listed as `Forensic` in the `chocolateyinstall` and `Utilities` in `chocolatelyuninstall`.

I'm indifferent on which category `bstrings` should be in. I was looking to try to minimize throwing everything into `Utilities` if we can help it, but `bstrings` also doesn't seem to fit well in other categories (at the moment), other than possibly `Forensic`, so I'd be okay with it being in either.

As far as ones that fall into `Office`, if we're suggesting changing the category from `Office` to `Documents` (which I think is better), then those would likely need to be changed.

`RegistryExplorer` can now fall into the Registry category instead of `Forensic`.

naacbin:
> Following up from #903
> 
> A few issues to address which are all just category mismatches:
> 
>     1. `bstrings` is listed as `Utilities` in the `chocolateyinstall` and `Forensic` in `chocolatelyuninstall`.
> 
>     2. `EZViewer` is listed as `Office` in the `chocolateyinstall` and `Forensic` in `chocolatelyuninstall`.
> 
>     3. `Hasher` is listed as `Forensic` in the `chocolateyinstall` and `Utilities` in `chocolatelyuninstall`.
> 
> 
> I'm indifferent on which category `bstrings` should be in. I was looking to try to minimize throwing everything into `Utilities` if we can help it, but `bstrings` also doesn't seem to fit well in other categories (at the moment), other than possibly `Forensic`, so I'd be okay with it being in either.
> 
> As far as ones that fall into `Office`, if we're suggesting changing the category from `Office` to `Documents` (which I think is better), then those would likely need to be changed.
> 
> `RegistryExplorer` can now fall into the Registry category instead of `Forensic`.

I fixed the category mismatches. However, has I think this PR can be merged before #903, I didn't change categories of `EZViewer`, `Hasher` and `RegistryExplorer`. I will do it in your PR directly. If it's not the case I'll modify this one.
- `EZViewer` -> `Documents`
- `Hasher` -> `File Information`
- `RegistryExplorer` -> `Registry`

emtuls:
If this one gets merged first, I can modify them to https://github.com/mandiant/VM-Packages/pull/903, so no worries. 🙂

Thank you for pushing through with this one!

Ana06:
@vm-packages any objection on merging this PR and addressing categories discussions/changes in #903 afterwards?

Ana06 (approved):

Issue #469, opened 2023-06-28
do not check the hash for signed packages


Ana06:
Microsoft often distribute tools with a URL that doesn't include the version making the package break with every new update (as the hash changes). But I assume Microsoft signs the tools and we could use the signature as versification instead of the hash. This would prevent packages like sysinternals from breaking. @mandiant/flare-vm what do you think? Any ideas of how we can implement this? 🤔

day1player:
This will be a relevant topic for the upcoming metasploit package as well. Metasploit releases "latest" msi installers every week.. Is this not something that CI/CD can check for us? If the hash is a mismatch, could we submit a PR?

mr-tz:
1. If we can download and install packages without any hash we could do that for select packages (like the ones mentioned)
2. Alternatively, we create a CI step like @day1player suggests

2. sounds more correct, but 1. could be much easier

Ana06:
> This will be a relevant topic for the upcoming metasploit package as well. Metasploit releases "latest" msi installers every week.. Is this not something that CI/CD can check for us? If the hash is a mismatch, could we submit a PR?

We could implement a CI that updates the hash (similar to the current update action) but still the package is broken for some time (from the moment it breaks after the PR is generated and merged), which is not ideal.  For community packages, the community updates the hash and CI updates the community package (for example sysinternals and chrome), but still the package is broken for some time.

Users have already the option to use the `--ignore-checksums` if they don't want to check the hash. 

**Checking signatures instead of hashes for signed packages would be IMO the best option. We probably can use [`sigverif.exe`](https://learn.microsoft.com/en-us/windows/win32/seccrypto/using-signtool-to-verify-a-file-signature) for this.**

day1player:
@Ana06 I looked into this as well, there is a powershell function [`Get-AuthenticodeSignature`](https://learn.microsoft.com/en-us/powershell/module/microsoft.powershell.security/get-authenticodesignature?view=powershell-7.3) I was going to play around with we could probably wrap into an assert function.

Ana06:
Packages that would benefit from this: sysinternal, cyberchef

Ana06:
## Status summary
To address this issue we need to:
- Create a helper in common.vm to download/install a file checking its signature instead of the hash with chocolatey
- Use this helper in sysinternals.vm
- Create googlechrome.vm that uses this helper to install chrome. (then we can re-add the dependency in cyberchef: https://github.com/mandiant/VM-Packages/issues/296)

PR #779, opened 2023-12-11
Add googlechrome to VM-Packages

binjo:
To address issue #469, added:
1. download/install Chrome by checking if installer's signature is valid
2. add associations for opening url/html file in Chrome
3. disable Chrome auto-update related services and registries

### DIFF
--- a/packages/googlechrome.vm/googlechrome.vm.nuspec
+++ b/packages/googlechrome.vm/googlechrome.vm.nuspec
@@ -1,0 +1,21 @@
+<?xml version="1.0" encoding="utf-8"?>
+<package xmlns="http://schemas.microsoft.com/packaging/2015/06/nuspec.xsd">
+  <metadata>
+    <id>googlechrome.vm</id>
+    <version>0.0.0.20231210</version>
+    <authors>Mandiant</authors>
+    <description><![CDATA[
+Chrome is a fast, simple, and secure web browser, built for the modern web.
+
+### Notes
+
+- This package uses Chrome's administrative MSI installer and installs the 32-bit on 32-bit OSes and the 64-bit version on 64-bit OSes. If this package is installed on a 64-bit OS and the 32-bit version of Chrome is already installed, the package keeps installing/updating the 32-bit version of Chrome.
+- This package always installs the latest version of Google Chrome, regardless of the version specified in the package. Google does not officially offer older versions of Chrome for download. Because of this you may get checksum mismatch between the time Google releases a new installer, and the package is automatically updated.
+- **If the package is out of date please check [Version History](#versionhistory) for the latest submitted version. If you have a question, please ask it in [Chocolatey Community Package Discussions](https://github.com/chocolatey-community/chocolatey-packages/discussions) or raise an issue on the [Chocolatey Community Packages Repository](https://github.com/chocolatey-community/chocolatey-packages/issues) if you have problems with the package. Disqus comments will generally not be responded to.**
+]]></description>
+    <dependencies>
+      <dependency id="chocolatey-core.extension" />
+      <dependency id="common.vm" />
+    </dependencies>
+  </metadata>
+</package>
--- a/packages/googlechrome.vm/googlechrome.vm.nuspec
+++ b/packages/googlechrome.vm/googlechrome.vm.nuspec
@@ -4,15 +4,7 @@
     <id>googlechrome.vm</id>
     <version>0.0.0.20231210</version>
     <authors>Mandiant</authors>
-    <description><![CDATA[
-Chrome is a fast, simple, and secure web browser, built for the modern web.
-
-### Notes
-
-- This package uses Chrome's administrative MSI installer and installs the 32-bit on 32-bit OSes and the 64-bit version on 64-bit OSes. If this package is installed on a 64-bit OS and the 32-bit version of Chrome is already installed, the package keeps installing/updating the 32-bit version of Chrome.
-- This package always installs the latest version of Google Chrome, regardless of the version specified in the package. Google does not officially offer older versions of Chrome for download. Because of this you may get checksum mismatch between the time Google releases a new installer, and the package is automatically updated.
-- **If the package is out of date please check [Version History](#versionhistory) for the latest submitted version. If you have a question, please ask it in [Chocolatey Community Package Discussions](https://github.com/chocolatey-community/chocolatey-packages/discussions) or raise an issue on the [Chocolatey Community Packages Repository](https://github.com/chocolatey-community/chocolatey-packages/issues) if you have problems with the package. Disqus comments will generally not be responded to.**
-]]></description>
+    <description>Chrome is a fast, simple, and secure web browser, built for the modern web.</description>
     <dependencies>
       <dependency id="chocolatey-core.extension" />
       <dependency id="common.vm" />
--- a/packages/googlechrome.vm/googlechrome.vm.nuspec
+++ b/packages/googlechrome.vm/googlechrome.vm.nuspec
@@ -6,7 +6,6 @@
     <authors>Mandiant</authors>
     <description>Chrome is a fast, simple, and secure web browser, built for the modern web.</description>
     <dependencies>
-      <dependency id="chocolatey-core.extension" />
       <dependency id="common.vm" />
     </dependencies>
   </metadata>
--- a/packages/googlechrome.vm/tools/chocolateyinstall.ps1
+++ b/packages/googlechrome.vm/tools/chocolateyinstall.ps1
@@ -48,7 +48,8 @@ try {
         }
     } else {
         $file = $packageArgs['file']
-        VM-Write-Log "WARN" "`tSignature for downloaded $file is not valid"
+        VM-Write-Log "ERROR" "Signature for downloaded $file is not valid"
+        exit 1
     }
 } catch {
     VM-Write-Log-Exception $_
--- a/packages/googlechrome.vm/tools/chocolateyinstall.ps1
+++ b/packages/googlechrome.vm/tools/chocolateyinstall.ps1
@@ -1,44 +1,42 @@
 $ErrorActionPreference = 'Stop'
 Import-Module vm.common -Force -DisableNameChecking
 
-$toolsPath = Split-Path $MyInvocation.MyCommand.Definition
-. $toolsPath\sfta.ps1
-. $toolsPath\helpers.ps1
-
-$downloadDir = Get-PackageCacheLocation
-$installer          = 'googlechromeinstaller.msi'
-$packageArgs        = @{
-    packageName     = $env:ChocolateyPackageName
-    file            = Join-Path $downloadDir $installer
-    url             = 'https://dl.google.com/tag/s/dl/chrome/install/googlechromestandaloneenterprise.msi'
-    url64bit        = 'https://dl.google.com/tag/s/dl/chrome/install/googlechromestandaloneenterprise64.msi'
-    fileType        = 'MSI'
-    silentArgs      = "/quiet /norestart /l*v `"$($env:TEMP)\$($env:chocolateyPackageName).$($env:chocolateyPackageVersion).MsiInstall.log`""
-    validExitCodes  = @(0)
-}
-
-# Download the installer
-$packageArgs['file'] = Get-ChocolateyWebFile @packageArgs
-
-Start-Sleep 2
-
-$sigValid = (Get-AuthenticodeSignature -FilePath $packageArgs['file']).Status -eq 'Valid'
-
 try {
+    $toolsPath = Split-Path $MyInvocation.MyCommand.Definition
+    . $toolsPath\sfta.ps1
+    . $toolsPath\helpers.ps1
+
+    $downloadDir = Get-PackageCacheLocation
+    $installer          = 'googlechromeinstaller.msi'
+    $packageArgs        = @{
+        packageName     = $env:ChocolateyPackageName
+        file            = Join-Path $downloadDir $installer
+        url             = 'https://dl.google.com/tag/s/dl/chrome/install/googlechromestandaloneenterprise.msi'
+        url64bit        = 'https://dl.google.com/tag/s/dl/chrome/install/googlechromestandaloneenterprise64.msi'
+        fileType        = 'MSI'
+        silentArgs      = "/quiet /norestart /l*v `"$($env:TEMP)\$($env:chocolateyPackageName).$($env:chocolateyPackageVersion).MsiInstall.log`""
+        validExitCodes  = @(0)
+    }
+
+    # Download the installer
+    $packageArgs['file'] = Get-ChocolateyWebFile @packageArgs
+
+    $sigValid = (Get-AuthenticodeSignature -FilePath $packageArgs['file']).Status -eq 'Valid'
+
     if ($sigValid) {
         $toolName = 'Google Chrome'
         $category = 'Utilities'
 
-        VM-Write-Log "INFO" "Signature valid ... installing $category"
+        VM-Write-Log "INFO" "Signature valid ... installing as $category" # for lint check
         Install-ChocolateyInstallPackage @packageArgs
 
-        VM-Write-Log "INFO" "`tSetup associations"
+        VM-Write-Log "INFO" "Setup associations"
         Set-PTA ChromeHTML http
         Set-PTA ChromeHTML https
         Set-FTA ChromeHTML .htm
         Set-FTA ChromeHTML .html
 
-        VM-Write-Log "INFO" "`tDisable update settings"
+        VM-Write-Log "INFO" "Disable update settings"
         Disable-Chrome-Updates
 
         # Delete Desktop shortcut
--- a/packages/googlechrome.vm/tools/chocolateyinstall.ps1
+++ b/packages/googlechrome.vm/tools/chocolateyinstall.ps1
@@ -1,0 +1,55 @@
+$ErrorActionPreference = 'Stop'
+Import-Module vm.common -Force -DisableNameChecking
+
+$toolsPath = Split-Path $MyInvocation.MyCommand.Definition
+. $toolsPath\sfta.ps1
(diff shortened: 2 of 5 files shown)


packages/googlechrome.vm/googlechrome.vm.nuspec
```diff
@@ -0,0 +1,21 @@
...
+
+### Notes
+
+- This package uses Chrome's administrative MSI installer and installs the 32-bit on 32-bit OSes and the 64-bit version on 64-bit OSes. If this package is installed on a 64-bit OS and the 32-bit version of Chrome is already installed, the package keeps installing/updating the 32-bit version of Chrome.
+- This package always installs the latest version of Google Chrome, regardless of the version specified in the package. Google does not officially offer older versions of Chrome for download. Because of this you may get checksum mismatch between the time Google releases a new installer, and the package is automatically updated.
+- **If the package is out of date please check [Version History](#versionhistory) for the latest submitted version. If you have a question, please ask it in [Chocolatey Community Package Discussions](https://github.com/chocolatey-community/chocolatey-packages/discussions) or raise an issue on the [Chocolatey Community Packages Repository](https://github.com/chocolatey-community/chocolatey-packages/issues) if you have problems with the package. Disqus comments will generally not be responded to.**
+]]></description>
```
Ana06, line 15:
```suggestion
    <description>Chrome is a fast, simple, and secure web browser, built for the modern web.</description>
```
From [our documentation](https://github.com/mandiant/VM-Packages/wiki/Package-Structure#description):
> The description should be short and not include version specific details or other information that is likely to change in a future version.

packages/googlechrome.vm/googlechrome.vm.nuspec
```diff
@@ -0,0 +1,21 @@
...
+    <dependencies>
+      <dependency id="chocolatey-core.extension" />
```
Ana06, line 17: [question] Why do we need this dependency?

packages/googlechrome.vm/tools/chocolateyinstall.ps1
```diff
@@ -0,0 +1,55 @@
...
+}
+
+# Download the installer
+$packageArgs['file'] = Get-ChocolateyWebFile @packageArgs
+
+Start-Sleep 2
+
+$sigValid = (Get-AuthenticodeSignature -FilePath $packageArgs['file']).Status -eq 'Valid'
+
+try {
```
Ana06, line 27: we should move this `try` to the top (before executing the two helper files) instead some of the previous commands fail

packages/googlechrome.vm/tools/chocolateyinstall.ps1
```diff
@@ -0,0 +1,55 @@
...
+    url64bit        = 'https://dl.google.com/tag/s/dl/chrome/install/googlechromestandaloneenterprise64.msi'
+    fileType        = 'MSI'
+    silentArgs      = "/quiet /norestart /l*v `"$($env:TEMP)\$($env:chocolateyPackageName).$($env:chocolateyPackageVersion).MsiInstall.log`""
+    validExitCodes  = @(0)
+}
+
+# Download the installer
+$packageArgs['file'] = Get-ChocolateyWebFile @packageArgs
+
+Start-Sleep 2
```
Ana06, line 23: [question] do we run into problems with the downloaded file not being present it we remove this line?

packages/googlechrome.vm/tools/chocolateyinstall.ps1
```diff
@@ -0,0 +1,55 @@
...
+
+$sigValid = (Get-AuthenticodeSignature -FilePath $packageArgs['file']).Status -eq 'Valid'
+
+try {
+    if ($sigValid) {
+        $toolName = 'Google Chrome'
+        $category = 'Utilities'
```
Ana06, line 30: Are you using the category? I think we don't need to add Chrome to any category, just install it in the system, as it is a dependency for other tools (cyberchef) not really an analysis tool.

packages/googlechrome.vm/tools/chocolateyinstall.ps1
```diff
@@ -0,0 +1,55 @@
...
+        Disable-Chrome-Updates
+
+        # Delete Desktop shortcut
+        $desktopShortcut = Join-Path ${Env:Public} "Desktop\$toolName.lnk"
+        if (Test-Path $desktopShortcut) {
+            Remove-Item $desktopShortcut -Force -ea 0
+        }
+    } else {
+        $file = $packageArgs['file']
+        VM-Write-Log "WARN" "`tSignature for downloaded $file is not valid"
```
Ana06, line 51: We should fail the package if the signature is not valid. 
```suggestion
        VM-Write-Log "ERROR" "Signature for downloaded $file is not valid"
        exit 1
```

packages/googlechrome.vm/tools/chocolateyinstall.ps1
```diff
@@ -0,0 +1,55 @@
...
+Start-Sleep 2
+
+$sigValid = (Get-AuthenticodeSignature -FilePath $packageArgs['file']).Status -eq 'Valid'
+
+try {
+    if ($sigValid) {
+        $toolName = 'Google Chrome'
+        $category = 'Utilities'
+
+        VM-Write-Log "INFO" "`tSignature valid ... installing $category"
```
Ana06, line 32:
```suggestion
        VM-Write-Log "INFO" "Signature valid ... installing $category"
```

I think we need to remove the tab in all messages. We use them to display a sub message of a related message displayed before.

Ana06 (changes requested):
Thanks for all the work @binjo!

packages/googlechrome.vm/googlechrome.vm.nuspec
binjo: the `Get-PackageCacheLocation` method is exported from `chocolatey-core.extension`

packages/googlechrome.vm/tools/chocolateyinstall.ps1
```diff
@@ -0,0 +1,54 @@
...
+        $category = 'Utilities'
+
+        VM-Write-Log "INFO" "Signature valid ... installing as $category" # for lint check
+        Install-ChocolateyInstallPackage @packageArgs
+
+        VM-Write-Log "INFO" "Setup associations"
+        Set-PTA ChromeHTML http
+        Set-PTA ChromeHTML https
+        Set-FTA ChromeHTML .htm
+        Set-FTA ChromeHTML .html
```
Ana06, line 37: I think this is not working perfectly. After installing the package, opening an `.html` file shows:
Screenshot 2023-12-12 at 12 14 22

packages/googlechrome.vm/tools/chocolateyinstall.ps1
```diff
@@ -0,0 +1,54 @@
...
+
+        VM-Write-Log "INFO" "Disable update settings"
+        Disable-Chrome-Updates
+
+        # Delete Desktop shortcut
+        $desktopShortcut = Join-Path ${Env:Public} "Desktop\$toolName.lnk"
```
Ana06, line 43: I think the correct path is
```suggestion
        $desktopShortcut = Join-Path ${Env:UserProfile} "Desktop\$toolName.lnk"
```

packages/googlechrome.vm/tools/chocolateyuninstall.ps1
```diff
@@ -0,0 +1,7 @@
+$ErrorActionPreference = 'Continue'
+Import-Module vm.common -Force -DisableNameChecking
+
+$toolName = 'Chrome'
+$category = 'Utilities'
+
+VM-Remove-Tool-Shortcut $toolName $category
```
Ana06, line 7: There is no shortcut to remove. In `chocolateyuninstall.ps1` you should uninstall GoogleChrome if it is easy to do. If it is difficult, we can remove the uninstaller.

packages/googlechrome.vm/tools/chocolateyinstall.ps1
Ana06: it is still here 👀

packages/googlechrome.vm/tools/chocolateyinstall.ps1
```diff
@@ -0,0 +1,54 @@
...
+        url64bit        = 'https://dl.google.com/tag/s/dl/chrome/install/googlechromestandaloneenterprise64.msi'
+        fileType        = 'MSI'
+        silentArgs      = "/quiet /norestart /l*v `"$($env:TEMP)\$($env:chocolateyPackageName).$($env:chocolateyPackageVersion).MsiInstall.log`""
+        validExitCodes  = @(0)
+    }
+
+    # Download the installer
+    $packageArgs['file'] = Get-ChocolateyWebFile @packageArgs
+
+    $sigValid = (Get-AuthenticodeSignature -FilePath $packageArgs['file']).Status -eq 'Valid'
```
Ana06, line 24: I wonder if it is enough with checking that it has a valid signature of if we need to check that the signer certificate comes from Google Chrome as well? @vm-packages opinions? (We can extract this signature to a helper for other tools in the future, so it is a general question)

packages/googlechrome.vm/tools/chocolateyinstall.ps1
binjo: `$Env:Public` is correct one, the installer created `.lnk` is in the public folder, however I just noticed when open the Chrome for the first time, it will create another `.lnk` which is in `$Env:UserProfile` 😂

packages/googlechrome.vm/tools/chocolateyinstall.ps1
binjo: my testing env seems working, after installing `cyberchef.vm`, the default opening `.html` file is Chrome without the prompt as yours.

packages/googlechrome.vm/tools/chocolateyinstall.ps1
day1player: I think this would be a good opportunity to add a `VM-Delete-Desktop-Shortcut` function that just checks both directories. Maybe not required for this PR but that function has been on my list for a while

packages/googlechrome.vm/tools/chocolateyinstall.ps1
day1player: This is another function that has been on my list; `VM-Assert-Signature`. I think we should maybe open a PR for this function and hash it out there

day1player: Personally I think I'd be ok with just checking if it is Valid since that will check it against CAs, so self signed binaries wont be shown as valid

packages/googlechrome.vm/tools/chocolateyinstall.ps1
binjo: I just tested again, the installer created `.lnk` file is in the public folder, after install `cyberchef.vm` and open the `.html` file, a new Chrome `.lnk` shortcut file appeared in `$Env:UserProfile`. This looks infeasible for post install deletion.

packages/googlechrome.vm/tools/chocolateyinstall.ps1
Ana06: @day1player I think we should either introduce `VM-Assert-Signature` or a helper function to install tools using the signature instead of the hash.

packages/googlechrome.vm/tools/helpers.ps1
```diff
@@ -0,0 +1,51 @@
...
+            $Arguments.Add("Path", $regGoogleUpdatePath)
+        }
+        $Arguments.Add("Force", $true)
+        New-ItemProperty @Arguments | Out-Null
+    }
+
+}
+
+# https://github.com/chocolatey-community/chocolatey-packages/blob/master/extensions/chocolatey-core.extension/extensions/Get-PackageCacheLocation.ps1
+function Get-PackageCacheLocation {
```
Ana06, line 38: why is this function needed? I suggest using `Join-Path ${Env:TEMP} $installerName` directly as we do in other packages to avoid having to maintain this code. 

If we would want to do this differently in all packages, then the helper function should go into common.

packages/googlechrome.vm/tools/chocolateyinstall.ps1
Ana06: I have tested again in a tested environment and opening the cyberchef lnk to the html file shows this pop up. @binjo how is your testing environment set up?

packages/googlechrome.vm/tools/chocolateyinstall.ps1
Ana06: We could open chrome in the installer, close it, and then delete the lnk.

packages/googlechrome.vm/googlechrome.vm.nuspec
```diff
@@ -0,0 +1,12 @@
+<?xml version="1.0" encoding="utf-8"?>
+<package xmlns="http://schemas.microsoft.com/packaging/2015/06/nuspec.xsd">
+  <metadata>
+    <id>googlechrome.vm</id>
+    <version>0.0.0.20231210</version>
+    <authors>Mandiant</authors>
+    <description>Chrome is a fast, simple, and secure web browser, built for the modern web.</description>
```
Ana06, line 7: [nitpick] I would short this description

packages/googlechrome.vm/googlechrome.vm.nuspec
```diff
@@ -0,0 +1,12 @@
+<?xml version="1.0" encoding="utf-8"?>
+<package xmlns="http://schemas.microsoft.com/packaging/2015/06/nuspec.xsd">
+  <metadata>
+    <id>googlechrome.vm</id>
+    <version>0.0.0.20231210</version>
+    <authors>Mandiant</authors>
```
Ana06, line 6: This field should contain the authors of the tool
```suggestion
    <authors>Google LLC.</authors>
```

Ana06 (changes requested):

packages/googlechrome.vm/tools/chocolateyinstall.ps1
```diff
@@ -0,0 +1,54 @@
...
+try {
+    $toolsPath = Split-Path $MyInvocation.MyCommand.Definition
+    . $toolsPath\sfta.ps1
+    . $toolsPath\helpers.ps1
+
+    $downloadDir = Get-PackageCacheLocation
+    $installer          = 'googlechromeinstaller.msi'
+    $packageArgs        = @{
+        packageName     = $env:ChocolateyPackageName
+        file            = Join-Path $downloadDir $installer
```
Ana06, line 13: According to the documentation `Get-ChocolateyWebFile` needs `fileFullPath`. Maybe it works because it is the second argument. 🤔

packages/googlechrome.vm/tools/chocolateyinstall.ps1
```diff
@@ -0,0 +1,54 @@
+$ErrorActionPreference = 'Stop'
+Import-Module vm.common -Force -DisableNameChecking
+
+try {
+    $toolsPath = Split-Path $MyInvocation.MyCommand.Definition
+    . $toolsPath\sfta.ps1
+    . $toolsPath\helpers.ps1
+
+    $downloadDir = Get-PackageCacheLocation
+    $installer          = 'googlechromeinstaller.msi'
```
Ana06, line 10: [nitpick] I would remove this variable as it is only used once

packages/googlechrome.vm/tools/chocolateyinstall.ps1
```diff
@@ -0,0 +1,54 @@
...
+    $packageArgs        = @{
+        packageName     = $env:ChocolateyPackageName
+        file            = Join-Path $downloadDir $installer
+        url             = 'https://dl.google.com/tag/s/dl/chrome/install/googlechromestandaloneenterprise.msi'
+        url64bit        = 'https://dl.google.com/tag/s/dl/chrome/install/googlechromestandaloneenterprise64.msi'
+        fileType        = 'MSI'
+        silentArgs      = "/quiet /norestart /l*v `"$($env:TEMP)\$($env:chocolateyPackageName).$($env:chocolateyPackageVersion).MsiInstall.log`""
+        validExitCodes  = @(0)
```
Ana06, line 18:
```suggestion
```

I think we should use different arguments for `Install-ChocolateyInstallPackage` and `Get-ChocolateyWebFile` as they don't really overlap and it makes more difficult to read the code.

Ana06, line 18: `validExitCodes  = @(0)`is the default, so it can be removed

packages/googlechrome.vm/tools/chocolateyinstall.ps1
```diff
@@ -0,0 +1,54 @@
...
+        Install-ChocolateyInstallPackage @packageArgs
+
+        VM-Write-Log "INFO" "Setup associations"
+        Set-PTA ChromeHTML http
+        Set-PTA ChromeHTML https
+        Set-FTA ChromeHTML .htm
+        Set-FTA ChromeHTML .html
+
+        VM-Write-Log "INFO" "Disable update settings"
+        Disable-Chrome-Updates
```
Ana06, line 40: @binjo disabling updates was not part of the open issues related to chrome, why do you think is needed?

Does anyone has any concern with possible side-effects? 🤔

Ana06 (changes requested):

Ana06:
I have extracted the signature validation and creation of chrome https://github.com/mandiant/VM-Packages/pull/824 to this PR: https://github.com/mandiant/VM-Packages/pull/824

I have extracted the rest of the discussions in this PR to the following issues:
- https://github.com/mandiant/VM-Packages/issues/820
- https://github.com/mandiant/VM-Packages/issues/821
- https://github.com/mandiant/VM-Packages/issues/822
- https://github.com/mandiant/VM-Packages/issues/823

I think this will make the review/discussion process a bit easier. 😉

Issue #786, opened 2023-12-13
Introduce open Cmd Here context menu


Ana06:
### Details

Introduce the following functionality to the Windows 10 debloat package: `Right-click Explorer > Open Cmd Here context menu`

Resource: https://www.windowscentral.com/add-open-command-window-here-back-context-menu-windows-10

@mandiant/commando-vm is this something you want? Or should we add it in the FLARE-VM config instead?

PR #796, opened 2023-12-18
Enable openning cmd from context menu

d35ha:
Closes https://github.com/mandiant/VM-Packages/issues/786

### DIFF
--- a/packages/debloat.vm/tools/win10.xml
+++ b/packages/debloat.vm/tools/win10.xml
@@ -141,6 +141,8 @@
         <registry-item name="Disable Windows Update Automatic Download" path="HKLM:\SOFTWARE\Policies\Microsoft\Windows\WindowsUpdate\AU" value="AUOptions" type="DWord" data="2" />
         <registry-item name="Disable Windows Update Automatic Restart" path="HKLM:\SOFTWARE\Microsoft\Windows NT\CurrentVersion\Image File Execution Options\MusNotification.exe" value="Debugger" type="String" data="cmd.exe" />
         <registry-item name="Disable Microsoft Connectivity Test (msftconnecttest.com)" path="HKLM:\SYSTEM\CurrentControlSet\services\NlaSvc\Parameters\Internet" value="EnableActiveProbing" type="DWord" data="0" />
+        <registry-item name="Enable opening cmd from context menu - Default" path="Registry::HKEY_CLASSES_ROOT\Directory\Background\shell\ctx_cmd" value="(Default)" type="String" data="Cmd Here" />
+        <registry-item name="Enable opening cmd from context menu - Command" path="Registry::HKEY_CLASSES_ROOT\Directory\Background\shell\ctx_cmd\command" value="(Default)" type="String" data="cmd.exe /s /k pushd &quot;%V&quot;" />
     </registry-items>
     <path-items>
         <!--
--- a/packages/debloat.vm/debloat.vm.nuspec
+++ b/packages/debloat.vm/debloat.vm.nuspec
@@ -2,7 +2,7 @@
 <package xmlns="http://schemas.microsoft.com/packaging/2015/06/nuspec.xsd">
   <metadata>
     <id>debloat.vm</id>
-    <version>0.0.0.20231110</version>
+    <version>0.0.0.20231218</version>
     <description>Debloat and performance configurations for Windows OS</description>
     <authors>Mandiant</authors>
     <dependencies>


Ana06 (changes requested):
Thanks for all the research @d35ha!! 😄

packages/debloat.vm/tools/win10.xml
```diff
@@ -178,5 +178,15 @@
             <cmd value="powercfg /X standby-timeout-ac 0" />
             <cmd value="powercfg /X standby-timeout-dc 0" />
         </custom-item>
+        <custom-item name="Enable openning cmd from context menu">
+            <cmd value="New-Item -Path Registry::HKEY_CLASSES_ROOT\Directory\shell\flare-ctx-menu-cmd\command -Force -ErrorAction SilentlyContinue" />
+            <cmd value="New-Item -Path Registry::HKEY_CLASSES_ROOT\Directory\Background\shell\flare-ctx-menu-cmd\command -Force -ErrorAction SilentlyContinue" />
+            <cmd value="Set-ItemProperty -Path Registry::HKEY_CLASSES_ROOT\Directory\shell\flare-ctx-menu-cmd -Name '(Default)' -Value 'Open command window here' -Force -ErrorAction SilentlyContinue" />
```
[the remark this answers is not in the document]
d35ha: Done, please review.

d35ha:
@Ana06  
I noticed that for directories there's an already existing option for opening cmd console with the target directory as the current directory, should I only enable it for directories background (Right click on the directory background)?

cmd_here

Ana06:
@d35ha 
> I noticed that for directories there's an already existing option for opening cmd console with the target directory as the current directory, should I only enable it for directories background (Right click on the directory background)?

Yes, I think this is a good idea!

What about calling the one for directories background `Cmd here` (instead of `Open command window here`) to keep it consistent with the existent one?

d35ha:
@Ana06 
> Yes, I think this is a good idea!
> 
> What about calling the one for directories background `Cmd here` (instead of `Open command window here`) to keep it consistent with the existent one?

Done, also removed the icon for consistency.

packages/debloat.vm/tools/win10.xml
```diff
@@ -141,6 +141,8 @@
         <registry-item name="Disable Windows Update Automatic Download" path="HKLM:\SOFTWARE\Policies\Microsoft\Windows\WindowsUpdate\AU" value="AUOptions" type="DWord" data="2" />
         <registry-item name="Disable Windows Update Automatic Restart" path="HKLM:\SOFTWARE\Microsoft\Windows NT\CurrentVersion\Image File Execution Options\MusNotification.exe" value="Debugger" type="String" data="cmd.exe" />
         <registry-item name="Disable Microsoft Connectivity Test (msftconnecttest.com)" path="HKLM:\SYSTEM\CurrentControlSet\services\NlaSvc\Parameters\Internet" value="EnableActiveProbing" type="DWord" data="0" />
+        <registry-item name="Enable openning cmd from context menu - Default" path="Registry::HKEY_CLASSES_ROOT\Directory\Background\shell\ctx_cmd" value="(Default)" type="String" data="Cmd Here" />
```
Ana06, line 144: Typo
```suggestion
        <registry-item name="Enable opening cmd from context menu - Default" path="Registry::HKEY_CLASSES_ROOT\Directory\Background\shell\ctx_cmd" value="(Default)" type="String" data="Cmd Here" />
```

packages/debloat.vm/tools/win10.xml
```diff
@@ -141,6 +141,8 @@
...
+        <registry-item name="Enable openning cmd from context menu - Command" path="Registry::HKEY_CLASSES_ROOT\Directory\Background\shell\ctx_cmd\command" value="(Default)" type="String" data="cmd.exe /s /k pushd &quot;%V&quot;" />
```
Ana06, line 145: Typo
```suggestion
        <registry-item name="Enable opening cmd from context menu - Command" path="Registry::HKEY_CLASSES_ROOT\Directory\Background\shell\ctx_cmd\command" value="(Default)" type="String" data="cmd.exe /s /k pushd &quot;%V&quot;" />
```

Ana06 (commented):
@mandiant/commando-vm this PR enables opening cmd from context menu for you as well. I think it is a nice feature also for Commando VM, but mentioning you here in case you have something against it. 😄

packages/debloat.vm/tools/win10.xml
d35ha: Fixed.

packages/debloat.vm/tools/win10.xml
d35ha: Fixed.

Ana06:
@d35ha 
> I noticed that for directories there's an already existing option for opening cmd console with the target directory as the current directory, should I only enable it for directories background (Right click on the directory background)?

I wonder what is adding this... I don't have it there in my fresh Windows installation or after a minimal FLARE-VM installation with only few packages. But it is there after I install FLARE-VM with the default configuration. Is a package installing it? which one? 🤔 🕵️ @vm-packages does someone has any idea?

d35ha:
@Ana06 
It was tricky but got it, the responsible package is the `Malcode Analyst Pack` (map.vm).

Capture

Ana06 (approved):
@d35ha 
> It was tricky but got it, the responsible package is the Malcode Analyst Pack (map.vm).

😮 That's great research @d35ha!!!

Then I see two options:
- Leave this PR as it is, as I think it address what is requested in https://github.com/mandiant/VM-Packages/issues/786
- Add also the `Cmd here` option for directories as well using the same registry keys as nmap to ensure that we don't duplicate this entry if the package is installed. This only makes a difference if nmap is not installed.

As we are installing nmap in the default FLARE-VM configuration I tend to think we don't need to make extra changes and we can leave it just as it is. @d35ha what do you think?

d35ha:
@Ana06 
Honestly, I'm not sure how nmap is related to this problem.
Actually, the `Malcode Analyst Pack` doesn't use registry the same way we're using for registering the context menu options.
It's depending on using a set of COM interfaces upon the triggering of the context menu (that's why it was tricky to trace), it's overcomplicated for us.
I prefer to leave it as it is.

d35ha:
@Ana06 
Honestly, I'm not sure how nmap is related to this problem.
Actually, the `Malcode Analyst Pack` doesn't use registry the same way we're using for registering the context menu options.
It's depending on using a set of COM interfaces upon the triggering of the context menu (that's why it was tricky to trace), it's overcomplicated for us.
I prefer to leave it as it is.

Ana06:
ok, let's merge it then. Thanks for all the research @d35ha!!

Issue #794, opened 2023-12-13
VM-Remove-DesktopFiles deletes fakenet_logs


Ana06:
### What's the problem?

`VM-Remove-DesktopFiles` deletes `fakenet_logs` because it is a shortcut and we are excluding folders with that name.

### Steps to Reproduce

1. Install FLARE-VM including fakenet
2. Run `VM-Remove-DesktopFiles`or `VM-Clean-Up`

### Environment

Environment independent.

### Detected by test suite

No

Ana06:
Workaround until this bug is fixed: `VM-Clean-Up -excludeFiles "fakenet_logs.lnk"``

This also means we only have to move it to files to fix the bug.

PR #797, opened 2023-12-18
Fix fakenet logs

Ana06:
`VM-Remove-DesktopFiles` deletes `fakenet_logs` because it is a shortcut and we are excluding folders with that name. Exclude it as a file instead.

Do not exclude the `MICROSOFT Windows 10 License Terms.txt` file by default for more flexibility. We originally decided to add this file as having at least one file by default simplified the implementation and made the linter happy. Now that we have added `fakenet_logs.lnk`, this file can be removed.

Improve the documentation of `VM-Remove-DesktopFiles` to add real examples we are using.

Fixes https://github.com/mandiant/VM-Packages/issues/794

### DIFF
--- a/packages/common.vm/common.vm.nuspec
+++ b/packages/common.vm/common.vm.nuspec
@@ -2,7 +2,7 @@
 <package xmlns="http://schemas.microsoft.com/packaging/2015/06/nuspec.xsd">
   <metadata>
     <id>common.vm</id>
-    <version>0.0.0.20231201</version>
+    <version>0.0.0.20231218</version>
     <description>Common libraries for VM-packages</description>
     <authors>Mandiant</authors>
   </metadata>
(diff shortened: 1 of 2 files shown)


emtuls:
@Ana06 Other than maybe combining the 2 commits into a single commit, it looks good to me!

emtuls (approved):
We could possibly merge the 2 commits into a single commit. Other than that, it looks good!

Ana06:
@emtuls thanks for the review!

> We could possibly merge the 2 commits into a single commit.

I have left them separately because they are addressing two different things: adding the fakenet_logs shortcut and removing the license terms files. I did it on purpose, but they are small enough that it could be argued that they belong together.

It is a good practice to have small and atomic commits, as small as possible but with working code: https://github.blog/2022-06-30-write-better-commits-build-better-projects/ I have been asking in some PR to squash the changes because I thought it would be easier for the author to squash all together than splitting into small and atomic commits, what would be a bit better. Ideally the git history should help up us finding when and why a change was made and we should avoid having commits like `fix linter`. Sorry if I have been a bit picky in some PRs 😄

Issue #676, opened 2023-09-29
Unify shortcut format


Ana06:
`VM-Install-Shortcut` ensures shortcuts are created always in the same way. I personally find it useful to display the executed command for console applications as some times the help message does not include how to re-run the tool. 

`VM-Install-Single-Ps1` and `VM-Install-Raw-GitHub-Repo` use `Install-ChocolateyShortcut` (instead of `VM-Install-Shortcut`) to run with Powershell. `VM-Install-Single-Ps1` Powershell with cmd while `VM-Install-Raw-GitHub-Repo`runs Powershell directly.  I find it confusing not to know what they run as in any of them the executed command is not displayed (I can check the shortcut, but I think this is not obvious for all users and also not a nice way to do it). 

I suggest we add a Powershell option to `VM-Install-Shortcut` that renders what is being executed and uses Powershell directly.

Ana06:
Documenting discussion with @MalwareMechanic:
>- We agree that it's useful to display what commands get executed when a LNK is clicked
>- We agree that shortcut helpers should be unified to support PowerShell as the default prompt

Ana06:
This comment from https://github.com/mandiant/VM-Packages/pull/800#discussion_r1442065869 could be checked/addressed when addressing this issue:
> the cd not needed because we have it already in -WorkingDirectory $executableDir? If so, we should change this in VM-Install-Shortcut.

Ana06:
@emtuls this issue is related to other issues you have worked on recently. I had assigned to myself, but I won't have time in the next weeks. So if you have time, please take it (and remove me). 😄

PR #800, opened 2023-12-19
Adjust Apktool package to use openjdk

### LINKED ISSUE #722: Stop using the apktool community package
### Details

The apktool community package requires javaruntime ≥ 1.8.0 (last version of Java Oracle with a commercial license). Install apktool from scratch (without using the community package) and with openjdk.vm (introduced in https://github.com/mandiant/VM-Packages/pull/720) as dependency.

emtuls:
This fixes #722.


packages/apktool.vm/apktool.vm.nuspec
```diff
@@ -2,12 +2,12 @@
 <package xmlns="http://schemas.microsoft.com/packaging/2015/06/nuspec.xsd">
   <metadata>
     <id>apktool.vm</id>
-    <version>2.9.0.20231024</version>
+    <version>2.9.1.20241219</version>
```
Ana06, line 5: [nitpick] The date in the forth segment is not needed as you are increasing the version

Ana06 (changes requested):

packages/apktool.vm/tools/chocolateyinstall.ps1
```diff
@@ -4,15 +4,30 @@ Import-Module vm.common -Force -DisableNameChecking
 try {
   $toolName = 'apktool'
   $category = 'Java & Android'
-  $shimPath = 'bin\apktool.exe'
+  $rawToolPath = Join-Path ${Env:RAW_TOOLS_DIR} "$toolName"
+
+  # Download apktool.bat
+  $wrapperPath = Join-Path $rawToolPath "$toolName.bat"
+  $wrapperSource = 'https://raw.githubusercontent.com/iBotPeaches/Apktool/master/scripts/windows/apktool.bat'
```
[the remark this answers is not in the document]
emtuls: I can use that commit instead. 👍 

I just followed what was on the 'Install Guide' in their documentation for Windows and how to run it: https://apktool.org/docs/install/

I assumed that the `.bat` file is something that does proper setup for `apktool` and wanted to follow what they recommend for installation, but if you think we should maybe just call `apktool` directly using our own shortcut, I can look into that instead.

packages/apktool.vm/tools/chocolateyinstall.ps1
```diff
@@ -4,15 +4,30 @@ Import-Module vm.common -Force -DisableNameChecking
...
+  Get-ChocolateyWebFile -PackageName '$toolName wrapper script' -FileFullPath $wrapperPath -Url $wrapperSource -Checksum $wrapperChecksum -ChecksumType "sha256"
+
+  # Download apktool.jar
+  $toolPath = Join-Path $rawToolPath "$toolName.jar"
+  $toolSource = 'https://github.com/iBotPeaches/Apktool/releases/download/v2.9.1/apktool_2.9.1.jar'
+  $toolChecksum = "de7ce8aa109acb649e7f69cfe91030ffc20dbcc46edd8abbf6c2d1e36cfccd7b"
+  Get-ChocolateyWebFile -PackageName $toolName -FileFullPath $toolPath -Url $toolSource -Checksum $toolChecksum -ChecksumType "sha256"
+
+  # Add apktool to Path
+  $env:Path += $rawToolPath
```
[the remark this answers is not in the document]
emtuls: I agree that I think it should become a new function `VM-Add-To-Path` in `common.vm`. I initially looked for such a function, but didn't find it, so I went with this route. I can make a PR for that and change this PR to utilize that. 👍 

As for if this method of adding to the Path Environment Variable works, it does work. :)

packages/apktool.vm/apktool.vm.nuspec
emtuls: Gotcha! I can make that change back to stay consistent.

packages/apktool.vm/tools/chocolateyinstall.ps1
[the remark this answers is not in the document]
emtuls: Created the function here: https://github.com/mandiant/VM-Packages/pull/806

packages/apktool.vm/tools/chocolateyinstall.ps1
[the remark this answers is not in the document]
emtuls: Updated so that it uses the commit!

Just need to know how to go forward with whether or not we should use the `.bat` (as suggested by the tool author's install guide) or if we should just do everything in our own shortcut instead.

emtuls:
This PR now requires https://github.com/mandiant/VM-Packages/pull/806 to be merged before it will work.

packages/apktool.vm/tools/chocolateyinstall.ps1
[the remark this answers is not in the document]
Ana06: I was a bit confused as I think the metapackage that we used to use did it differently. But I think **we should do it as suggested by the too author's guide and test it** to ensure that it works.

Ana06 (changes requested):
https://github.com/mandiant/VM-Packages/pull/806 merged! We need to require common.vm >= `0.0.0.20231221` to ensure that the `VM-Add-To-Path` function is available. This is the nuspec syntax:
```xml
<dependency id="common.vm" version="0.0.0.20231221" />
```

packages/apktool.vm/tools/chocolateyinstall.ps1
```diff
@@ -4,15 +4,29 @@ Import-Module vm.common -Force -DisableNameChecking
...
   $shortcutDir = Join-Path ${Env:TOOL_LIST_DIR} $category
   $shortcut = Join-Path $shortcutDir "$toolName.lnk"
 
-  $executablePath = Join-Path ${Env:ChocolateyInstall} $shimPath -Resolve
   $executableCmd  = Join-Path ${Env:WinDir} "system32\cmd.exe"
   $executableDir  = Join-Path ${Env:UserProfile} "Desktop"
-  $executableArgs = "/K `"cd `"$executableDir`" && `"$executablePath`""
+  $executableArgs = "/K `"$toolName`""
 
   Install-ChocolateyShortcut -shortcutFilePath $shortcut -targetPath $executableCmd -Arguments $executableArgs -WorkingDirectory $executableDir
```
Ana06, line 31: [nitpick] what about using `VM-Install-Shortcut`? We should standarize how we create shortcuts (this is tracked in https://github.com/mandiant/VM-Packages/issues/676)

packages/apktool.vm/tools/chocolateyinstall.ps1
```diff
@@ -4,15 +4,29 @@ Import-Module vm.common -Force -DisableNameChecking
...
+  VM-Add-To-Path $rawToolPath
 
   $shortcutDir = Join-Path ${Env:TOOL_LIST_DIR} $category
   $shortcut = Join-Path $shortcutDir "$toolName.lnk"
 
-  $executablePath = Join-Path ${Env:ChocolateyInstall} $shimPath -Resolve
   $executableCmd  = Join-Path ${Env:WinDir} "system32\cmd.exe"
   $executableDir  = Join-Path ${Env:UserProfile} "Desktop"
-  $executableArgs = "/K `"cd `"$executableDir`" && `"$executablePath`""
+  $executableArgs = "/K `"$toolName`""
```
Ana06, line 29: [question] Is the `cd` not needed because we have it already in `-WorkingDirectory $executableDir`? If so, we should change this in `VM-Install-Shortcut`. 🤔

Ana06 (commented):
@emtuls can you please rebase that we can merge this PR? 👍

packages/apktool.vm/tools/chocolateyinstall.ps1
emtuls: I would need to test that theory, but I believe that is the case. Something to look into!

Ana06: I'll add this to https://github.com/mandiant/VM-Packages/issues/676 👍

packages/apktool.vm/tools/chocolateyinstall.ps1
```diff
@@ -4,18 +4,26 @@ Import-Module vm.common -Force -DisableNameChecking
...
+  # Download apktool.jar
+  $toolPath = Join-Path $rawToolPath "$toolName.jar"
+  $toolSource = 'https://github.com/iBotPeaches/Apktool/releases/download/v2.9.1/apktool_2.9.1.jar'
+  $toolChecksum = "de7ce8aa109acb649e7f69cfe91030ffc20dbcc46edd8abbf6c2d1e36cfccd7b"
+  Get-ChocolateyWebFile -PackageName $toolName -FileFullPath $toolPath -Url $toolSource -Checksum $toolChecksum -ChecksumType "sha256"
 
-  Install-ChocolateyShortcut -shortcutFilePath $shortcut -targetPath $executableCmd -Arguments $executableArgs -WorkingDirectory $executableDir
+  # Add apktool to Path
+  VM-Add-To-Path $rawToolPath
+
```
Ana06, line 23: Can we get rid of this extra empty line?

Ana06 (commented):
You need to increase the version (the weekly updates updated to `2.9.1`, so we need to update it to `2.9.1.20230104` after the rebase) 😉

packages/apktool.vm/apktool.vm.nuspec
```diff
@@ -6,8 +6,8 @@
     <authors>Connor Tumbleson, Ryszard Wisniewski</authors>
     <description>A tool for reverse engineering 3rd party, closed, binary Android apps.</description>
     <dependencies>
-      <dependency id="common.vm" />
-      <dependency id="apktool" version="[2.9.1]" />
+      <dependency id="common.vm" version="0.0.0.20240104" />
```
Ana06, line 9: This version doesn't exist:
```suggestion
      <dependency id="common.vm" version="0.0.0.20231221" />
```

packages/apktool.vm/tools/chocolateyinstall.ps1
```diff
@@ -4,17 +4,24 @@ Import-Module vm.common -Force -DisableNameChecking
 try {
   $toolName = 'apktool'
   $category = 'Java & Android'
-  $shimPath = 'bin\apktool.exe'
+  $rawToolPath = Join-Path ${Env:RAW_TOOLS_DIR} "$toolName"
```
Ana06, line 7: We should add `-Resolve` to all `Join-Path` to fail the package early if something goes wrong to help with debugging (sorry I had missed this before)

packages/apktool.vm/tools/chocolateyinstall.ps1
```diff
@@ -4,17 +4,24 @@ Import-Module vm.common -Force -DisableNameChecking
...
+  $toolSource = 'https://github.com/iBotPeaches/Apktool/releases/download/v2.9.1/apktool_2.9.1.jar'
+  $toolChecksum = "de7ce8aa109acb649e7f69cfe91030ffc20dbcc46edd8abbf6c2d1e36cfccd7b"
+  Get-ChocolateyWebFile -PackageName $toolName -FileFullPath $toolPath -Url $toolSource -Checksum $toolChecksum -ChecksumType "sha256"
 
-  Install-ChocolateyShortcut -shortcutFilePath $shortcut -targetPath $executableCmd -Arguments $executableArgs -WorkingDirectory $executableDir
+  # Add apktool to Path
+  VM-Add-To-Path $rawToolPath
+
+  VM-Install-Shortcut -toolname $toolname -category $category -executablePath $toolPath -consoleApp $true
   VM-Assert-Path $shortcut
```
Ana06, line 25: You need to remove this line as `$shortcut` does not exist anymore. This check is done inside `VM-Install-Shortcut`
```suggestion
```

Ana06 (changes requested):

packages/apktool.vm/tools/chocolateyinstall.ps1
emtuls: Ah yes, my brain got confused when I was looking at the `VM-Install-Shortcut` function and my function. Good catch, thank you!

packages/apktool.vm/tools/chocolateyinstall.ps1
```diff
@@ -4,18 +4,24 @@ Import-Module vm.common -Force -DisableNameChecking
 try {
   $toolName = 'apktool'
   $category = 'Java & Android'
-  $shimPath = 'bin\apktool.exe'
+  $rawToolPath = Join-Path ${Env:RAW_TOOLS_DIR} "$toolName" -Resolve
```
Ana06, line 7:
```suggestion
  $rawToolPath = Join-Path ${Env:RAW_TOOLS_DIR} "$toolName"
```

Sorry, we need to remove the `-Resolve` as they are before the path is created and add a `VM-Assert-Path` after the `Get-ChocolateyWebFile`

Ana06 (changes requested):
I was wrong with my last comment, sorry for the confusion

Ana06 (approved):
It seems to work well now! Thanks @emtuls

Issue #790, opened 2023-12-13
New package: xmlmacrodeobfuscator


Ana06:
### Details

XLMMacroDeobfuscator:
- https://github.com/DissectMalware/XLMMacroDeobfuscator

Ana06:
@KJaeYoung is familiar with this tool and is able to answer any possible question about how to install/use this tool. 😃

PR #801, opened 2023-12-20
Add the XLMMacroDeobfuscator library package

d35ha:
Closes https://github.com/mandiant/VM-Packages/issues/790


packages/python3.xlmmacrodeobfuscator.vm/tools/chocolateyinstall.ps1
```diff
@@ -0,0 +1,19 @@
...
+Import-Module vm.common -Force -DisableNameChecking
+
+try {
+  $toolName = "XLMMacroDeobfuscator"
+  $category = "Office"
+
+  $whlUrl = "https://github.com/DissectMalware/$toolName/releases/download/v0.2.7/$toolName-0.2.7-py3-none-any.whl"
+
+  # Install the wheel file using python3 pip
+  VM-Execute-Custom-Command -name "Install with python3 pip" -cmds "pip install $whlUrl"
```
Ana06, line 11: Is there a reason to use `VM-Execute-Custom-Command` instead of `Invoke-Expression`? `VM-Execute-Custom-Command` was introduced to be used in `VM-Apply-Configurations` used to load custom config files. We don't use it in any package at the moment.

I think we should specify the concrete python version as we do in [libraries.python3.vm](https://github.com/mandiant/VM-Packages/blob/main/packages/libraries.python3.vm/tools/chocolateyinstall.ps1) to ensure there are not issues with other possible installed versions in the system.

```suggestion
Invoke-Expression "py -3.10 -m pip install $whlUrl"
```

packages/python3.xlmmacrodeobfuscator.vm/tools/chocolateyinstall.ps1
```diff
@@ -0,0 +1,19 @@
...
+
+  $executablePath = (Get-Command xlmdeobfuscator -errorAction SilentlyContinue).Source
+  VM-Assert-Path $executablePath
+
+  VM-Install-Shortcut -toolName $toolName -category $category -executablePath $executablePath -consoleApp $true
```
Ana06, line 16: Is there a reason to use `VM-Execute-Custom-Command` instead of `Invoke-Expression`? `VM-Execute-Custom-Command` was introduced to be used in `VM-Apply-Configurations` used to load custom config files. We don't use it in any package at the moment.

I think we should specify the concrete python version as we do in [libraries.python3.vm](https://github.com/mandiant/VM-Packages/blob/main/packages/libraries.python3.vm/tools/chocolateyinstall.ps1) to ensure there are not issues with other possible installed versions in the system.

```suggestion
Invoke-Expression "py -3.10 -m pip install $whlUrl"
```

Same in the uninstaller.

packages/python3.xlmmacrodeobfuscator.vm/tools/chocolateyuninstall.ps1
```diff
@@ -0,0 +1,8 @@
+$ErrorActionPreference = 'Continue'
+Import-Module vm.common -Force -DisableNameChecking
+
+$toolName = "XLMMacroDeobfuscator"
+$category = 'Office'
+
+VM-Execute-Custom-Command -name "Uninstall with python3 pip" -cmds "pip uninstall $toolName -y"
+VM-Uninstall $toolName $category
```
Ana06, line 8: `VM-Uninstall` is undoing more things that what you are doing in the installer (it calls `Uninstall-BinFile`). I would copy only the lines in `VM-Uninstall` that apply here.

Ana06 (changes requested):
I think we need to add the tool to the Path, or is it added already? We may want to add a new `VM-Add-To-Path` function to common to add a file/directory path to the Path with the two lines code I used in https://github.com/mandiant/VM-Packages/pull/793/files. We should use this function in didier-stevens-suite.vm and in this package (and potentially in other packages).

packages/python3.xlmmacrodeobfuscator.vm/tools/chocolateyinstall.ps1
d35ha: `VM-Execute-Custom-Command` executes pip within a new process but directly executing it throws the exception `System.Management.Automation.RemoteException`

image

packages/python3.xlmmacrodeobfuscator.vm/tools/chocolateyuninstall.ps1
d35ha: `VM-Remove-Tool-Shortcut` is more fitting.

d35ha:
@Ana06 
The tool executable is already added to the path (`C:\Python310\Scripts`) with the name "xlmdeobfuscator.exe".

packages/python3.xlmmacrodeobfuscator.vm/tools/chocolateyinstall.ps1
```diff
@@ -0,0 +1,19 @@
+$ErrorActionPreference = 'Stop'
+Import-Module vm.common -Force -DisableNameChecking
+
+try {
+  $toolName = "XLMMacroDeobfuscator"
+  $category = "Office"
+
+  $whlUrl = "https://github.com/DissectMalware/$toolName/releases/download/v0.2.7/$toolName-0.2.7-py3-none-any.whl"
```
Ana06, line 8: is there a reason we can't use the version of pypi: https://pypi.org/project/XLMMacroDeobfuscator/ (just `py -3.10 -m pip install XLMMacroDeobfuscator`)?

Ana06: @KJaeYoung do we need a shortcut in the Tools folder for XLMMacroDeobfuscator?

Ana06: If we don't need the shortcut and we can use the version of pypi, I think we should add this tool to python3.libraries.vm config instead of creating a new package. 🤔

packages/python3.xlmmacrodeobfuscator.vm/tools/chocolateyinstall.ps1
Ana06: This is strange as our python3.libraries.vm package installs correctly. Does this happen if we use the pypi version with `py -3.10 -m pip install XLMMacroDeobfuscator`?

Ana06 (changes requested):

packages/python3.xlmmacrodeobfuscator.vm/tools/chocolateyinstall.ps1
d35ha: The problem persists, see the attached logs file.
[chocolatey.log](https://github.com/mandiant/VM-Packages/files/13729101/chocolatey.log)

packages/python3.xlmmacrodeobfuscator.vm/tools/chocolateyinstall.ps1
KJaeYoung: > @KJaeYoung do we need a shortcut in the Tools folder for XLMMacroDeobfuscator?

Nope, no need for a shortcut in the Tools folder. Just needs to be accessible via the command line.

d35ha: No specific reason, both techniques are the same.

Ana06: pypi handles the version updates for us without having to deal with hashes, so I think we should use the Pypi version as we are doing for other python libraries.

Ana06: Thanks for the answer @KJaeYoung.

Two options then:
1. Leave this as a separate package with the shortcut
2. Add this package to the [python3.libraries.vm modules file](https://github.com/mandiant/VM-Packages/blob/main/packages/libraries.python3.vm/tools/modules.xml) instead of creating a new package

option 1 gives us more flexibility and the shortcut, while option 2 simplifies the code and keeps it consistent with the current python libraries. I don't really have a strong opinion here. @mandiant/flare-vm @d35ha any preferences?

d35ha: @Ana06 
I prefer the second option, more consistent and minimal changes.

d35ha: @Ana06 
Added to the python libraries list.

Ana06 (approved):

Issue #756, opened 2023-11-24
7zip improvements


Ana06:
### Details

I suggest:
- At the moment `7zip` is on path, I suggest changing it by `7z`.
- Remove 7zip from the Utilities folder, I don't think is needed to add a shortcut in the Tools directory.

PR #803, opened 2023-12-20
Improve 7zip-15-05.vm

Ana06:
At the moment `7zip` is on path, change it by `7z`.

@vm-packages do we agree on removing 7zip from the Utilities folder? I tend to think we don't need a shortcut for this tool as it is not really an analysis tool and it is already added to the right menu. But it can be used for things like extracting NSIS scripts. So it can be considered an analysis tool as well.

Closes https://github.com/mandiant/VM-Packages/issues/756

### DIFF
--- a/packages/7zip-15-05.vm/7zip-15-05.vm.nuspec
+++ b/packages/7zip-15-05.vm/7zip-15-05.vm.nuspec
@@ -2,7 +2,7 @@
 <package xmlns="http://schemas.microsoft.com/packaging/2015/06/nuspec.xsd">
   <metadata>
     <id>7zip-15-05.vm</id>
-    <version>15.05.0.20230926</version>
+    <version>15.05.0.20231220</version>
     <authors>Igor Pavlov</authors>
     <description>7-Zip file archiver. This version is able to extract NSIS scripts.</description>
     <dependencies>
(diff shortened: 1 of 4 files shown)


emtuls (commented):
Yup! I agree that it shouldn't be in Utilities. Adds a bit of extra clutter for something that isn't used enough in that sense, but there is -some- use.

As an alternative, we could possibly put them inside of the `Raw Tools Dir` if we think there's a potential that having access to the binary shortcuts might be useful?

Ana06:
@emtuls 
> As an alternative, we could possibly put them inside of the Raw Tools Dir if we think there's a potential that having access to the binary shortcuts might be useful?

How do you purpose this shortcut to be used? I can't think of a reason to want to keep the shortcut in any location apart from finding it in the Tools directory (if we consider it to be an analysis tool, what I argue is not), but maybe I am missing something. 🤔

emtuls:
> @emtuls
> 
> > As an alternative, we could possibly put them inside of the Raw Tools Dir if we think there's a potential that having access to the binary shortcuts might be useful?
> 
> How do you purpose this shortcut to be used? I can't think of a reason to want to keep the shortcut in any location apart from finding it in the Tools directory (if we consider it to be an analysis tool, what I argue is not), but maybe I am missing something. 🤔

I don't think I can think of any reason I would, as I prefer the right click option, but I was airing on the side of caution that there may be users that prefer the method of dragging a file onto the shortcut to open it in 7z, or they might not have a right click (a bit rare though, especially with modern hot keys built into Windows).

Outside of that. I'm all for making the change to remove it if those use cases aren't ones we should consider. :)

day1player:
Agreed we dont need a shortcut for it

Issue #791, opened 2023-12-13
New package: offvis.vm


Ana06:
### Details

Microsoft OffVis

Ana06:
@KJaeYoung is familiar with this tool and is able to answer any possible question about how to install/use this tool. 😃

d35ha:
This tool requires .NET Framework 3.5 and trying to install the [choco version](https://community.chocolatey.org/packages/DotNet3.5) fails.
@Ana06 any ideas?

d35ha:
@Ana06 
It's working now (idk why it didn't work at first), please test it extensively.

PR #805, opened 2023-12-20
Add the OffVis tool package

d35ha:
Closes https://github.com/mandiant/VM-Packages/issues/791

### DIFF
--- a/packages/offvis.vm/offvis.vm.nuspec
+++ b/packages/offvis.vm/offvis.vm.nuspec
@@ -1,0 +1,13 @@
+<?xml version="1.0" encoding="utf-8"?>
+<package xmlns="http://schemas.microsoft.com/packaging/2015/06/nuspec.xsd">
+  <metadata>
+    <id>offvis.vm</id>
+    <version>1.0</version>
+    <authors>Microsoft</authors>
+    <description>The Microsoft Office Visualization Tool (OffVis) is a tool from Microsoft that helps understanding the Microsoft Office binary file format in order to deconstruct .doc-, .xls- and .ppt-based targeted attacks.</description>
+    <dependencies>
+      <dependency id="common.vm" />
+      <dependency id="dotnet3.5" />
+    </dependencies>
+  </metadata>
+</package>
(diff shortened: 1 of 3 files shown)


packages/offvis.vm/offvis.vm.nuspec
```diff
@@ -0,0 +1,13 @@
+<?xml version="1.0" encoding="utf-8"?>
+<package xmlns="http://schemas.microsoft.com/packaging/2015/06/nuspec.xsd">
+  <metadata>
+    <id>offvis.vm</id>
+    <version>1.0</version>
+    <authors>Microsoft</authors>
+    <description>The Microsoft Office Visualization Tool (OffVis) is a tool from Microsoft that helps understanding the Microsoft Office binary file format in order to deconstruct .doc-, .xls- and .ppt-based targeted attacks.</description>
+    <dependencies>
+      <dependency id="common.vm" />
+      <dependency id="dotnet3.5" />
```
Ana06, line 10: do we need this concrete version of dotnet? We should try to use the same version in all tools if possible. We have introduced metapackages for some dependencies to avoid problems: https://github.com/mandiant/VM-Packages/issues/673 So thinking if we should do something like this for dotnet (assuming we don't need a concrete version or that at least we can share the major vesion with other packages)

d35ha: I'm afraid that the concrete version is needed.

image

Ana06: I am now remembering, we are also installing this concrete version in FLARE-VM https://github.com/mandiant/flare-vm/blob/main/config.xml#L28 because otherwise some all .NET binaries don't run

d35ha: Then, I think it's redundant here, removed.

d35ha: Rollbacked.

Ana06 (approved):
Thanks @d35ha! 🎉

PR #806, opened 2023-12-20
Add 'VM-Add-To-Path' function to common.vm

emtuls:
This creates a method that will allow users to add something to the Machine Path Environment Variable, which is useful for many tools.

PR #809, opened 2023-12-21
Use VM-Add-To-Path function

emtuls:
This PR updates the package to use the new function `VM-Add-To-Path` which adds a path to the Machine Environment Variable Path and was added in `common.vm` https://github.com/mandiant/VM-Packages/pull/806 (which will need to be merged before checks in this PR will be successful).


Ana06 (changes requested):
We need to require common.vm >= `0.0.0.20231221` to ensure that the `VM-Add-To-Path` function is available. This is the nuspec syntax:
 <dependency id="common.vm" version="0.0.0.20231221" />

Ana06 (approved):

PR #812, opened 2023-12-24
Add dotnet-6 metapackage

naacbin:
I have created this metapackage to address #778 request.
I tried to add the `dotnet-6.0-sdk` for sfextract. However, when I install this package it doesn't really install the SDK on my system I had to use `dotnet-6.0-sdk-4xx`. Even, with this package in place when trying to install sfextract I have the following error message :
```text
C:\Users\User\AppData\Local\ChocoCache\6eac32f8-677f-4cec-9471-67015f93f927\restore.csproj : error NU1100: Unable to resolve 'sfextract (>= 2.1.0)' for 'net6.0'.
C:\Users\User\AppData\Local\ChocoCache\6eac32f8-677f-4cec-9471-67015f93f927\restore.csproj : error NU1100: Unable to resolve 'sfextract (>= 2.1.0)' for 'net6.0/any'.
```

### DIFF
--- a/packages/dotnet-6.vm/dotnet-6.vm.nuspec
+++ b/packages/dotnet-6.vm/dotnet-6.vm.nuspec
@@ -1,0 +1,14 @@
+<?xml version="1.0" encoding="utf-8"?>
+<package xmlns="http://schemas.microsoft.com/packaging/2015/06/nuspec.xsd">
+  <metadata>
+    <id>dotnet-6.vm</id>
+    <version>0.0.0.20231224</version>
+    <authors>Microsoft</authors>
+    <description>Metapackage for .NET6 to ensure all packages use the same version.</description>
+    <dependencies>
+      <dependency id="common.vm" />
+      <dependency id="dotnet-6.0-desktopruntime" version="(6.0.14, 6.0.26]" />
+      <dependency id="dotnet-6.0-runtime" version="(6.0.14, 6.0.26]" />
+    </dependencies>
+  </metadata>
+</package>


Ana06:
@naacbin
 > I tried to add the dotnet-6.0-sdk for sfextract. However, when I install this package it doesn't really install the SDK on my system I had to use dotnet-6.0-sdk-4xx. Even, with this package in place when trying to install sfextract I have the following error message

This is strange, because it is installing correctly in the [daily run](https://github.com/mandiant/VM-Packages/wiki/Daily-Failures) Have you tried to install it in an empty system? It could be a dependency incompatibility issue, which is exactly what we want to avoid with the metapackages (as all packages request the same version).

Ana06 (commented):
Thanks for the PR @naacbin!

We should change the package/s that use dotnet directly to use the the metapackage in this PR requiring the new common version.

packages/dotnet-6.vm/dotnet-6.vm.nuspec
```diff
@@ -0,0 +1,14 @@
...
+<package xmlns="http://schemas.microsoft.com/packaging/2015/06/nuspec.xsd">
+  <metadata>
+    <id>dotnet-6.vm</id>
+    <version>0.0.0.20231224</version>
+    <authors>Microsoft</authors>
+    <description>Metapackage for .NET6 to ensure all packages use the same version.</description>
+    <dependencies>
+      <dependency id="common.vm" />
+      <dependency id="dotnet-6.0-desktopruntime" version="[6.0.25]" />
+      <dependency id="dotnet-6.0-runtime" version="[6.0.25]" />
```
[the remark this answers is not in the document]
naacbin: I don't think so, I have tried with only `dotnet-6.0-desktopruntime` and CLI application can't run.

naacbin:
> @naacbin
> 
> > I tried to add the dotnet-6.0-sdk for sfextract. However, when I install this package it doesn't really install the SDK on my system I had to use dotnet-6.0-sdk-4xx. Even, with this package in place when trying to install sfextract I have the following error message
> 
> This is strange, because it is installing correctly in the [daily run](https://github.com/mandiant/VM-Packages/wiki/Daily-Failures) Have you tried to install it in an empty system? It could be a dependency incompatibility issue, which is exactly what we want to avoid with the metapackages (as all packages request the same version).
> 
> Does sfextract work if we only install the desktopruntime (instead of the sdk)?

No, the install doesn't work with only "desktopruntime" too. I will test on an empty system in the next week (if I find time to do it).

naacbin:
I don't know why it didn't worked on my previous machine but now I can install `sfextract`. However, I don't think SDK dependency should be added to this package as it requires a lot of space (800Mb) and is only used by `sfextract`. This could be reconsidered if other packages in the future use the SDK.

If it good on your side, we could merge this one and I will update #778

packages/dotnet-6.vm/dotnet-6.vm.nuspec
```diff
@@ -0,0 +1,14 @@
+<?xml version="1.0" encoding="utf-8"?>
+<package xmlns="http://schemas.microsoft.com/packaging/2015/06/nuspec.xsd">
+  <metadata>
+    <id>dotnet-6.vm</id>
+    <version>0.0.0.20231224</version>
+    <authors>Microsoft</authors>
+    <description>Metapackage for .NET6 to ensure all packages use the same version.</description>
+    <dependencies>
+      <dependency id="common.vm" />
+      <dependency id="dotnet-6.0-desktopruntime" version="[6.0.25]" />
```
Ana06, line 10: I suggest we use the following syntax to avoid that the automation updates the dependency, as that could break the tools using this metapackage:

```suggestion
      <dependency id="dotnet-6.0-desktopruntime" version="(6.0.13, 6.0.25]" />
```

Ana06: We could also restrict it more with `(6.0.24, 6.0.25]` but I think allowing a bigger range of minor versions may prevent installing several versions.

Ana06 (changes requested):
@naacbin can you add using this dependency in arsenalimagemounter.vm to this PR? If sfextract works with this dependency instead of the actual sdk please update it too.

naacbin:
> @naacbin can you add using this dependency in arsenalimagemounter.vm to this PR? If sfextract works with this dependency instead of the actual sdk please update it too.

- For `arsenalimagemounter.vm`, I used a metapackage that already [require this dependency](https://community.chocolatey.org/packages/arsenalimagemounter#dependencies). Should I remove it from the package or replace it with the one of this PR ?
- `sfextract` doesn't work with this package, only the SDK.

packages/dotnet-6.vm/dotnet-6.vm.nuspec
naacbin: Done

Ana06:
> For arsenalimagemounter.vm, I used a metapackage that already [require this dependency](https://community.chocolatey.org/packages/arsenalimagemounter#dependencies). Should I remove it from the package or replace it with the one of this PR ?

I propose to add both the dependency for the metapackage and this dependency (as they are compatible). Otherwise, arsenalimagemounter could install a newer version of .NET and we will end with two version of .NET package, which may cause conflicts.

We have had problems in the test suite when different packages installed different versions of the same dependency. We are trying to figure out the best way to avoid this to happen, so @naacbin if you have ideas how to improve this, please let us know. 😉

Ana06 (approved):

packages/dotnet-6.vm/dotnet-6.vm.nuspec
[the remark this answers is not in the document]
Ana06: @naacbin which CLI applications does not run without `dotnet-6.0-runtime`?

naacbin:
> > For arsenalimagemounter.vm, I used a metapackage that already [require this dependency](https://community.chocolatey.org/packages/arsenalimagemounter#dependencies). Should I remove it from the package or replace it with the one of this PR ?
> 
> I propose to add both the dependency for the metapackage and this dependency (as they are compatible). Otherwise, arsenalimagemounter could install a newer version of .NET and we will end with two version of .NET package, which may cause conflicts.
> 
> We have had problems in the test suite when different packages installed different versions of the same dependency. We are trying to figure out the best way to avoid this to happen, so @naacbin if you have ideas how to improve this, please let us know. 😉

I remove the commit to change `arsenalimagemounter` as it will be handle in a different PR to support dotnet-8 #852.
This PR can be merged.

Ana06 (approved):

PR #814, opened 2024-01-02
update sha-256 value

waletdevsec:


Ana06 (changes requested):
Thanks for yyour contribution @waletdevsec! 😀 

When modifying a package, you need to increase the version in the nuspec (`packages/metasploit.vm/metasploit.vm.nuspec` in this case) so that a new version is pushed to myget when merging the PR.  See the _When do I update the a package's version string?_  section in [out Wiki](https://github.com/mandiant/VM-Packages/wiki/Package-Structure#non-metapackages)

Can you please increase the version as indicated in the documentation and approve the Google CLA so that we can accept the change?

Ana06:
Closing due to inactivity, @waletdevsec please reopen if you are able to address the requested changes.

Proposing also a solution to fix this package and avoid that it fails again in https://github.com/mandiant/VM-Packages/issues/828

PR #817, opened 2024-01-05
Improve Clean Up

Ana06:
- Do not try to delete PS_Transcripts from PowerShell as we can easily delete it from cmd after closing all PowerShell consoles with the command:
  ```
  rmdir /s /q "%UserProfile%\Desktop\PS_Transcripts"
  ```
  The previous solution was complicated and didn't work properly.
- `Clear-RecycleBin` fails sometimes due to the following bug: https://github.com/PowerShell/PowerShell/issues/6743 This bug has been fixed in PowerShell >= 7, but we support PowerShell >= 5.
- Simplify error logging in `VM-Remove-DesktopFiles`.


emtuls (approved):
Looks good! Thanks for fixing up those issues that went unnoticed! :)

PR #818, opened 2024-01-10
[taskbar shortcut] Open cmd in Desktop

Ana06:
Working directory only works if cmd is not launched as Admin. As we need to launch is as admin, change to the Desktop location explicitly.


emtuls (approved):
Looks good! Good catch. 👍

PR #827, opened 2024-01-12
Fix sysinternals (and avoid that it breaks again!)

Ana06:
Re-write sysinternal.vm using `VM-Assert-Signature` to use the tools signature instead of the metapackage. The metapackage uses the downloaded zip hash, that breaks with every update, breaking our package.

@vm-packages I would appreciate some local testing. 😉 

Closes https://github.com/mandiant/VM-Packages/issues/469


binjo (approved):
LGTM

Ana06:
Thanks for the review @binjo!

mr-tz:
very cool!

PR #832, opened 2024-01-12
Fix broken rubeus by using a URL whose hash doesn't change

Ana06:
Use a GitHub download URL which includes the commit hash instead of using the `master` branch to avoid that the hash changes breaking the package.


day1player (approved):

day1player:
Not sure how that one made it in there 😂

PR #844, opened 2024-01-17
Create Pinnable CyberChef that uses Chrome

emtuls:
This attempts to fix https://github.com/mandiant/flare-vm/issues/559 and https://github.com/mandiant/VM-Packages/issues/296

It creates a shortcut that is pinnable, but to complete it, we will need to also merge https://github.com/mandiant/flare-vm/pull/562 after this is merged.


Ana06 (approved):
Tested locally, it works great!!! 🚀 Thanks a lot @emtuls!! 😻

Issue #842, opened 2024-01-17
Decouple "executablePath" from "IconLocation" inside "VM-Install-Shortcut"


emtuls:
### What's the problem?

Right now, we are forced to use the icon of whatever is set as the "executablePath" as that is what is passed in for the argument `-iconLocation`.

What if we would like to use a different icon located somewhere else or one that is not the executable we are using for the shortcut we are creating?

This came up when exploring ways to create a shortcut for `cyberchef` as I noticed that I am forced into providing a path to `google chrome` as an `executablePath` argument in order for the function to use it's icon. 

I propose having a separate argument for `IconLocation` that we can pass into `VM-Install-Shortcut` ourselves which would alleviate this issue.

### Steps to Reproduce

Try to create a console app shortcut using `VM-Install-Shortcut` with an executable used that is different than the icon you would like the shortcut to have.

Ana06:
I think this is a good idea!

emtuls:
Will make this modification after https://github.com/mandiant/VM-Packages/pull/840 is merged to avoid merge conflicts. 🙂

PR #845, opened 2024-01-19
Decouple executablePath from iconLocation

emtuls:
This closes https://github.com/mandiant/VM-Packages/issues/842


Ana06 (approved):
LGTM. Do we have a case where we want to use this new feature?

emtuls:
Hmmm. I would have used it for `cyberchef.vm` but with the modification I had to make, it doesn't fit the criteria for this anymore. 

It looks like `idafree.vm` could be a suitable candidate though!

Issue #675, opened 2023-09-29
Fix shortcut icon for .py and .bat files


Ana06:
`VM-Install-From-Zip` is being used for `.py` and `.bat` files, but the icon in the shortcut doesn't seem to work well in these cases. 

Related https://github.com/mandiant/VM-Packages/issues/674

Ana06:
@emtuls this issue is related to other issues you have worked on recently. Would you like to work on it? 😄

emtuls:
@Ana06 Sure thing! Do you have any examples of which ones do this and what should their icon's be?

A quick look for `Python` ones show: `ida.diaphora.vm` and `didier-stevens-beta/suite.vm` and I noticed this about these packages:
- `ida.diaphora.vm` is not intended to be run as a python script directly, so I'm not sure if we want to modify this?
- `didier-stevens-beta/suite.vm` installs a lot of tools, but only each shortcut is broken since it does not have python in its `Target`.

For the python ones, I defaulted to using the icon from `python.exe`, and I will also write a fix for the `didier-stevens-beta/suite.vm` packages so that the shortcuts work out of the box.

PR #854, opened 2024-01-23
Fix shortcuts in didier-stevens-beta.vm

emtuls:
Somewhat related to https://github.com/mandiant/VM-Packages/issues/675

This fixes the shortcuts created for the `didier-stevens-beta.vm` package so that each shortcut properly executes with python.

### DIFF
--- a/packages/didier-stevens-beta.vm/tools/chocolateyinstall.ps1
+++ b/packages/didier-stevens-beta.vm/tools/chocolateyinstall.ps1
@@ -17,15 +17,16 @@ try {
     $toolDir = Get-Item "${Env:RAW_TOOLS_DIR}\Beta-*"
     VM-Assert-Path $toolDir
 
-    # Add shortcut for commonly used office tools
+    # Add shortcut for commonly used office python tools
     ForEach ($toolName in @('onedump')) {
-      $executablePath = Join-Path $toolDir "$toolName.py"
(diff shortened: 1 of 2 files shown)


packages/didier-stevens-beta.vm/tools/chocolateyinstall.ps1
```diff
@@ -17,15 +17,16 @@ try {
     $toolDir = Get-Item "${Env:RAW_TOOLS_DIR}\Beta-*"
     VM-Assert-Path $toolDir
 
-    # Add shortcut for commonly used office tools
+    # Add shortcut for commonly used office python tools
     ForEach ($toolName in @('onedump')) {
-      $executablePath = Join-Path $toolDir "$toolName.py"
-      VM-Install-Shortcut $toolName $category $executablePath -consoleApp $true -arguments "--help"
+      $executablePath = (Get-Command python).Source
```
Ana06, line 22: The shortcut is not broken at the moment, executing the `.py` file directory works well:
Screenshot 2024-02-19 at 08 17 19

I would prefer to change only the icon (to be the python one) instead of changing the way we execute the tool.

Ana06 (approved):
If there are no objections in https://github.com/mandiant/VM-Packages/pull/855#discussion_r1494105091, then we can merge this one as well (the packages are identical)

PR #855, opened 2024-01-23
Fix shortcuts in didier-stevens-suite.vm

emtuls:
Somewhat related to https://github.com/mandiant/VM-Packages/issues/675

This fixes the shortcuts created for the `didier-stevens-suite.vm` package so that each shortcut properly executes with python.

### DIFF
--- a/packages/didier-stevens-suite.vm/tools/chocolateyinstall.ps1
+++ b/packages/didier-stevens-suite.vm/tools/chocolateyinstall.ps1
@@ -17,10 +17,12 @@ try {
     $toolDir = Get-Item "${Env:RAW_TOOLS_DIR}\DidierStevensSuite-*"
     VM-Assert-Path $toolDir
 
-    # Add shortcut for commonly used PDF tools
+    # Add shortcut for commonly used python PDF tools
     ForEach ($toolName in @('pdfid', 'pdf-parser')) {
-      $executablePath = Join-Path $toolDir "$toolName.py"
-      VM-Install-Shortcut $toolName $category $executablePath -consoleApp $true -arguments "--help"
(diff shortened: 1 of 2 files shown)


packages/didier-stevens-suite.vm/tools/chocolateyinstall.ps1
```diff
@@ -17,10 +17,12 @@ try {
     $toolDir = Get-Item "${Env:RAW_TOOLS_DIR}\DidierStevensSuite-*"
     VM-Assert-Path $toolDir
 
-    # Add shortcut for commonly used PDF tools
+    # Add shortcut for commonly used python PDF tools
     ForEach ($toolName in @('pdfid', 'pdf-parser')) {
-      $executablePath = Join-Path $toolDir "$toolName.py"
-      VM-Install-Shortcut $toolName $category $executablePath -consoleApp $true -arguments "--help"
+      $executablePath = (Get-Command python).Source
```
Ana06, line 22: could we use python only as icon (now that we have this argument) instead of changing how we are executing the tool? 🤔

emtuls: The reason I did this is because it makes the shortcut more useful. 🙂 Without the change, I get asked how I want to open the file (though this is likely because I have VSCode also installed, which seems to make things difficult).

Without the change but with the icon being Python (your suggestion, but with VSCode installed):
image


With the change I made (even with VSCode installed):
image

Ana06: That is a good reason to specify the application to run it with. @mandiant/flare-vm any objections?

Ana06 (approved):

PR #858, opened 2024-01-23
Add better shortcut icons for .py and .bat

emtuls:
This fixes https://github.com/mandiant/VM-Packages/issues/675

I tested `.bat` with: https://github.com/mandiant/VM-Packages/blob/79c35a70707ea79f01559eceb76eac000e029bd4/packages/dex2jar.vm

And `.py` with: https://github.com/mandiant/VM-Packages/tree/main/packages/didier-stevens-suite.vm

### DIFF
--- a/packages/common.vm/tools/vm.common/vm.common.psm1
+++ b/packages/common.vm/tools/vm.common/vm.common.psm1
@@ -282,6 +282,7 @@ function VM-Install-Shortcut{
     $shortcutDir = Join-Path ${Env:TOOL_LIST_DIR} $category
     $shortcut = Join-Path $shortcutDir "$toolName.lnk"
 
+    # Set the default icon to be the executable's icon
     if (-Not $iconLocation) {$iconLocation = $executablePath}
 
     if ($consoleApp) {
@@ -320,6 +321,19 @@ function VM-Install-Shortcut{
         Install-ChocolateyShortcut @shortcutArgs
     }
     VM-Assert-Path $shortcut
+
+    # If the targets is a .bat file, change the shortcut icon to Windows default
+    $extension = [System.IO.Path]::GetExtension($executablePath)
+    if ($extension -eq ".bat") {
+        $Shell = New-Object -ComObject ("WScript.Shell")
+        $Shortcut = $Shell.CreateShortcut($shortcut)
+
+        $IconArrayIndex = -68 # This is the specific icon that Windows uses for .bat files by default
+        $IconLocation = "C:\WINDOWS\system32\imageres.dll"
+        $Shortcut.IconLocation = "$IconLocation,$IconArrayIndex"
+
+        $Shortcut.Save()
+    }
 }
 
 # This functions returns $toolDir (outputed by Install-ChocolateyZipPackage) and $executablePath
--- a/packages/common.vm/common.vm.nuspec
+++ b/packages/common.vm/common.vm.nuspec
@@ -2,7 +2,7 @@
 <package xmlns="http://schemas.microsoft.com/packaging/2015/06/nuspec.xsd">
   <metadata>
     <id>common.vm</id>
-    <version>0.0.0.20240220</version>
+    <version>0.0.0.20240304</version>
     <description>Common libraries for VM-packages</description>
     <authors>Mandiant</authors>
   </metadata>


packages/common.vm/tools/vm.common/vm.common.psm1
```diff
@@ -281,7 +281,17 @@ function VM-Install-Shortcut{
     )
     $shortcutDir = Join-Path ${Env:TOOL_LIST_DIR} $category
     $shortcut = Join-Path $shortcutDir "$toolName.lnk"
+    $extension = [System.IO.Path]::GetExtension($executablePath)
 
+    # If the shortcut is for a .py, we should set the shortcut icon to be Python (requires Python to be installed)
```
Ana06, line 286: does this fail if Python is not installed?

packages/common.vm/tools/vm.common/vm.common.psm1
```diff
@@ -281,7 +281,17 @@ function VM-Install-Shortcut{
...
+    if ($extension -eq ".py") {
+        $pythonPath = (Get-Command python).Source
```
Ana06, line 288: according to your comment in https://github.com/mandiant/VM-Packages/pull/855#discussion_r1463383496, we should execute .py files with Python instead of changing the icon, do we then need this code?

I think the only case where that could make sense is diaphora, but it is a very special case, so I would rather have that in the package that here if it is the only reason.

packages/common.vm/tools/vm.common/vm.common.psm1
```diff
@@ -320,6 +330,18 @@ function VM-Install-Shortcut{
         Install-ChocolateyShortcut @shortcutArgs
     }
     VM-Assert-Path $shortcut
+
+    # If the file targets a .bat, we should set the shortcut to match (requires shortcut to have been created already)
```
Ana06, line 334: if the shortcut is not created, we do not reach this code, or am I missing something?

packages/common.vm/tools/vm.common/vm.common.psm1
emtuls: Hmmm. The `$pythonPath = (Get-Command python).Source` would likely error out. 

I can add a better check here to mitigate this, like:
```
If (($extension -eq ".py") -and (Get-Command python)) {
    $iconLocation = (Get-Command python).Source
}
Else {
???
}
```
Not sure what we should do if it isn't installed...Also, would we ever reach this if it wasn't if we're controlling the calls to `VM-Install-Shortcut`?

packages/common.vm/tools/vm.common/vm.common.psm1
emtuls: While I agree that we -should- execute python files with Python in the `$executablePath`, this was more of a fallback option.

I can remove it if we notate somewhere in the documentation that python tools should use Python for it's executable path and I'll go through each of our current Python tools list and make the changes. :)

packages/common.vm/tools/vm.common/vm.common.psm1
emtuls: That would actually be exactly what we want! If there is no shortcut, then we don't want to attempt to modify the shortcut icon if it isn't there. :)

packages/common.vm/tools/vm.common/vm.common.psm1
Ana06: please document this in the wiki, ideally this week as I'll be restructuring the wiki next week to propose a better version of the documentation. Do not worry much where the best place for this is, as I can move it next week. :wink:

packages/common.vm/tools/vm.common/vm.common.psm1
Ana06: I meant I found the comment difficult to understand and that could be simplified. I should have marked as nitpick

Ana06 (commented):
After the python checked is removed (and documented) I am ok with merging the .bat improvement. It is a nice improvement and an elegant solution, thanks @emtuls

packages/common.vm/tools/vm.common/vm.common.psm1
emtuls: Added!

Ana06:
@emtuls I think you got rid of your changes in the last commit. I have restored the changes in a new commit with you as author. Can you please check the changes and that the email in the commit is correct?

PR #859, opened 2024-01-24
Remove cd command in VM-Install-Shortcut

emtuls:
This fixes what was referenced here: https://github.com/mandiant/VM-Packages/issues/676#issuecomment-1877578921

After some testing, it looks like we do not need the `cd C:\Users\User\Desktop` piece of the command as we will already be in there due to `WorkingDirectory` being set there for `Install-ChocolateyShortcut`.

### DIFF
--- a/packages/common.vm/common.vm.nuspec
+++ b/packages/common.vm/common.vm.nuspec
@@ -2,7 +2,7 @@
 <package xmlns="http://schemas.microsoft.com/packaging/2015/06/nuspec.xsd">
   <metadata>
     <id>common.vm</id>
-    <version>0.0.0.20240119</version>
+    <version>0.0.0.20240123</version>
     <description>Common libraries for VM-packages</description>
     <authors>Mandiant</authors>
   </metadata>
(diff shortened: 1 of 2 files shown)


Ana06 (commented):
I am a bit confused... if this works... why doesn't it work in https://github.com/mandiant/VM-Packages/pull/818/files#diff-0938dff836c574dfc87598934cd56593b6cf2005e9a2598c08df2240a17a0244L46? 😕

emtuls:
@Ana06 Turns out that `Start In` is explicitly ignored for elevated command prompts to mitigate a potential security issue: https://superuser.com/a/1067979

For shortcuts that are command line based and require admin privileges, we should do what we were doing before where we set the `Target` to perform `cmd /K "cd $executableDir"`

Otherwise, we can just use the `WorkingDirectory` field. :)

Ana06:
@emtuls thanks for the research! I was really confused and this clarifies. I agree we should keep it so that it works for all cases and that we can close this PR.

PR #871, opened 2024-01-26
Add powershell support in VM-Install-Shortcut

emtuls:
This fixes https://github.com/mandiant/VM-Packages/issues/676

### DIFF
--- a/packages/common.vm/tools/vm.common/vm.common.psm1
+++ b/packages/common.vm/tools/vm.common/vm.common.psm1
@@ -219,7 +219,6 @@ function VM-Install-Raw-GitHub-Repo {
     )
     try {
         $toolDir = Join-Path ${Env:RAW_TOOLS_DIR} $toolName
-        $shortcutDir = Join-Path ${Env:TOOL_LIST_DIR} $category
 
         # Remove files from previous zips for upgrade
         VM-Remove-PreviousZipPackage ${Env:chocolateyPackageFolder}
@@ -242,15 +241,10 @@ function VM-Install-Raw-GitHub-Repo {
         }
 
         if ($powershellCommand) {
-            $executableArgs = "-ExecutionPolicy Bypass -NoExit -Command $powershellCommand"
-            $powershellPath = Join-Path "${Env:WinDir}\system32\WindowsPowerShell\v1.0" "powershell.exe" -Resolve
-            $shortcut = Join-Path $shortcutDir "$toolName.lnk"
-            Install-ChocolateyShortcut -shortcutFilePath $shortcut -targetPath $powershellPath -Arguments $executableArgs -WorkingDirectory $toolDir -IconLocation $powershell
+            VM-Install-Shortcut -toolName $toolName -category $category -arguments $powershellCommand -executableDir $toolDir -powershell
         } else {
-            $shortcut = Join-Path $shortcutDir "$toolName.lnk"
-            Install-ChocolateyShortcut -shortcutFilePath $shortcut -targetPath $toolDir
+            VM-Install-Shortcut -toolName $toolName -category $category -executablePath $toolDir
         }
-        VM-Assert-Path $shortcut
 
         return $toolDir
     } catch {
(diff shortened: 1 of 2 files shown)


packages/common.vm/tools/vm.common/vm.common.psm1
```diff
@@ -242,15 +241,11 @@ function VM-Install-Raw-GitHub-Repo {
         }
 
         if ($powershellCommand) {
-            $executableArgs = "-ExecutionPolicy Bypass -NoExit -Command $powershellCommand"
             $powershellPath = Join-Path "${Env:WinDir}\system32\WindowsPowerShell\v1.0" "powershell.exe" -Resolve
-            $shortcut = Join-Path $shortcutDir "$toolName.lnk"
-            Install-ChocolateyShortcut -shortcutFilePath $shortcut -targetPath $powershellPath -Arguments $executableArgs -WorkingDirectory $toolDir -IconLocation $powershell
+            VM-Install-Shortcut -toolName $toolName -category $category -executablePath $powershellPath -arguments $powershellCommand -executableDir $toolDir -powershell $true
```
Ana06, line 245: `executablePath` is not used for powerShell inside `VM-Install-Shortcut`. I think we need to make this argument optional to simplify the call for PowerShell and remove the need to assign `$powershellPath`

Ana06 (changes requested):
Nice! This is going to simplify the code and help keeping packages consistent. 😄  

I think we should not provide `$executablePath` as argument for PowerShell.

packages/common.vm/tools/vm.common/vm.common.psm1
emtuls: I think this was to make the shortcut icon show powershell, which seems a bit weird to do it this way looking back at it.

I can change it instead such that if `$powershell` is set to `$true`, that `$IconLocation = Join-Path "${Env:WinDir}\system32\WindowsPowerShell\v1.0" "powershell.exe" -Resolve` on ~line 297 just after setting the `$executableCmd` and `$executableArgs`

Ana06: we can change `$executablePath` to optional

Ana06 (changes requested):

packages/common.vm/tools/vm.common/vm.common.psm1
```diff
@@ -477,18 +478,9 @@ function VM-Install-Single-Ps1 {
...
-        $shortcut = Join-Path $shortcutDir "$toolName.lnk"
-        $targetCmd = Join-Path ${Env:WinDir} "system32\cmd.exe" -Resolve
-
-        if ($ps1Cmd) {
-            $targetArgs = "/K powershell.exe -ExecutionPolicy Bypass -NoExit -Command `"cd '$toolDir'; $ps1Cmd`""
-        } else {
-            $targetArgs = "/K powershell.exe -ExecutionPolicy Bypass -NoExit -Command `"cd '$toolDir'`""
-        }
-        $targetIcon = Join-Path (Join-Path ${Env:WinDir} "system32\WindowsPowerShell\v1.0") "powershell.exe" -Resolve
+        $powershellPath = Join-Path (Join-Path ${Env:WinDir} "system32\WindowsPowerShell\v1.0") "powershell.exe" -Resolve
```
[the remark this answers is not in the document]
emtuls: Ah yes. I took it out of one location and still left it in another. Will change it! 🙂

packages/common.vm/tools/vm.common/vm.common.psm1
```diff
@@ -266,11 +260,13 @@ function VM-Install-Shortcut{
         [string] $toolName,
         [Parameter(Mandatory=$true, Position=1)]
         [string] $category,
-        [Parameter(Mandatory=$true, Position=2)]
+        [Parameter(Mandatory=$false, Position=2)]
         [string] $executablePath,
         [Parameter(Mandatory=$false)]
         [bool] $consoleApp=$false,
         [Parameter(Mandatory=$false)]
+        [bool] $powershell=$false,
```
Ana06, line 268: [nitpick] it could be a switch instead of a bool. Or is there a reason for that?

Ana06 (approved):

packages/common.vm/tools/vm.common/vm.common.psm1
emtuls: I agree. It probably could do better as a switch. :)

Ana06 (approved):
LGTM :smile:

PR #873, opened 2024-01-29
Created new Appx function to accept Wildcards and edited 

geo-lit:
New Remove-Appx function can accept wildcards and iterates over the returned list of packages and deletes all matches if found.

New Win10 config file includes packages using wildcards to be version agnostic.


Ana06 (commented):
The code LGTM! @geo-lit you need to rebase to resolve merge conflicts.

Please also squash your commit to update the version in the same commit you modify the package code in order to keep a clean commit history that helps us knowing for example in which version a bug was introduced. Alternatively we can merge this PR with squash and merge.

PR #877, opened 2024-01-30
Fix the 7z command as 7zip is not a valid command

rethinksec:


Ana06 (commented):
@rethinksec thanks for the fix! As documented in our [Wiki](https://github.com/mandiant/VM-Packages/wiki/Package-Structure#version), you need to increase the version in the nuspec every time you update a package:

> **The version needs to be increased every time you modify an existing package.** For example, if you fix a bug in a package, you need to update the nuspec version string. Our guidance is to add the current date to the 4th segment using the format `YYYYMMDD`. For example, `1.6.2` becomes `1.6.2.20220113` and `3.02` becomes `3.02.0.20220113`. If the date is already in the 4th segment, you can increment it to the next day (e.g., `20220113` -> `20220114`) and when the PR is merged it should be good to go.

@rethinksec you also need to accept Google CLA for us to be able to merge this PR. Please let us know if you need any assistance.

[nitpick] I think it is also a good idea to request 7zip-15-05.vm >= 15.05.0.20231220 in the nuspec

rethinksec:
I have updated the nuspec file as requested & accept the Google CLA. Kindly review again.

Ana06 (approved):
Thanks @rethinksec!

Ana06:
@mandiant/commando-vm we are getting a:
```
throw "Non-Intel processor detected."
```
This makes the package failing in the CI and will produce that the package is not pushed to myget if we merge this PR. 😖  @day1player can you assist here? we need to either remove the check if it is not needed or disable this check in the CI.

Menn1s:
Sorry for the delay on this one; please see #892.

rethinksec:
I have updated the script as per #892

Ana06 (approved):
Thanks a lot @rethinksec!! I am going to merge this PR with `squash and merge` to get rid of the `Update hashcat.vm.nuspec` and `Merge branch 'mandiant:main' into main` commits in order to keep a clean commit history. For future contributions it would be appreciated if you could squash your commit instead.

Issue #831, opened 2024-01-12
peanatomist.vm broken in GH actions


Ana06:
### Details

GitHub actions (our daily run) has problems downloading https://rammerlabs.alidml.ru/files/0000-0002-29CD-0000/PEAnatomist-0.2.zip but I can download it. @vm-packages any ideas how to fix this issue?

Ana06:
Fixed in #878

PR #878, opened 2024-01-30
Update chocolateyinstall.ps1 from PEAnatomist

bsempir0x65:
Fix Issue that the website sometime works sometimes not. Copied the idea from scoop apps. #831


bsempir0x65:
Updated the spec number and will test it later on the new instance of Flare i made.
For information here is the scoop app i meant https://github.com/ACooper81/scoop-apps/blob/master/bucket/PEAnatomist-Portable.json

Ana06 (approved):
I have tested locally and it seems to work well! 🎉  Thanks for the contribution 
bsempir0x65!

PR #879, opened 2024-01-30
[libraries.python3] Install lznt1 Python library

Ana06:
LZNT1 compression is widely used in malware. So I think it would be useful to be able to use LZNT1 compression/decompression in FLARE-VM. Add the lznt1 Python library to libraries.python3.vm to provide this functionality.

Also **sort libraries alphabetically**.

There is also an issue in CyberChef to request LZNT1: https://github.com/gchq/CyberChef/issues/534 but It doesn't look like it will be added in the close feature. 😞


binjo (approved):
LGTM

mr-tz (approved):

emtuls (approved):

Issue #889, opened 2024-02-02
OpenJDK Path Uses Hardcoded Version Causing Auto Update To Fail


emtuls:
### What's the problem?

The OpenJDK package uses a hardcoded version number to locate `javaw.exe` as seen here: https://github.com/mandiant/VM-Packages/blob/main/packages/openjdk.vm/tools/chocolateyinstall.ps1#L6

This is causing the auto updater to fail: https://github.com/mandiant/VM-Packages/pull/864

### Steps to Reproduce

View logs for PR 864 here: https://github.com/mandiant/VM-Packages/actions/runs/7733944163/job/21094276941?pr=864

### Environment

Env Independent

### Detected by test suite

No

Ana06:
As mentioned in https://github.com/mandiant/VM-Packages/pull/890#discussion_r1475661066 (so that it is documented in the issue as well):
> I think we shouldn't update core tools (like OpenJDK, Python, etc.) automatically. The purposes of these metapackages is that only one version of the tool is required for all packages to ensure there are not dependency conflicts when installing several packages. In the other metapackages like this, we are fixing the dependency to prevent updates. For example in python3.vm:
`<dependency id="python3" version="[3.10.0, 3.11.0)" />`

Ana06:
But there may still be a bug here. Our bot should not send broken updates, which seems to be happening. Any idea where the issue could be @emtuls?

Ana06:
I think the issue may be related to updating apktool.vm together with openjdk.vm (a dependency of apktool.vm) and only checking that apktool.vm succeeds to install. I having analysed the issue in detail though.

emtuls:
Yea, when I looked at it, it seemed to be an issue with `openjdk` failing to update which was a dependency for `apktool`. Didn't think that when updating a dependency in the same run may cause an issue like that, but it's good to note.

Fixing the `openjdk` version in https://github.com/mandiant/VM-Packages/pull/890 should fix this. :)

PR #864, opened 2024-01-24
:robot: Package update

vm-packages:
Automated package update

PR #890, opened 2024-02-02
Fix Openjdk Hardcoded Version Issue

emtuls:
This fixes https://github.com/mandiant/VM-Packages/issues/889 which is blocking https://github.com/mandiant/VM-Packages/pull/864

### DIFF
--- a/packages/openjdk.vm/openjdk.vm.nuspec
+++ b/packages/openjdk.vm/openjdk.vm.nuspec
@@ -2,12 +2,12 @@
 <package xmlns="http://schemas.microsoft.com/packaging/2015/06/nuspec.xsd">
   <metadata>
     <id>openjdk.vm</id>
-    <version>21.0.1</version>
+    <version>0.0.0.20240202</version>
     <authors>Oracle</authors>
     <description>Metapackage for OpenJDK to ensure all packages use the same OpenJDK version.</description>
     <dependencies>
       <dependency id="common.vm" />
-      <dependency id="openjdk" version="[21.0.1]" />
+      <dependency id="openjdk" version="(21.0.0, 21.0.1]" />
     </dependencies>
   </metadata>
 </package>


packages/openjdk.vm/openjdk.vm.nuspec
```diff
@@ -2,12 +2,12 @@
...
   <metadata>
     <id>openjdk.vm</id>
-    <version>21.0.1</version>
+    <version>21.0.2</version>
     <authors>Oracle</authors>
     <description>Metapackage for OpenJDK to ensure all packages use the same OpenJDK version.</description>
     <dependencies>
       <dependency id="common.vm" />
-      <dependency id="openjdk" version="[21.0.1]" />
+      <dependency id="openjdk" version="[21.0.2]" />
```
Ana06, line 10: Thanks for the effort to fix this issue @emtuls! But I think we shouldn't update core tools (like OpenJDK, Python, etc.) automatically. The purposes of these metapackages is that only one version of the tool is required for all packages to ensure there are not dependency conflicts when installing several packages. In the other metapackages like this, we are fixing the dependency to prevent updates. For example in python3.vm:
`<dependency id="python3" version="[3.10.0, 3.11.0)" />`

I think we should do the same here and then we can continue to use a hardcoded path in order to keep a simpler code. Something like:
```suggestion
      <dependency id="openjdk" version="(21.0.0, 21.0.2]" />
```

Ana06 (changes requested):

packages/openjdk.vm/openjdk.vm.nuspec
emtuls: Hmmm. In order to fix this, the linter says I need to update the versioning. Should I set the version to `21.0.1.2024.0202`?

I believe we want `<dependency id="openjdk" version="(21.0.0, 21.0.1]" />` as opposed to `21.0.2` due to the hardcoded path being `21.0.1`

Ana06: I would keep it consistent with other similar packages. In [python3.vm](https://github.com/mandiant/VM-Packages/blob/main/packages/python3.vm/python3.vm.nuspec#L5) we are using `0.0.0.YYYYMMDD` which also follows our [metapackage version documentation](https://github.com/mandiant/VM-Packages/wiki/Package-Structure#metapackages).

packages/openjdk.vm/openjdk.vm.nuspec
```diff
@@ -2,12 +2,12 @@
 <package xmlns="http://schemas.microsoft.com/packaging/2015/06/nuspec.xsd">
   <metadata>
     <id>openjdk.vm</id>
-    <version>21.0.1</version>
+    <version>21.0.1.20240202</version>
```
Ana06, line 5: To be consistent with [python3.vm](https://github.com/mandiant/VM-Packages/blob/main/packages/python3.vm/python3.vm.nuspec#L5) and [nodejs.vm](https://github.com/mandiant/VM-Packages/blob/main/packages/nodejs.vm/nodejs.vm.nuspec#L5) this should be 
```suggestion
    <version>0.0.0.20240202</version>
```

This means we need to remove the current package/s from myget before merging this PR. @emtuls can you update the version and I'll take care of removing the packages and merging this PR?

But this may need further discussion

Ana06 (changes requested):

Issue #653, opened 2023-08-30
Hashcat VM


godylockz:
### What's the problem?

[hashcat.vm] vm.common.psm1 [+] ERROR : [ERR] Non-Intel processor detected.

Hashcat should run regardless on the processor / environment.

### Steps to Reproduce

1. Install commando vm 
2. choco install hashcat.vm

### Environment

Windows 10

### Detected by test suite

No

Ana06:
@Menn1s do you have some input on this check?

day1player:
I'm pretty sure hashcat requires an intel CPU but we can do some testing to double check.. @Menn1s or @nos3curity could one of you test this out?

day1player:
Ok looking at the site this is what I found:
<img width="457" alt="image" src="https://github.com/mandiant/VM-Packages/assets/37482972/b783f655-1780-4879-853f-e8e5bed5aac8">

Maybe we need to check if there is an OpenCL runtime instead of the processor? Will have to look into that.

day1player:
@MalwareMechanic I havent been able to look into this as all of my CPUs are intel but looking at the documentation it says "everything that comes with an OpenCL runtime"

Ana06:
@day1player can the check be then updated to fix the package in our test suite?

day1player:
I think this one we will have to look into on how to detect the OpenCL runtime

Ana06:
closing as I think it was fixed in #892. Please let me know if the issue is not fixed.

PR #892, opened 2024-02-02
Removing hashcat intel check

### LINKED ISSUE #742: Hashcat Failure Due to Processor Check
The hashcat package will fail if the system's processor is not "GenuineIntel", see below:

https://github.com/mandiant/VM-Packages/blob/eac039ddb9852d3cc155a400895de4bebeb1ccfc/packages/hashcat.vm/tools/chocolateyinstall.ps1#L20-L26

However, the GitHub actions to test installation may use non-GenuineIntel processors and the package will fail. Since we cannot guarantee the processor type during the test installation, I propose we complete the installation and raise a warning to the user that their system does not support this tool due to a non-GenuineIntel processor.

This allows the test suite to recognize a successful installation and will reduce confusion as to why the package is failing.

@mandiant/commando-vm thoughts since this is mainly a Commando VM tool?

Menn1s:
Removed the compatibility check; this can be handled by the installation process and the profiles used. Fixes #742 and #653. Also comes up in #877.

Hashcat should support a variety of systems and the check was not complex enough to accurately determine compatibility.

It does not seem trivial to get OpenCL on an ARM Windows system, however, I think this can be up to the user or a future fix to meet the compatibility.

### DIFF
--- a/packages/hashcat.vm/hashcat.vm.nuspec
+++ b/packages/hashcat.vm/hashcat.vm.nuspec
@@ -2,7 +2,7 @@
 <package xmlns="http://schemas.microsoft.com/packaging/2015/06/nuspec.xsd">
   <metadata>
     <id>hashcat.vm</id>
-    <version>6.2.6</version>
+    <version>6.2.6.20240212</version>
     <authors>Jens Steube (jsteube)</authors>
     <description>Hashcat is a fast password recovery utility.</description>
     <dependencies>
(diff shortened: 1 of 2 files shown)


Ana06 (commented):
Thanks for removing the intel check @Menn1s! can you update the package version? 😄  Also, it seems you have commits here that don't belong to this PR. Let me know if you need help while rebasing, I may be able to help. 😉

Ana06 (commented):
@Menn1s you have forgotten to remove the `$processor` variable and the bot is not happy. Would it be ok for you if we merge https://github.com/mandiant/VM-Packages/pull/877 that includes this change (and where the test pass already(?

Ana06:
@Menn1s approved that we merge https://github.com/mandiant/VM-Packages/pull/877 instead of this PR (documenting discussion outside GH)

Issue #902, opened 2024-02-16
Re-arrange Desktop icons in VM-Clean-Up 


Ana06:
### Details

`VM-Clean-Up` may delete Desktop icons leaving empty spaces which looks very ugly. I would like to sort/re-arrange the icons so that it looks nice again.

![image](https://github.com/mandiant/VM-Packages/assets/16052290/362a4b80-b269-4be1-a7be-a986c0d2bfdd)

This is needed to automate building FLARE-VM VMs.

@vm-packages any ideas how to do this?

PR #909, opened 2024-02-20
[VM-Clean-Up] Sort Desktop icons by type

Ana06:
Introduce `VM-Sort-Desktop-Icons` to sort Desktop icons by type using `WScript.Shell` to replicate the manual steps using the keyboard. This function is called from `VM-Remove-DesktopFiles` (called from `VM-Clean-Up`) to avoid leaving empty spaces after removing Desktop icons.

I would appreciate if someone could also test this code locally. Note you don't need to install the whole FLARe-VM or even the modified package to test the changes. It is enough to copy the functions `VM-Remove-DesktopFiles` and `VM-Remove-DesktopFiles` from the modified `packages/common.vm/tools/vm.common/vm.common.psm1`.

I am using Windows 10 and I am not sure if this code works in Windows 11 and we should prevent from running it in Windows 11 (or add a version for Windows 11). Note this code has to be run manually (by running `VM-Clean-Up` that the FLARE team is using to automate FLARE-VM builds) and is not part of the installer.

Closes https://github.com/mandiant/VM-Packages/issues/902

### DIFF
--- a/packages/common.vm/tools/vm.common/vm.common.psm1
+++ b/packages/common.vm/tools/vm.common/vm.common.psm1
@@ -1288,6 +1288,16 @@ public class Shell {
     }
 }
 
+# Sort Desktop icons by item type using WScript.Shell to replicate the manual steps
+function VM-Sort-Desktop-Icons {
+    VM-Write-Log "INFO" "Sorting Desktop icons"
+    (New-Object -ComObject Shell.Application).toggleDesktop();
+    Start-Sleep -Milliseconds 100;
+    $objShell = New-Object -ComObject WScript.Shell;
+    $objShell.SendKeys("^a {F5}+{F10}oi");
+    Start-Sleep -Milliseconds 100;
+}
+
 # Usage example:
 # VM-Remove-DesktopFiles -excludeFolders "Labs", "Demos" -excludeFiles "MICROSOFT Windows 10 License Terms.txt", "Labs.zip"
 # The function is run against both the Current User and 'Public' desktops due to some cases where desktop icons showing on
@@ -1327,6 +1337,8 @@ function VM-Remove-DesktopFiles {
             }
         }
     }
+
+    VM-Sort-Desktop-Icons
 }
 
 function VM-Clear-TempAndCache {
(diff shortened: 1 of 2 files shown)


emtuls:
I tested this and ran into a couple of issues depending on what might have been selected or not in the background before executing the function. Sometimes it didn't work and other times it would open a folder up that was on the desktop.

I believe I found an alternative that seems to work though!

I changed the key strokes line to be `$objShell.SendKeys("^a {F5}+{F10}oi");`

packages/common.vm/tools/vm.common/vm.common.psm1
```diff
@@ -1288,6 +1288,16 @@ public class Shell {
     }
 }
 
+# Sort Desktop icons by item type using WScript.Shell to replicate the manual steps
+function VM-Sort-Desktop-Icons {
+    VM-Write-Log "INFO" "Sorting Desktop icons"
+    (New-Object -ComObject Shell.Application).toggleDesktop();
+    Start-Sleep -Milliseconds 100;
+    $objShell = New-Object -ComObject WScript.Shell;
+    $objShell.SendKeys("^{ }+{F10}oi");
```
emtuls, line 1297: This didn't seem to always work depending on what might have been in focus before executing. I tried it with a folder having been selected before switching to the powershell terminal and executing the command and also with a folder not selected and ran into varying issues.

I was able to modify the key strokes to this and it seems to work more consistently:

emtuls (changes requested):
Comments explain issue.

packages/common.vm/tools/vm.common/vm.common.psm1
Ana06: doesn't Control+space (`^{ }`) unselect the folder for you?

Ana06: I see, if there are several folders selected it does run into issues

Ana06: I'll update the code, good catch @emtuls!

Issue #719, opened 2023-10-24
add IDA plugins listed at vmallet


williballenthin:
### Details

There's a nice list of actively maintained IDA plugins here: https://vmallet.github.io/ida-plugins/

We should use this list as inspiration for IDA plugins to add to FLARE-VM.

williballenthin:
- https://github.com/gaasedelen/lighthouse
  - https://github.com/A200K/IDA-Pro-SigMaker
  - https://github.com/airbus-cert/comida
  - https://github.com/airbus-cert/ttddbg
  - https://github.com/hasherezade/ida_ifl
  - https://github.com/williballenthin/idawilli plugins and scripts
  - https://github.com/alexander-hanel/FunctionTrapperKeeper
  - https://github.com/Jinmo/ifred
  - https://github.com/danigargu/deREferencing
  - https://gitlab.com/eshard/d810

Ana06:
@mandiant/flare-vm which of this plugins would you like to have in FLARE-VM?

I propose we start adding the following ones from the list provided above:
- https://github.com/gaasedelen/lighthouse
- https://github.com/A200K/IDA-Pro-SigMaker
- https://github.com/airbus-cert/comida
- https://github.com/hasherezade/ida_ifl
- https://gitlab.com/eshard/d810

I would suggest also adding https://github.com/nihilus/idastealth

https://github.com/airbus-cert/ttddbg seems to have issues with IDA 8, so I would leave it by now.

I think we could consider adding some parts of https://github.com/williballenthin/idawilli and https://github.com/Ana06/idapython but this requires some discussions and I think we should not address this as part of this PR.

binjo:
I'd love to have these two:
- https://github.com/oopsmishap/HexRaysPyTools
- https://github.com/gaasedelen/tenet

emtuls:
I second these ones at the very least (the others look good as well!):
- https://github.com/danigargu/deREferencing
- https://github.com/gaasedelen/tenet
- https://github.com/gaasedelen/lighthouse
- https://github.com/hasherezade/ida_ifl
- https://github.com/airbus-cert/comida

Ana06:
[...]

PR #911, opened 2024-02-20
Add the lighthouse IDA Pro plugin

d35ha:
Required by https://github.com/mandiant/VM-Packages/issues/719.
Only works with IDA Pro and it requires DynamoRIO or Intel PIN.

### DIFF
--- a/packages/ida.plugin.lighthouse.vm/tools/chocolateyinstall.ps1
+++ b/packages/ida.plugin.lighthouse.vm/tools/chocolateyinstall.ps1
@@ -1,0 +1,50 @@
+$ErrorActionPreference = 'Stop'
+Import-Module vm.common -Force -DisableNameChecking
+
+try {
+    # Download the plugin
+    $toolName    = "lighthouse"
+    $toolVersion = "0.9.3"
+    $toolDir     = Join-Path ${Env:chocolateyPackageFolder} "temp_$([guid]::NewGuid())"
+    $toolUrl     = "https://github.com/gaasedelen/lighthouse/archive/refs/tags/v0.9.3.zip"
+    $toolSha256  = "b9f1888827f68d69672d1202783b47f7603fb80f42f9b70edeb9349d40c4f33c"
+    $packageArgs =
+    @{
+        packageName   = ${Env:ChocolateyPackageName}
+        unzipLocation = $toolDir
+        url           = $toolUrl
+        checksum      = $toolSha256
+        checksumType  = "sha256"
+    }
+    Install-ChocolateyZipPackage @packageArgs
+
+    # Install the plugin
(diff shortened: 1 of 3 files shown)


packages/ida.plugin.lighthouse.vm/tools/chocolateyinstall.ps1
```diff
@@ -0,0 +1,50 @@
...
+    # Download the plugin
+    $toolName    = "lighthouse"
+    $toolVersion = "0.9.3"
+    $toolDir     = Join-Path ${Env:chocolateyPackageFolder} "temp_$([guid]::NewGuid())"
+    $toolUrl     = "https://github.com/gaasedelen/lighthouse/archive/refs/tags/v0.9.3.zip"
+    $toolSha256  = "b9f1888827f68d69672d1202783b47f7603fb80f42f9b70edeb9349d40c4f33c"
+    $packageArgs =
+    @{
+        packageName   = ${Env:ChocolateyPackageName}
+        unzipLocation = $toolDir
```
Ana06, line 14: Use `$pluginsDir` directly as unzip location instead of downloading to `%TEMP%` and copying items over

packages/ida.plugin.lighthouse.vm/tools/chocolateyinstall.ps1
```diff
@@ -0,0 +1,50 @@
...
+
+    # Remove the temp installation folder
+    Remove-Item $toolDir -Recurse
+
+    # Install the code coverage binaries
+    $toolName  = "CodeCoverage"
+    $zipUrl    = "https://github.com/gaasedelen/lighthouse/releases/download/v$toolVersion/CodeCoverage-v$toolVersion-98830.zip"
+    $zipSha256 = "5b3e9659e934f8d7f4751cc0e20aa76de28822bc4f28b4936aa9be9c9e108595"
+
+    $toolDir = Join-Path ${Env:RAW_TOOLS_DIR} $toolName
```
Ana06, line 36: why do we need this in `%RAW_TOOLS_DIR%`?

Ana06 (changes requested):
@d35ha thanks for the PR!

@d35ha 
>  it requires DynamoRIO or Intel PIN.
Then they should be installed by this package (likely as a dependency)


As mentioned in https://github.com/mandiant/VM-Packages/pull/913#pullrequestreview-1895632502, we can only merge PRs in which the linter and test succeed. We need to exclude packages without categories (no shortcut is created in the Tools directory) in the linter. Now we have `ida.plugin.capa.vm` in the [`EXCLUSIONS` array](https://github.com/mandiant/VM-Packages/blob/main/scripts/test/lint.py#L317). I suggest we change it by `ida.plugin`, which will exclude all IDA plugins.

PR #913, opened 2024-02-21
Add the SigMaker IDA Pro plugin

d35ha:
Required by https://github.com/mandiant/VM-Packages/issues/719.

PR #914, opened 2024-02-21
Add the comida IDA Pro plugin

d35ha:
Required by https://github.com/mandiant/VM-Packages/issues/719.


Ana06 (commented):
The comments in https://github.com/mandiant/VM-Packages/pull/913 also apply here.

PR #933, opened 2024-03-04
[installer] Fix creation of Admin Command Prompt

Ana06:
Fix argument in the creation of the Admin Command Prompt shortcut (added to the FLARE-VM taskbar) that fails the installation of `installer.vm`. The bug was introduced in https://github.com/mandiant/VM-Packages/pull/818, but it was not used until the version was fixed 5 days ago in https://github.com/mandiant/VM-Packages/pull/925.

I tried to fix this issue in https://github.com/mandiant/VM-Packages/pull/931 :see_no_evil: I should have tested that properly. Sorry for that. I have tested this change locally (by using our `test_install.ps1` script removing the `installer.vm` exception).

Fixes https://github.com/mandiant/flare-vm/issues/573

### DIFF
--- a/packages/installer.vm/installer.vm.nuspec
+++ b/packages/installer.vm/installer.vm.nuspec
@@ -2,7 +2,7 @@
 <package xmlns="http://schemas.microsoft.com/packaging/2015/06/nuspec.xsd">
   <metadata>
     <id>installer.vm</id>
-    <version>0.0.0.20240304</version>
+    <version>0.0.0.20240305</version>
     <authors>Mandiant</authors>
     <description>Generic installer for custom virtual machines.</description>
     <dependencies>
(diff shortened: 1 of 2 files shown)


emtuls (approved):
Tested locally and it works on my end. Good to merge!

packages/installer.vm/installer.vm.nuspec
```diff
@@ -2,7 +2,7 @@
 <package xmlns="http://schemas.microsoft.com/packaging/2015/06/nuspec.xsd">
   <metadata>
     <id>installer.vm</id>
-    <version>0.0.0.20240304</version>
+    <version>0.0.0.20240305</version>
```
Ana06, line 5: we are in the future :rocket:

4zt3ch: so can i just use the installer.ps1 script again on the read me page? or do i have to do something special now?

Ana06: you can just use the installer.ps1 in the flare-vm repo. It will take the latest package from myget

PR #935, opened 2024-03-04
Package proposal: magika

emtuls:
This package is an awesome open source utility that can detect file formats much faster and better than the standard `file` utility that we currently use.

https://github.com/google/magika

### DIFF
--- a/packages/magika.vm/magika.vm.nuspec
+++ b/packages/magika.vm/magika.vm.nuspec
@@ -1,0 +1,13 @@
+<?xml version="1.0" encoding="utf-8"?>
+<package xmlns="http://schemas.microsoft.com/packaging/2015/06/nuspec.xsd">
+  <metadata>
+    <id>magika.vm</id>
+    <version>0.5.1</version>
+    <authors>Yanick Fratantonio, Luca Invernizzi, Marina Zhang, Giancarlo Metitieri, Thomas Kurt, Francois Galilee, Alexandre Petit-Bianco, Loua Farah, Ange Albertini, Elie Bursztein</authors>
+    <description>Magika is an AI powered file type detection tool that uses deep learning to provide accurate detection.</description>
+    <dependencies>
+      <dependency id="common.vm" />
+      <dependency id="python3.vm" />
+    </dependencies>
+  </metadata>
+</package>
--- a/packages/magika.vm/tools/chocolateyinstall.ps1
+++ b/packages/magika.vm/tools/chocolateyinstall.ps1
@@ -1,0 +1,16 @@
+$ErrorActionPreference = 'Stop'
+Import-Module vm.common -Force -DisableNameChecking
+
+try {
+    $toolName = 'magika'
+    $category = 'File Information'
+
+    VM-Pip-Install $toolName
+
+    $executablePath = "$(where.exe $toolName)"
+    $arguments = "--help"
+
+    VM-Install-Shortcut $toolName $category $executablePath -consoleApp $true -arguments $arguments -iconLocation $iconLocation
+} catch {
+    VM-Write-Log-Exception $_
+}
--- a/packages/common.vm/common.vm.nuspec
+++ b/packages/common.vm/common.vm.nuspec
@@ -2,7 +2,7 @@
 <package xmlns="http://schemas.microsoft.com/packaging/2015/06/nuspec.xsd">
   <metadata>
     <id>common.vm</id>
-    <version>0.0.0.20240404</version>
+    <version>0.0.0.20240409</version>
     <description>Common libraries for VM-packages</description>
     <authors>Mandiant</authors>
   </metadata>
--- a/packages/common.vm/tools/vm.common/vm.common.psm1
+++ b/packages/common.vm/tools/vm.common/vm.common.psm1
@@ -1550,3 +1550,13 @@ function VM-Get-MSIInstallerPathByProductName {
         VM-Write-Log-Exception $_ "An error occurred: $_"
     }
 }
+
+function VM-Pip-Install {
+    param (
+        [string]$package
+    )
+    # Create output file to log python module installation details
+    $outputFile = VM-New-Install-Log ${Env:VM_COMMON_DIR}
+
+    Invoke-Expression "py -3.10 -m pip install $package --disable-pip-version-check 2>&1 >> $outputFile"
+}
--- a/packages/magika.vm/tools/chocolateyuninstall.ps1
+++ b/packages/magika.vm/tools/chocolateyuninstall.ps1
@@ -1,0 +1,9 @@
+$ErrorActionPreference = 'Continue'
+Import-Module vm.common -Force -DisableNameChecking
+
+$toolName = 'magika'
+$category = 'File Information'
+
+Invoke-Expression "py -3.10 -m pip uninstall $toolName -y"
+
+VM-Uninstall $toolName $category


emtuls:
@Ana06 `pip` seems to produce an error even though everything installs correctly, causing an issue with the package saying that it installed incorrectly and will stop due to `$ErrorActionPreference = 'Stop'`

I attempted to copy what was done in `libraries.python3.vm`, but it did not work unless I changed `$ErrorActionPreference = 'Continue'` at the top.

Do you have any suggestions on other ways to mitigate this? If this is the only way, then I will have to add an Exclusion inside of `lint.py` and this will also be something that needs to be done for all future packages that install a tool via `pip`.

packages/magika.vm/magika.vm.nuspec
```diff
@@ -0,0 +1,13 @@
+<?xml version="1.0" encoding="utf-8"?>
+<package xmlns="http://schemas.microsoft.com/packaging/2015/06/nuspec.xsd">
+  <metadata>
+    <id>magika.vm</id>
+    <version>0.5.0</version>
+    <authors>Google</authors>
+    <description>Magika is an AI powered file type detection tool that uses deep learning to provide accurate detection.</description>
+    <dependencies>
+      <dependency id="common.vm" />
+      <dependency id="python3.vm" />
```
Ana06, line 10: we need to require `libraries.python3.vm` to ensure there are not incompatibilities and both use the same version.

packages/magika.vm/tools/chocolateyinstall.ps1
```diff
@@ -0,0 +1,19 @@
+$ErrorActionPreference = 'Continue'
+Import-Module vm.common -Force -DisableNameChecking
+
+try {
+    $toolName = 'magika'
+    $category = 'File Information'
+    $executablePath = $toolName
+    $arguments = " -h"
+    $iconLocation = Join-Path ${Env:WinDir} "system32\cmd.exe" -Resolve
```
Ana06, line 9: why is this needed? can we move this to the `VM-Install-Shortcut` helper instead?

packages/magika.vm/tools/chocolateyinstall.ps1
```diff
@@ -0,0 +1,19 @@
+$ErrorActionPreference = 'Continue'
```
Ana06, line 1: The only reason we have `$ErrorActionPreference = 'Continue'` in libraries.python3.vm is to install other libraries if one fails. There is only one library here, so we want to ensure it installs without errors. Which error are you exactly getting? (If you change this to make the linter happy we should be able to see it in the run)

packages/magika.vm/tools/chocolateyinstall.ps1
```diff
@@ -0,0 +1,19 @@
...
+    $toolName = 'magika'
+    $category = 'File Information'
+    $executablePath = $toolName
+    $arguments = " -h"
+    $iconLocation = Join-Path ${Env:WinDir} "system32\cmd.exe" -Resolve
+
+    # Create output file to log python module installation details
+    $outputFile = VM-New-Install-Log ${Env:VM_COMMON_DIR}
+
+    Invoke-Expression "py -3.10 -m pip install $toolName 2>&1 >> $outputFile"
```
Ana06, line 14: What about moving this to a `VM-Pip-Install` helper so that installing python libraries in the future is as easy as `VM-Pip-Install $toolName`

packages/magika.vm/tools/chocolateyinstall.ps1
```diff
@@ -0,0 +1,19 @@
...
+
+try {
+    $toolName = 'magika'
+    $category = 'File Information'
+    $executablePath = $toolName
+    $arguments = " -h"
+    $iconLocation = Join-Path ${Env:WinDir} "system32\cmd.exe" -Resolve
+
+    # Create output file to log python module installation details
+    $outputFile = VM-New-Install-Log ${Env:VM_COMMON_DIR}
```
Ana06, line 12: We need to keep the logs for all python libraries consistent. Either we should use the same file for all of them, or at least different files within the same location.

packages/magika.vm/magika.vm.nuspec
```diff
@@ -0,0 +1,13 @@
+<?xml version="1.0" encoding="utf-8"?>
+<package xmlns="http://schemas.microsoft.com/packaging/2015/06/nuspec.xsd">
+  <metadata>
+    <id>magika.vm</id>
+    <version>0.5.0</version>
+    <authors>Google</authors>
```
Ana06, line 6: According to the [project page](https://github.com/google/magika?tab=readme-ov-file#citation), this is not an official Google project and the authors are: `Fratantonio, Yanick and Invernizzi, Luca and Zhang, Marina and Metitieri, Giancarlo and Kurt, Thomas and Galilee, Francois and Petit-Bianco, Alexandre and Farah, Loua and Albertini, Ange and Bursztein, Elie`

Ana06 (changes requested):

packages/magika.vm/tools/chocolateyinstall.ps1
emtuls: This one is slightly tricky...

Whenever we use `-consoleApp`, we expect the `executablePath` to point to an executable, which we can then use as an icon.

For this case it's a little different because there is no executable, it's just a command line command so it leads to no useable icon. I can pass in "the path to cmd.exe" here, but I don't think this makes it any better imo and if anything, it's actually worse because it starts to bloat the `Target Path` in the shortcut.

Inside of `VM-Install-Shortcut` I can't set everything that is a `consoleApp` to use `cmd.exe` as it's icon, because that's not always the case, and I can't think of a way that I can test if `executablepath` actually points to a valid "executable" vs "a command" in a way that wouldn't throw an error (by using an assert).

One option I could think of would be to have a different argument in `VM-Install-Shortcut` which would be for `consoleCommand` to fit this case, but I don't know if that's any cleaner than just having `iconLocation` set to `cmd.exe` like how it is here currently, because of how rare of a case this is.

packages/magika.vm/tools/chocolateyinstall.ps1
emtuls: I agree, but I can't seem to find a good way to programmatically get the same location, as it's located in `C:\ProgramData\chocolatey\lib\libraries.python3.vm\tools` and is based on the "Invocation Location": `$(Split-Path -parent $MyInvocation.MyCommand.Definition)`

We might need to adjust this location for `libraries.python3.vm` to be something else so that it's easier to find. I also think that the location that it ends up in is really weird, since most things `chocolatey` end up in `C:\ProgramData\chocolatey\logs`, and then we also use `C:\ProgramData\_VM` as a location for our logs.

emtuls: Added a PR to adjust `libraries.python3.vm` to use the same location as this package, which is more standardized.

https://github.com/mandiant/VM-Packages/pull/946

packages/magika.vm/tools/chocolateyinstall.ps1
Ana06: Thanks for the detailed explanation @emtuls. The issue is that you are using `magika` as `$executablePath`. I think it is counteractive to not use a path for a variable call `$executablePath`. I suggest  `$executablePath = $(where.exe magika)`. This way we use the binary that was already an icon.

image

packages/magika.vm/tools/chocolateyinstall.ps1
```diff
@@ -0,0 +1,16 @@
...
+try {
+    $toolName = 'magika'
+    $category = 'File Information'
+    $executablePath = "$toolName"
+    $arguments = "-h"
+    $iconLocation = Join-Path ${Env:WinDir} "system32\cmd.exe" -Resolve
+
+    VM-Pip-Install $toolName
+
+    VM-Install-Shortcut $toolName $category $executablePath -consoleApp $true -arguments $arguments -iconLocation $iconLocation
```
Ana06, line 13: Proposed changed:

```suggestion
    $executablePath = $(where.exe magika)
    $arguments = "--help"

    VM-Pip-Install $toolName
    VM-Install-Shortcut $toolName $category $executablePath -consoleApp $true -arguments $arguments
```

Ana06 (changes requested):

Ana06 (approved):

Issue #934, opened 2024-03-04
microsoft-windows-terminal.vm is broken


Ana06:
### What's the problem?

Installing `microsoft-windows-terminal.vm` does not display any errors but the shortcut is broken and the tool does not appear to be installed
![image](https://github.com/mandiant/VM-Packages/assets/16052290/6cd603c8-6bd7-4ce3-b67a-f7c84a39cd0f)

### Steps to Reproduce

1. run `choco install microsoft-windows-terminal.vm` in an minimal FLARE-VM installation (only `hashmyfiles` and notepadplusplus`  installed)

### Environment

Virtualization software: **Virtualbox**
VM OS version: `10.0.19045`
VM PowerShell version: `5.1.19041.1682`
VM Chocolatey version: `2.2.2`
VM Boxstarter version: `3.0.3`

```
PS C:\Users\flare > VM-Get-Host-Info
2024/03/04 19:36:13 vm.common.psm1 [+] INFO : Host Information

VM OS version and Service Pack
-----


Version                 : 10.0.19045
BuildNumber             : 19045
OSArchitecture          : 64-bit
ServicePackMajorVersion : 0
Caption                 : Microsoft Windows 10 Pro





VM OS RAM (MB)
-----
0


VM OS HDD Space / Usage
-----

DeviceID DriveType ProviderName VolumeName       Size        FreeSpace
-------- --------- ------------ ----------       ----        ---------
A:       2
C:       3                                       85897244672 65030762496
D:       5                      VBox_GAs_7.0.14  53526528    0
E:       5                      2024-03-04T18:18 53549056    0




VM AV Details
-----
AntiVirusProduct classname does not exist...

VM PowerShell Version
-----
5.1.19041.1682

VM CLR Version
-----
4.0.30319.42000

VM Chocolatey Version
-----
2.2.2

VM Boxstarter Version
-----

Boxstarter|3.0.3
Boxstarter.Bootstrapper|3.0.3
Boxstarter.Chocolatey|3.0.3
Boxstarter.Common|3.0.3
Boxstarter.HyperV|3.0.3
Boxstarter.WinConfig|3.0.3




VM Installed Packages
-----
Boxstarter|3.0.3
Boxstarter.Bootstrapper|3.0.3
Boxstarter.Chocolatey|3.0.3
Boxstarter.Common|3.0.3
Boxstarter.HyperV|3.0.3
Boxstarter.WinConfig|3.0.3
chocolatey|2.2.2
[...]

PR #936, opened 2024-03-05
Package Fix: Microsoft-Windows-Terminal

emtuls:
This fixes https://github.com/mandiant/VM-Packages/issues/934 by just downloading the latest package from the github and using it directly.

I made a few other changes to this as well, like a name change from `microsoft-windows-terminals` to a more simple `windows-terminal`, which is what most people in the community call it anyway.

I had to do a custom installation instead of our helper functions due to needing `runAsAdmin` which is only supported by calling `Install-Shortcut` directly and not able to be done from `VM-Install-From-Zip` or `VM-Install-Raw-GitHub-Repo`.

Fixes https://github.com/mandiant/VM-Packages/issues/934


packages/windows-terminal.vm/tools/chocolateyinstall.ps1
```diff
@@ -0,0 +1,41 @@
...
+      url            = $zipUrl
+      checksum       = $zipSha256
+      checksumType   = 'sha256'
+  }
+  Install-ChocolateyZipPackage @packageArgs
+  VM-Assert-Path $toolDir
+
+  # If $innerFolder is set to $true, after unzipping there should be only one folder
+  # GitHub ZIP files typically unzip to a single folder that contains the tools.
+  if ($innerFolder) {
```
Ana06, line 31:
```suggestion
```
[nitpick] I would have gotten rid of the comment, the `if` and the `$innerFolder` variable to simplify the code. `garbageman.vm` does this similarly. I think it should get removed in both, but it can be addressed in a different PR.

packages/windows-terminal.vm/tools/chocolateyuninstall.ps1
```diff
@@ -1,9 +1,11 @@
...
+$toolName = 'Windows Terminal'
 $category = 'Productivity Tools'
 $shortcutDir = Join-Path ${Env:TOOL_LIST_DIR} $category
 $shortcut = Join-Path $shortcutDir "$toolName.lnk"
+$toolDir = Join-Path ${Env:RAW_TOOLS_DIR} $toolName
 
-Remove-Item $shortcut -Force -ea 0 | Out-Null
\ No newline at end of file
+Remove-Item $shortcut -Force -ea 0 | Out-Null
+Remove-Item $toolDir
```
Ana06, line 11: why not using the helper here?
```
VM-Uninstall $toolName $category
```

If there is a reason we can't use the helper function, at least we should remove the `$toolDir` with:
```
Remove-Item $toolDir -Recurse -Force -ea 0 | Out-Null
```

Ana06 (changes requested):

Ana06 (commented):
I tested the package and it works perfectly. We should resolve the minor comment in the  uninstaller and we are good to go.

packages/windows-terminal.vm/tools/chocolateyinstall.ps1
```diff
@@ -0,0 +1,37 @@
...
+  Install-ChocolateyZipPackage @packageArgs
+  VM-Assert-Path $toolDir
+
+  # GitHub ZIP files typically unzip to a single folder that contains the tools.
+  $dirList = Get-ChildItem $toolDir -Directory
+  $toolDir = Join-Path $toolDir $dirList[0].Name -Resolve
+
+  $executablePath = Join-Path $toolDir $executableName -Resolve
+  VM-Install-Shortcut -toolName $toolName -category $category -executablePath $executablePath -runAsAdmin
+
```
Ana06, line 34: [nitpick]
```suggestion
```

Ana06 (approved):

Ana06 (approved):

PR #938, opened 2024-03-05
[pesieve] Use shortcut helper & console app

Ana06:
Change pe-sieve to use `VM-Install-Shortcut` to create the shortcut for consistency with other tools. Add the `consoleApp` option to keep the console open after executing pe-sieve.


emtuls (approved):
Looks good! Just tested. LGTM

Ana06:
Thanks for the testing @emtuls!

PR #939, opened 2024-03-05
[create_package_template] Improve template

Ana06:
Improve metapackage template in `create_package_template.py` by removing unnecessary and unused code. Note `VM-Install-Shortcut` already executes `VM-Assert-Path $shortcut` to verify that the shortcut has been created.

The CI would fail any package created with the previous template as the linter would complain about unused variables.


emtuls (approved):

emtuls:
Looks good! Thanks for updating that. 🙂

Issue #162, opened 2022-12-07
Package proposal: procdot.vm


seanthegeek:
### Package Name

procdot

### Tool Name

ProcDOT

### Package type

ZIP_EXE

### Tool's version number

1.22

### Category

Utilities

### Tool's authors

Christian Wojner

### Tool's description

Creates visual graphs from procmon output

### Download URL

https://procdot.com/download/procdot/binaries/procdot_1_22_57_windows.zip

### Download SHA256 Hash

927cd36dbb4dc0be94afb6021ca7f747dd3f17aad383583bc71aa6e36a762849

### Why is this tool a good addition?

FLARE VM does not have a tool to visualize procmon data. This would be a great addition.

Note: The ZIP has the password "procmon"

mr-tz:
The ZIP password seems to be a stopper at the moment. @MalwareMechanic may have an idea if it's worth handling this.

PR #949, opened 2024-03-07
:robot: Add windump.vm

vm-packages:
Automated PR from https://github.com/mandiant/VM-Packages/issues/948 :rocket:

PR #952, opened 2024-03-07
Add procdot.vm

emtuls:
This closes https://github.com/mandiant/VM-Packages/issues/162

This PR also requires https://github.com/mandiant/VM-Packages/pull/949 to be merged first as it is a dependency of this tool, as well as https://github.com/mandiant/VM-Packages/pull/981.

### DIFF
--- a/packages/procdot.vm/procdot.vm.nuspec
+++ b/packages/procdot.vm/procdot.vm.nuspec
@@ -1,0 +1,15 @@
+<?xml version="1.0" encoding="utf-8"?>
+<package xmlns="http://schemas.microsoft.com/packaging/2015/06/nuspec.xsd">
+  <metadata>
+    <id>procdot.vm</id>
+    <version>1.22.57</version>
+    <authors>Christian Wojner</authors>
+    <description>Creates visual graphs from procmon output</description>
+    <dependencies>
+      <dependency id="common.vm" />
+      <dependency id="7zip.vm" />
+      <dependency id="windump.vm" />
+      <dependency id="graphviz" />
+    </dependencies>
+  </metadata>
+</package>
--- a/packages/procdot.vm/tools/chocolateyinstall.ps1
+++ b/packages/procdot.vm/tools/chocolateyinstall.ps1
@@ -1,0 +1,64 @@
+$ErrorActionPreference = 'Stop'
+Import-Module vm.common -Force -DisableNameChecking
+
+$toolName = 'procdot'
+$category = 'Utilities'
+$toolDir = Join-Path ${Env:RAW_TOOLS_DIR} $toolName
+
+$zipUrl = 'https://procdot.com/download/procdot/binaries/procdot_1_22_57_windows.zip'
+$zipSha256 = '927cd36dbb4dc0be94afb6021ca7f747dd3f17aad383583bc71aa6e36a762849'
+$zipPassword = "procdot"
+$zipPath = Join-Path ${Env:RAW_TOOLS_DIR} "procdot_1_22_57_windows.zip"
+
+try {
+    # Remove files from previous zips for upgrade
+    VM-Remove-PreviousZipPackage ${Env:chocolateyPackageFolder}
+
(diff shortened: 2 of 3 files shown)


Ana06 (commented):
https://github.com/mandiant/VM-Packages/pull/949 is merged, please rebase :wink:

packages/procdot.vm/procdot.vm.nuspec
```diff
@@ -0,0 +1,15 @@
+<?xml version="1.0" encoding="utf-8"?>
+<package xmlns="http://schemas.microsoft.com/packaging/2015/06/nuspec.xsd">
+  <metadata>
+    <id>procdot.vm</id>
+    <version>1.22.57</version>
+    <authors>Christian Wojner</authors>
+    <description>Creates visual graphs from procmon output</description>
+    <dependencies>
+      <dependency id="common.vm" />
+      <dependency id="7zip-15-05.vm" />
```
Ana06, line 10: This is an issue if you install [7z-nsis.vm](https://github.com/mandiant/VM-Packages/tree/main/packages/7z-nsis.vm) as they would conflict. Options:
- to download and use a portable 7zip.
- To ask procdot author if he would mind releasing a version without password.
- Make 7z-nsis the default:
  - It may be enough to change it in the flare-vm config and ensure we are using it everywhere. 
  - another better alternative is to create a 7zip package which has as dependency the concrete zip package and use this in the flare-vm config. This makes it much easier to switch

packages/procdot.vm/tools/chocolateyinstall.ps1
```diff
@@ -0,0 +1,43 @@
...
+
+    # Download and unzip
+    $packageArgs = @{
+        packageName    = ${Env:ChocolateyPackageName}
+        unzipLocation  = $toolDir
+        url            = $zipUrl
+        checksum       = $zipSha256
+        checksumType   = 'sha256'
+        url64bit       = $zipUrl
+        checksum64     = $zipSha256
```
Ana06, line 24: this is only needed if there is 64 version
```suggestion
```

Ana06 (changes requested):

packages/procdot.vm/tools/chocolateyinstall.ps1
```diff
@@ -0,0 +1,43 @@
...
+        fileFullPath   = Join-Path ${Env:RAW_TOOLS_DIR} "procdot_1_22_57_windows.zip"
+    }
+    Get-ChocolateyWebFile @packageArgs
+    $zipPath = $packageArgs.fileFullPath
+    VM-Assert-Path $zipPath
+
+    # Unzip with a password
```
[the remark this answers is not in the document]
emtuls: I think I've only seen it once before, but it might be worth standardizing.

packages/procdot.vm/tools/chocolateyinstall.ps1
emtuls: There is both 64 bit and 32 bit versions in the same download. 🙂

packages/procdot.vm/procdot.vm.nuspec
emtuls: I created a `7zip` library package with `7zip-nsis` as the default: https://github.com/mandiant/VM-Packages/pull/981

That will need to be merged in before this PR will work.

packages/procdot.vm/tools/chocolateyinstall.ps1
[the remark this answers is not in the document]
Ana06: It may even cause a circular dependency... :see_no_evil:

Ana06 (approved):

emtuls:
Added some updates to include automatically adding a config file (if one doesn't exist), which would automatically set the paths to graphviz and WinDump, as well as skip some annoying initial pop-ups that occur when opening the tool.

Ana06 (approved):

PR #954, opened 2024-03-09
Fix broken memprocfs download link

citharus:
The download link of the 5.8.17 version does not exist anymore and has to be replaced with the newest release 5.9.3.
I updated the version in the `nuspec`, the `zipUrl` and `zipSha256`.
I tested it on my own VM and after the changes the package installed with no problems.

(I did not create an Issue because it was in the Daily Failures)

### DIFF
--- a/packages/memprocfs.vm/memprocfs.vm.nuspec
+++ b/packages/memprocfs.vm/memprocfs.vm.nuspec
@@ -2,7 +2,7 @@
 <package xmlns="http://schemas.microsoft.com/packaging/2015/06/nuspec.xsd">
   <metadata>
     <id>memprocfs.vm</id>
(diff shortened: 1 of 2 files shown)


naacbin:
Thanks for the PR @citharus ! However, I don't think we should use the URI you provide as the versions are archived quickly. So, the last version we can get to work is : https://github.com/ufrisk/MemProcFS/releases/download/v5_archive/MemProcFS_files_and_binaries_v5.9.4-win_x64-20240318.zip

Could you change you PR accordingly ? Also could you please squash your commits (in order to maintain a clean commit history) ?

naacbin:
LGTM, I've tested and it works well.

naacbin (approved):

Issue #749, opened 2023-11-22
Merge  `VM-Install-From-Zip` and `VM-Install-Raw-GitHub-Repo`


Ana06:
### Details

From https://github.com/mandiant/VM-Packages/pull/727>
> The code of the `VM-Install-From-Zip` and `VM-Install-Raw-GitHub-Repo` is very similar, which is not surprising as a GitHub-Raw install is the same as installing a Zip.

Any thoughts about merging these two functions?

Ana06:
It may make sense to address it together with https://github.com/mandiant/VM-Packages/issues/674

day1player:
What about configuring an alias? Then we can leave all packages as they are. I like that they are named separately since people might not know automatically that they are the same

Ana06:
A raw GH package is just a Zip that we download from GH. There is no difference at all with Zips we download from somewhere else. I think having both causes confusion. `VM-Install-From-Zip` has been lately improved and supports more options. I would like to get rid of `VM-Install-Raw-GitHub-Repo` both in the code and in the issues template and script to create packages. I think this will allow us to add some of the options to the issue template easily.

Proposal to progressively implement this (in different PR to be sent only after the previous steps have been merged):

- [ ] Merge `VM-Install-From-Zip` and `VM-Install-Raw-GitHub-Repo` into `VM-Install-From-Zip` but leave `VM-Install-Raw-GitHub-Repo` as a wrapper/alias that just calls `VM-Install-Raw-GitHub-Repo` so that we don't have to modify all the packages using `VM-Install-Raw-GitHub-Repo`
- [ ] Modify the script to generate packages, the issue template, and GH integration to only use `VM-Install-From-Zip`
- [ ] Update all the packages to use `VM-Install-From-Zip`

Ana06:
@emtuls would you like to work on this? You have worked on the shortcuts lately that are related. :wink:

Ana06:
This issue is not fixed. WE have simplified `VM-Install-Raw-GitHub-Repo` in #959, but the objective is to get rid of it completely. Steps needed to close this PR:
- [ ] Modify the script to generate packages, the issue template, and GH integration to only use `VM-Install-From-Zip`
- [ ] Update all the packages to use `VM-Install-From-Zip`

Ana06:
@emtuls you have forgotten one package: https://github.com/mandiant/VM-Packages/blob/main/packages/minidump.vm/tools/chocolateyinstall.ps1#L10

[...]

PR #959, opened 2024-03-23
Merge Install-Raw-Github-Repo and Install-From-Zip

emtuls:
This fixes https://github.com/mandiant/VM-Packages/issues/749

I updated a couple of commands that were not working in their current state.

Most packages needed `-innerFolder` added to them.

I updated the way we use `-withoutBinFile`. The current way we used it forced the tool to have `.exe`, which doesn't always make sense, especially when the flag is for 'without a binary file'. I take this to mean that the tool does -not- have any `.exe` associated with it (or we do not want the `.exe` associated with the shortcut), which was the case for many tools that are used in CommandoVM. By separating it into its own if statement, I think it brings more clarity into how we are using it and its intent.

All changed tools have been tested and work just as the original way they were set as, before any changes, other than ones previously broken, which are now fixed.

### DIFF
--- a/packages/adconnectdump.vm/adconnectdump.vm.nuspec
+++ b/packages/adconnectdump.vm/adconnectdump.vm.nuspec
@@ -2,7 +2,7 @@
 <package xmlns="http://schemas.microsoft.com/packaging/2015/06/nuspec.xsd">
   <metadata>
     <id>adconnectdump.vm</id>
-    <version>0.0.0.20230710</version>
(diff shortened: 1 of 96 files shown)


Ana06 (commented):
@emtuls can you rebase so that we ensure we test it with the latest changes? It touches a lot of files, so I want to ensure we don't break anything.

Ana06 (approved):

Issue #910, opened 2024-02-20
[arsenalimagemounter.vm] Install missing drivers


Ana06:
### Details

From https://github.com/mandiant/VM-Packages/pull/895:

I get the following message when opening Arsenal Image Mounter:
![Screenshot 2024-02-14 at 11 10 48](https://github.com/mandiant/VM-Packages/assets/16052290/db231da9-67b2-4e7c-a683-ebd57b098717)

Options to install the missing driver (proposed by @naacbin):
- [DriverFiles.zip](https://github.com/ArsenalRecon/Arsenal-Image-Mounter/blob/master/DriverSetup/DriverFiles.zip) file contains a folder DriverFiles\Win10\x64 with 6 drivers 5 have to be installed.
- Launch the application and click yes with autohotkey.

naacbin:
Close as #963 as been merged

PR #963, opened 2024-03-30
Resolve installation of drivers for Arsenal Image Mounter

naacbin:
Resolve installation of missing drivers for Arsernal Image Mounter (#910)


Ana06 (commented):
Thanks for the contribution @naacbin! I have installed this locally and it seems to work well! :partying_face: Is there any reason why we need the drivers install to be done in a separate package? If we are not planing to reuse this in another package, I would directly include the code in the arsenalimagemounter.vm package, maybe in a separate ps script if you think is too long.

naacbin:
I only separate into two packages to be more clear on what does which package. It should be fixed now.

scripts/test/lint.py
```diff
@@ -305,6 +305,7 @@ class UsesInvalidCategory(Lint):
         ".ollydumpex.vm",
         ".scyllahide.vm",
         "7zip-15-05.vm",
+        "arsenalimagemounter-drivers.vm",
```
Ana06, line 308:
```suggestion
```

Ana06 (commented):
Apart from the minor comment LGTM, @mandiant/commando-vm do you want to take a look? :smile:

day1player:
LGTM

Ana06 (approved):
Thanks @naacbin!

PR #964, opened 2024-03-31
Fix version of dependencies

naacbin:
Some dependencies have minor typo in their version or use non-existent version.
Typo:
- nasm
- openvpn
Not full version:
- netcat
- net-reactor-slayer
- vcbuildtools
Non existent / not verified version:
- vcredist140
- dotnet-6.0-sdk

### DIFF
--- a/packages/net-reactor-slayer.vm/net-reactor-slayer.vm.nuspec
+++ b/packages/net-reactor-slayer.vm/net-reactor-slayer.vm.nuspec
@@ -2,12 +2,12 @@
 <package xmlns="http://schemas.microsoft.com/packaging/2015/06/nuspec.xsd">
   <metadata>
     <id>net-reactor-slayer.vm</id>
-    <version>6.4.0.20230621</version>
+    <version>6.4.0.20240331</version>
     <authors>SychicBoy</authors>
     <description>NETReactorSlayer is an open source (GPLv3) deobfuscator and unpacker for Eziriz .NET Reactor.</description>
     <dependencies>
       <dependency id="common.vm" />
-      <dependency id="net-reactor-slayer" version="[6.4.0]" />
+      <dependency id="net-reactor-slayer" version="[6.4.0.0]" />
     </dependencies>
   </metadata>
 </package>
--- a/packages/common.vm/common.vm.nuspec
+++ b/packages/common.vm/common.vm.nuspec
@@ -2,7 +2,7 @@
 <package xmlns="http://schemas.microsoft.com/packaging/2015/06/nuspec.xsd">
   <metadata>
     <id>common.vm</id>
-    <version>0.0.0.20240305</version>
+    <version>0.0.0.20240321</version>
     <description>Common libraries for VM-packages</description>
     <authors>Mandiant</authors>
   </metadata>
--- a/packages/common.vm/tools/vm.common/vm.common.psm1
+++ b/packages/common.vm/tools/vm.common/vm.common.psm1
@@ -905,7 +905,7 @@ RAW_TOOLS_DIR: ${Env:RAW_TOOLS_DIR}
 }
 
 function VM-Remove-Appx-Package {
-# Function for removing Apps
+# Function for removing Appx Packages
     param(
         [Parameter(Mandatory=$true)]
         [ValidateNotNullOrEmpty()]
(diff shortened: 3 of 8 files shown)


Ana06 (commented):
This versions (or at least some of them) were updated by our bot and passed the test suite. How is this possible? :confused:  :detective:

naacbin:
Yes, in fact the packages works at it is. However, I'm currently playing with proxies to get a local cache of packages and I can't make the magic behind chocolatey easily. Even if the version is not correct, it will pick the one closest to the one specified. In any case, I think it's better to fix as it doesn't break anything and avoid some 404 (if you look on request made by choco). 😉

packages/sfextract.vm/sfextract.vm.nuspec
```diff
@@ -2,12 +2,12 @@
 <package xmlns="http://schemas.microsoft.com/packaging/2015/06/nuspec.xsd">
   <metadata>
     <id>sfextract.vm</id>
-    <version>2.1.0</version>
+    <version>2.1.0.20240331</version>
     <authors>Joery Droppers</authors>
     <description>command-line utility to extract files from single file bundles in .NET</description>
     <dependencies>
       <dependency id="common.vm" />
-      <dependency id="dotnet-6.0-sdk" version="[6.0.400, 6.1)" />
```
Ana06, line 10: We use this syntax on purpose so that this dependency is not updated by our bot. We want to install a single version of dotnet that all our tools use, but we don't really care which one exactly, as I do not expect dotnet minor versions to break our tools. By allowing several version, we decrease the change of a conflict (for example if a dotnet version was already installed and make our updates PR less noisy.

packages/vcredist140.vm/vcredist140.vm.nuspec
```diff
@@ -2,12 +2,12 @@
 <package xmlns="http://schemas.microsoft.com/packaging/2015/06/nuspec.xsd">
   <metadata>
     <id>vcredist140.vm</id>
-    <version>0.0.0.20231019</version>
+    <version>0.0.0.20230331</version>
     <description>Metapackage for Python 3 to ensure all packages use the same Python version.</description>
     <authors>Mandiant</authors>
     <dependencies>
       <dependency id="common.vm" />
-      <dependency id="vcredist140" version="[14.36.32532, 14.37)" />
```
Ana06, line 10: This is also in purpose to allow minor version updates similarly as with `dotnet-6.0-sdk`.

packages/nasm.vm/nasm.vm.nuspec
```diff
@@ -2,12 +2,12 @@
...
   <metadata>
     <id>nasm.vm</id>
-    <version>2.16.1.20240217</version>
+    <version>2.16.1.20240331</version>
     <authors>NASM Authors</authors>
     <description>Netwide Assembler</description>
     <dependencies>
       <dependency id="common.vm" />
-      <dependency id="nasm" version="[2.16.1.20221231]" />
+      <dependency id="nasm" version="[2.16.01.20221231]" />
```
Ana06, line 10: I think this is correct as it is, as it is the version chocolatey returns:
```
PS C:\Windows\system32 > choco find -er nasm
Chocolatey v2.2.2
nasm|2.16.1.20221231
FLARE-VM 04/03/2024 09:14:03
```
This is also what the update script uses. Is this causing any error? Can you please share more details?

packages/net-reactor-slayer.vm/net-reactor-slayer.vm.nuspec
```diff
@@ -2,12 +2,12 @@
...
   <metadata>
     <id>net-reactor-slayer.vm</id>
-    <version>6.4.0.20230621</version>
+    <version>6.4.0.20240331</version>
     <authors>SychicBoy</authors>
     <description>NETReactorSlayer is an open source (GPLv3) deobfuscator and unpacker for Eziriz .NET Reactor.</description>
     <dependencies>
       <dependency id="common.vm" />
-      <dependency id="net-reactor-slayer" version="[6.4.0]" />
+      <dependency id="net-reactor-slayer" version="[6.4.0.0]" />
```
Ana06, line 10:
```
PS C:\Windows\system32 > choco find -er net-reactor-slayer
Chocolatey v2.2.2
net-reactor-slayer|6.4.0
FLARE-VM 04/03/2024 09:15:17
```

Ana06 (commented):
@naacbin can you please share more information on the errors this is causing? These appear to be the versions returns by chocolatey. So If this is causing problems, we should likely report this upstream too.

packages/sfextract.vm/sfextract.vm.nuspec
naacbin: Yes, but shouldn't we say install to a known static version ? As for example `6.1` don't exist for dotnet-6-0.

packages/vcredist140.vm/vcredist140.vm.nuspec
naacbin: I might not understand correctly what it means. `[14.36.32532, 14.37)` -> from 14.36.32532 to 14.37 or from 14.36.32532 to 14.37.*.*

Ana06: It means `>= 14.36.32532 and < 14.37`. So basically, allow minor updates from current version as we don't expect libraries like this to break tools in minor updates.

packages/sfextract.vm/sfextract.vm.nuspec
Ana06: It means `>= 6.0.400 and < 6.1`. So basically, allow minor updates from current version as we don't expect libraries like this to break tools in minor updates. Using a concrete version has shown to be too noisy in the updates and it causes issues if you have tools requiring different minor versions. I'll check we have this documented (I think it is not).

packages/nasm.vm/nasm.vm.nuspec
Ana06: It seems to be a problem in the CLI, but not when installing the package:
image

The bot updated to 2.16.01.20221231 in https://github.com/mandiant/VM-Packages/commit/a3761cd153f64f990235bccd66cf78d245e7f6a3 (Jun 1, 2023) and to 2.16.1.20221231 in https://github.com/mandiant/VM-Packages/commit/cd949932263a7c2ce455d5d1bc6fdf71bae788c2 (Jun 14, 2023). It looks like this was introduced in Chocolatey in https://github.com/chocolatey/choco/pull/3175 and it is the intended new behaviour (normalized version). I think our code should use what is consistent with choco output (as otherwise the bot will change it back) and that the issue is that the CLI URLs do not supported normalized versions. I think you should report to chocolatey the fact that the URLs do not work with normalized versions, being inconsistent with command line output.

packages/sfextract.vm/sfextract.vm.nuspec
naacbin: Okay, I will remove this change from the PR

packages/vcredist140.vm/vcredist140.vm.nuspec
naacbin: I will remove this change too

packages/nasm.vm/nasm.vm.nuspec
naacbin: Thank for the answer, I will look to fix this on my side and I will remove this from the PR

naacbin:
Don't really know how I missed up this PR but since it won't be merged anyway, I'm closing it.

Issue #967, opened 2024-04-01
Tor-Browser Shortcut still on Desktop


emtuls:
### What's the problem?

The Tor-Browser package installs and attempts to remove the desktop shortcut, but it only removed it from the `Public` profile. It still remains in the Users profile.

### Steps to Reproduce

Install `tor-browser.vm`
View shortcut still on desktop.

### Environment

Environment Agnostic.

### Detected by test suite

No

PR #968, opened 2024-04-01
Remove Tor Desktop Shortcut

emtuls:
This fixes https://github.com/mandiant/VM-Packages/issues/967

### DIFF
--- a/packages/common.vm/common.vm.nuspec
+++ b/packages/common.vm/common.vm.nuspec
@@ -2,7 +2,7 @@
 <package xmlns="http://schemas.microsoft.com/packaging/2015/06/nuspec.xsd">
   <metadata>
     <id>common.vm</id>
-    <version>0.0.0.20240402</version>
+    <version>0.0.0.20240404</version>
     <description>Common libraries for VM-packages</description>
     <authors>Mandiant</authors>
   </metadata>
--- a/packages/tor-browser.vm/tools/chocolateyinstall.ps1
+++ b/packages/tor-browser.vm/tools/chocolateyinstall.ps1
@@ -12,11 +12,7 @@ try {
   Install-ChocolateyShortcut -shortcutFilePath $shortcut -targetPath $executablePath -RunAsAdmin
   VM-Assert-Path $shortcut
 
-  # Delete Desktop shortcut
-  $desktopShortcut = Join-Path ${Env:Public} "Desktop\$toolName.lnk"
-  if (Test-Path $desktopShortcut) {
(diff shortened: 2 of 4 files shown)


packages/tor-browser.vm/tools/chocolateyinstall.ps1
```diff
@@ -12,10 +12,14 @@ try {
...
-    Remove-Item $desktopShortcut -Force -ea 0
+  # Delete Desktop shortcuts
+  $publicDesktopShortcut = Join-Path ${Env:Public} "Desktop\$toolName.lnk"
+  if (Test-Path $publicDesktopShortcut) {
+    Remove-Item $publicDesktopShortcut -Force -ea 0
+  }
+  $userDesktopShortcut = Join-Path ${Env:UserProfile} "Desktop\$toolName.lnk"
+  if (Test-Path $userDesktopShortcut) {
+    Remove-Item $userDesktopShortcut -Force -ea 0
   }
```
Ana06, line 23: what about avoiding code duplication using for loop. Something like:

```suggestion
  # Delete Desktop shortcuts
  ForEach ($location in @(${Env:Public}, ${Env:UserProfile}) {
    $desktopShortcut = Join-Path $location "Desktop\$toolName.lnk"
    if (Test-Path $desktopShortcut) {
      Remove-Item $desktopShortcut -Force -ea 0
    }
```

Ana06 (commented):
Good catch! I wonder if this is a common enough case and we should make this a helper function.

emtuls:
It seems useful enough IMO. I do see a few packages we have where it could be useful and it simplifies future desktop shortcut removal from guessing the right way to do it.

packages/common.vm/common.vm.nuspec
```diff
@@ -2,7 +2,7 @@
 <package xmlns="http://schemas.microsoft.com/packaging/2015/06/nuspec.xsd">
   <metadata>
     <id>common.vm</id>
-    <version>0.0.0.20240321</version>
+    <version>0.0.0.20240402</version>
```
Ana06, line 5: You need to rebase as the version is the same as in https://github.com/mandiant/VM-Packages/pull/962/files#diff-e403a7afac0855f61eaed356192afac07ee3faf31d423f5e9afd8318ee50c188R5 recently merged

Ana06 (changes requested):
Apart from the version increase, LGTM, thanks!

Ana06 (approved):

Issue #977, opened 2024-04-05
Chrome & CyberChef have annoying popups before usable


emtuls:
### What's the problem?

When trying to use CyberChef, we hit a bunch of annoying pop-ups before we are able to actually use the tool. 
This includes:
- Sign in to chrome page on initial execution.
- Enhanced Ad Privacy pop up
- "What's New" tab

### Steps to Reproduce

Install `googlechrome.vm` or `cyberchef.vm` and try to use the tool.

### Environment

Environment agnostic

### Detected by test suite

No

Issue #929, opened 2024-02-29
[CI] Exclude microsoft-windows-terminal.vm from windows-2019 in the daily run


Ana06:
### What's the problem?

`microsoft-windows-terminal.vm` introduced in https://github.com/mandiant/VM-Packages/pull/876, does not install in windows-2019, reason why we decided to test it only in windows-2022. However in that PR we only excluded in [`ci.yml`](https://github.com/mandiant/VM-Packages/blob/main/.github/workflows/ci.yml), but not in [`daily.yml`](https://github.com/mandiant/VM-Packages/blob/main/.github/workflows/daily.yml), so the package is failing in the daily run.

Any ideas on what is the best way to implement this? I purpose we introduce a new parameter `excluded_packages` to the `test_install.ps1` script to make it easier to exclude from both actions instead of duplicating code in the github actions.

### Steps to Reproduce

Check the [daily run results](https://github.com/mandiant/VM-Packages).

### Environment

Daily run, windows-2019.

### Detected by test suite

Yes

Ana06:
Since 2024-03-06, windows-terminal does not fail to install in the daily run, @emtuls have you change something related to this and I do not remember? :detective:

emtuls:
Interesting...

I did make a change to use the `Github` package instead of `Chocolatey` in https://github.com/mandiant/VM-Packages/pull/936 which I believe is that same time frame. Previously, their code would determine if the OS was of the correct version or not, so I wonder if this is something not working (or possibly changed) in their code in Github release vs Chocolatey.

The Github Repo still states: `Windows Terminal requires Windows 10 2004 (build 19041) or later` (https://github.com/microsoft/terminal?tab=readme-ov-file#installing-and-running-windows-terminal), which the Github Actions `Windows-2019` Runner is not: `OS Version: 10.0.17763 Build 5576` (https://github.com/actions/runner-images/blob/main/images/windows/Windows2019-Readme.md)

Daily run seems to say it installs fine: https://github.com/mandiant/VM-Packages/actions/runs/8166465011/job/22325257278#step:3:8084
Whereas before it didn't: https://github.com/mandiant/VM-Packages/actions/runs/8150554446/job/22277009226#step:3:4860

emtuls:
If we were to add it to `test_install.ps1`, we could take advantage of the `exclude_tests` that already exists in it, but we will also need to have a parameter added to determine which OS version we are running a test on so that we can filter appropriately.

Ana06:
[...]

PR #978, opened 2024-04-05
Bypass annoying chrome pop-ups

emtuls:
This fixes https://github.com/mandiant/VM-Packages/issues/977

### DIFF
--- a/packages/googlechrome.vm/tools/chocolateyinstall.ps1
+++ b/packages/googlechrome.vm/tools/chocolateyinstall.ps1
@@ -20,7 +20,35 @@ try {
         fileType        = 'MSI'
         silentArgs      = "/quiet /norestart /l*v `"$($env:TEMP)\$($env:chocolateyPackageName).$($env:chocolateyPackageVersion).MsiInstall.log`""
     }
-    Install-ChocolateyInstallPackage @packageArgs
+    Install-ChocolateyInstallPackage @packageArgs -ErrorAction SilentlyContinue
+
+    VM-Assert-Path "${env:ProgramFiles}\Google\Chrome\Application\chrome.exe"
 } catch {
     VM-Write-Log-Exception $_
 }
+
+# Try to set configuration, but do not fail the package if it fails
+$ErrorActionPreference = 'Continue'
+
+# Expand the path to the Chrome User Data folder and create the "User Data" folder if it doesn't exist.
+$userDataPath = ${Env:LOCALAPPDATA} + "\Google\Chrome\User Data"
+New-Item -ItemType Directory -Path $userDataPath -Force -ea 0 | Out-Null
+
+# Create the empty file "First Run" to bypass the "Sign in to Chrome" initial pop-up.
+New-Item -Path $userDataPath -Name "First Run" -ItemType File -Force -ea 0 | Out-Null
+
+# Create the "Local State" file with data required to bypass the "What's New" pop-up tab.
+# Number must be higher than the current Chrome Version number and using 999 gives us some time before it fails.
+$contentOptions = @{
+    Path = Join-Path $userDataPath "Local State"
+    Value = '{"browser":{"last_whats_new_version":999}}'
+}
+Set-Content @contentOptions
+
+# Create the "Default" folder with a "Preferences" file with data required to bypass "Enhanced ad privacy" pop-up.
+New-Item -Path $userDataPath -Name "Default" -ItemType Directory -Force -ea 0 | Out-Null
+$contentOptions = @{
+    Path = Join-Path $userDataPath "Default\Preferences"
+    Value = "`{`"privacy_sandbox`":{`"m1`":{`"row_notice_acknowledged`":true}}`}"
+}
+Set-Content @contentOptions
--- a/packages/googlechrome.vm/googlechrome.vm.nuspec
+++ b/packages/googlechrome.vm/googlechrome.vm.nuspec
@@ -2,7 +2,7 @@
 <package xmlns="http://schemas.microsoft.com/packaging/2015/06/nuspec.xsd">
   <metadata>
     <id>googlechrome.vm</id>
-    <version>0.0.0.20240111</version>
+    <version>0.0.0.20240405</version>
     <authors>Google LLC.</authors>
     <description>Chrome is a popular web browser.</description>
     <dependencies>


emtuls:
This is a bit frustrating that it is erroring out, but I think I finally found the issue.

It looks like the `Windows-2019` and `Windows-2022` both have up to date Chrome installs, and because "chrome is already installed", we are receiving an MSI installer error. Not really sure how to mitigate this...
https://github.com/actions/runner-images/blob/main/images/windows/Windows2019-Readme.md#browsers-and-drivers
https://github.com/actions/runner-images/blob/main/images/windows/Windows2022-Readme.md#browsers-and-drivers

Might just need to accept that it will fail the install check, but if you test it locally, it should work.

emtuls:
Should we add this to `exclude_tests` in `test_install.ps1` as this is similar to how we are having issues with `windows-terminal`? 

From https://github.com/mandiant/VM-Packages/issues/929, we were looking to exclude it from `daily.yml`, but by putting it into an exclusion inside of `test_install.ps1`, and this file already has an `exclude_tests` that we can use?

Ana06:
This is a nice idea @emtuls!

> It looks like the Windows-2019 and Windows-2022 both have up to date Chrome installs, and because "chrome is already installed", we are receiving an MSI installer error.

I think the error is already in the test suite, see today's daily: https://github.com/mandiant/VM-Packages/actions/runs/8625498921/job/23642184219#step:3:1758

Ana06:
I am not able to reproduce the error locally. do you @emtuls? In any case, I think instead of ignoring testing, we should ignore the error and ensure that the Chrome executable is present.

emtuls:
Locally it shouldn't have an issue, since it is not already installed on your system, whereas it is already installed on the `windows-2019/2022` images that github actions run on.

I see. What would be the best way to ignore the error in the package when it installs? Trying not to change `$ErrorActionPreference = 'Stop'` to just be `Continue` instead, as I don't think that's the best route. 🤔

Ana06:
@emtuls 
> Locally it shouldn't have an issue, since it is not already installed on your system,

What I meant is that I tried to install google chrome in several ways locally and then installing the packages to try to replicate the GH setup and I was not able to reproduce the error.

I think using `-ErrorAction SilentlyContinue` and checking if chrome is installed afterwards is the way to go.

packages/googlechrome.vm/tools/chocolateyinstall.ps1
```diff
@@ -20,7 +20,33 @@ try {
         fileType        = 'MSI'
         silentArgs      = "/quiet /norestart /l*v `"$($env:TEMP)\$($env:chocolateyPackageName).$($env:chocolateyPackageVersion).MsiInstall.log`""
     }
-    Install-ChocolateyInstallPackage @packageArgs
+    Install-ChocolateyInstallPackage @packageArgs -ErrorAction SilentlyContinue
+
+    $chromePath = "${env:ProgramFiles}\Google\Chrome\Application\chrome.exe"
+    VM-Assert-Path $chromePath
```
Ana06, line 26: [nitpick] to make it more compact:
```suggestion
    Install-ChocolateyInstallPackage @packageArgs -ErrorAction SilentlyContinue
    VM-Assert-Path "%ProgramFiles%\Google\Chrome\Application\chrome.exe"
```

packages/googlechrome.vm/tools/chocolateyinstall.ps1
```diff
@@ -20,7 +20,33 @@ try {
...
+    # Create the "Default" folder with a "Preferences" file with data required to bypass "Enhanced ad privacy" pop-up.
+    New-Item -Path $userDataPath -Name "Default" -ItemType Directory -Force -ea 0 | Out-Null
+    $contentOptions = @{
+        Path = Join-Path $userDataPath "Default\Preferences"
+        Value = "`{`"privacy_sandbox`":{`"m1`":{`"row_notice_acknowledged`":true}}`}"
+    }
+    Set-Content @contentOptions
 } catch {
     VM-Write-Log-Exception $_
 }
```
Ana06, line 52: I think we shouldn't fail the package if setting the configuration fails:

```suggestion
} catch {
    VM-Write-Log-Exception $_
}

# Try to set configuration, but do not fail the package if it fails
$ErrorActionPreference = 'Continue'

# Expand the path to the Chrome User Data folder and create the "User Data" folder if it doesn't exist.
$userDataPath = ${Env:LOCALAPPDATA} + "\Google\Chrome\User Data"
New-Item -ItemType Directory -Path $userDataPath -Force -ea 0 | Out-Null

# Create the empty file "First Run" to bypass the "Sign in to Chrome" initial pop-up.
New-Item -Path $userDataPath -Name "First Run" -ItemType File -Force -ea 0 | Out-Null

# Create the "Local State" file with data required to bypass the "What's New" pop-up tab.
# Number must be higher than the current Chrome Version number and using 999 gives us some time before it fails.
$contentOptions = @{
    Path = Join-Path $userDataPath "Local State"
    Value = '{"browser":{"last_whats_new_version":999}}'
}
Set-Content @contentOptions

# Create the "Default" folder with a "Preferences" file with data required to bypass "Enhanced ad privacy" pop-up.
New-Item -Path $userDataPath -Name "Default" -ItemType Directory -Force -ea 0 | Out-Null
$contentOptions = @{
    Path = Join-Path $userDataPath "Default\Preferences"
    Value = "`{`"privacy_sandbox`":{`"m1`":{`"row_notice_acknowledged`":true}}`}"
}
Set-Content @contentOptions
```

emtuls:
> @emtuls
> 
> > Locally it shouldn't have an issue, since it is not already installed on your system,
> 
> What I meant is that I tried to install google chrome in several ways locally and then installing the packages to try to replicate the GH setup and I was not able to reproduce the error.
> 
> I think using `-ErrorAction SilentlyContinue` and checking if chrome is installed afterwards is the way to go.

Interesting...That was the conclusion I arrived at based on searching the error on the internet, though it could be many things, but it was something that someone mentioned, and I did notice that the Github Action VM's already had it installed. 😬

I'd have to play around with it some more to really try to narrow it down better. I wish I could access the Github Action VM to test from there.

packages/googlechrome.vm/tools/chocolateyinstall.ps1
emtuls: Interestingly enough, that seems to fail?
image

Ana06:
@emtuls

> I wish I could access the Github Action VM to test from there.

You can debug GH actions in several ways. I like to run them in my own fork with [logging enabled](https://docs.github.com/en/actions/monitoring-and-troubleshooting-workflows/enabling-debug-logging). Another alternatives are to use the [debugging with ssh action](https://github.com/marketplace/actions/debugging-with-ssh) or [Act](https://github.com/nektos/act).

Ana06 (approved):
Thanks @emtuls!

PR #980, opened 2024-04-10
Add Signature Verification to Regcool

emtuls:
This adds signature checks instead of `sha256` checks to `regcool.vm` so that it will not fail the `daily.yml` checks when an update occurs.

### DIFF
--- a/packages/regcool.vm/regcool.vm.nuspec
+++ b/packages/regcool.vm/regcool.vm.nuspec
@@ -2,7 +2,7 @@
 <package xmlns="http://schemas.microsoft.com/packaging/2015/06/nuspec.xsd">
   <metadata>
     <id>regcool.vm</id>
-    <version>2.0.0.20240408</version>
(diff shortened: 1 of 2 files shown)


packages/regcool.vm/regcool.vm.nuspec
```diff
@@ -2,7 +2,7 @@
 <package xmlns="http://schemas.microsoft.com/packaging/2015/06/nuspec.xsd">
   <metadata>
     <id>regcool.vm</id>
-    <version>2.0.0.20240408</version>
+    <version>2.0.0.20240410</version>
```
Ana06, line 5:
```suggestion
    <version>0.0.0.20240410</version>
```
As we don't know which tool version we are installing (potentially 3)

Ana06 (changes requested):
Thanks for this fix @emtuls!

Ana06 (approved):

PR #981, opened 2024-04-10
Add 7zip Library Wrapper

emtuls:
This ensures that we don't have conflicts with 7zip package installations.

I updated the package name for `7z-nsis` to be `7zip-nsis` instead and set it as the default dependency inside of `7zip.vm`

I also updated the packages that had a `7zip` dependency to ensure that they use `7zip.vm` instead and ensured that they installed properly and worked as intended.


Ana06 (approved):
LGTM

PR #982, opened 2024-04-10
De-ConsoleApp Windump

emtuls:
This will make `Windump` not a "consoleapp" anymore. It still functions normally, but gives the added benefit of allowing an easier way to find `Windump` when trying to add it to `Procdot` settings.


Ana06 (approved):
Thanks @emtuls!

PR #983, opened 2024-04-10
Remove Cmd Here from Registry

emtuls:
This removes the registry keys for Right Click -> `Cmd Here` that get registered automatically during installation of `map.vm`.

We have moved towards using `Windows-Terminal` for our default "Right Click Cmd Here" option, and this causes a conflict.

### DIFF
--- a/packages/map.vm/tools/chocolateyinstall.ps1
+++ b/packages/map.vm/tools/chocolateyinstall.ps1
@@ -28,3 +28,13 @@ try {
 } catch {
   VM-Write-Log-Exception $_
 }
+
+# Try to remove right click registry keys, but do not fail the package if it fails.
+$ErrorActionPreference = 'Continue'
+
+$cmdHereRegistryPath1 = "HKLM:\SOFTWARE\Classes\Directory\background\shell\ctx_cmd"
+$cmdHereRegistryPath2 = "HKLM:\SOFTWARE\Classes\Folder\shell\Cmd Here"
+
+ForEach ($regKey in @($cmdHereRegistryPath1, $cmdHereRegistryPath2)) {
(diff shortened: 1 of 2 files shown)


packages/map.vm/tools/chocolateyinstall.ps1
```diff
@@ -25,6 +25,16 @@ try {
   $shortcut = Join-Path $shortcutDir "$toolName.lnk"
   Install-ChocolateyShortcut -shortcutFilePath $shortcut -targetPath $toolDir
   VM-Assert-Path $shortcut
+
+  $cmdHereRegistryPath1 = "HKLM:\SOFTWARE\Classes\Directory\background\shell\ctx_cmd"
```
Ana06, line 29: I think I would move the catch before this code and add `$ErrorActionPreference = 'Continue'` to not fail the package if modifying the registry fails (then you can also simplify the code removing the Test-Path).

packages/map.vm/tools/chocolateyinstall.ps1
```diff
@@ -28,3 +28,14 @@ try {
...
   VM-Write-Log-Exception $_
 }
+
+# Try to remove right click registry keys, but do not fail the package if it fails.
+$ErrorActionPreference = 'Continue'
+
+$cmdHereRegistryPath1 = "HKLM:\SOFTWARE\Classes\Directory\background\shell\ctx_cmd"
+$cmdHereRegistryPath2 = "HKLM:\SOFTWARE\Classes\Folder\shell\Cmd Here"
+
+# Check if the key exists before deleting
```
Ana06, line 38:
```suggestion
```

Ana06 (approved):

Issue #688, opened 2023-10-12
Load windbg symbols


Ana06:
From https://github.com/mandiant/VM-Packages/issues/252#issuecomment-1476121411 (got lost as we only installed windbg when we closed that issue):

> Load `explorer.exe`, it will take some time (~30min), but will load many symbols so that you don't have to wait when working on other samples.
> We can use [.reload /f](https://learn.microsoft.com/en-us/windows-hardware/drivers/debugger/-reload--reload-module-).
> 
> Not sure if it should be part of the windbg.vm package or something like windbg.symbols.vm (so that we make downloading symbols optional).

I think I tend to prefer windbg.symbols.vm, other ideas?

binjo:
That amount of storage for symbols is probably not ideal for distribution. `pdblister` looks great for a package candidate, and gives the user options to download after installation.

schrodyn:
I would very much like to see symbols appearing in my FLARE-VM builds. Another alternative to downloading the symbols efficiently would be PDBReSym from the STrace project.

https://github.com/mandiant/STrace/blob/main/Rust/PDBReSym/src/main.rs#L437-L443

I've spoken with the author and depending on bandwidth and VM resources the tool completes downloading symbols in about 3 minutes =)

stevemk14ebr:
+1 to Conor's comment, I maintain PDBReSym publicly as part of my STrace project. Afaik it is the _fastest_ tool to download and cache all windows symbols. It's written in rust so is safe, and is something we should probably include as FLARE-VM as a Misc tool. Then as part of the FLARE-VM build we can have it run with the `cachesyms` flag which will automatically enumerate all C:\Windows binaries and cache their symbols in a format identically to how windbg and other tools do with the cache in C:\symbols by default, use `cachefolder` flag to change this default.

Ana06:
@mandiant/flare-vm which category should we add STrace to?

schrodyn:
FLARE-VM would just need to run `PDBReSym.exe cachesyms` and it will do the job. It will recurse through, by default, through `C:\Windows\System32\` and download all symbols and store them into `C:\symbols`. See help below.

The only issue that I can see might be the fact that PDBReSym isn't shipped as a compiled executable in the STrace releases. Maybe this is something Stephen can amend? Otherwise, the rust language toolchain would need to be installed in FLARE-VM and `cargo build` executed for `PDBReSym` and the executable placed into a suitable PATH.

[...]

PR #986, opened 2024-04-11
Add pdbresym.vm

Ana06:
Install PDBReSym and run `PDBReSym cachesyms` to iterate through `C:\Windows\System32` downloading all PDBs concurrently. The downloaded symbols are store into `C:\symbols`.

@stevemk14ebr and @schrodyn please take a look :wink: 

Closes https://github.com/mandiant/VM-Packages/issues/688

### DIFF
--- a/packages/pdbresym.vm/pdbresym.vm.nuspec
+++ b/packages/pdbresym.vm/pdbresym.vm.nuspec
@@ -1,0 +1,13 @@
+<?xml version="1.0" encoding="utf-8"?>
+<package xmlns="http://schemas.microsoft.com/packaging/2015/06/nuspec.xsd">
+  <metadata>
+    <id>pdbresym.vm</id>
+    <version>1.3.3</version>
+    <authors>Stephen Eckels</authors>
+    <description>Download PDBs</description>
+    <dependencies>
+      <dependency id="common.vm" version="0.0.0.20240411" />
+      <dependency id="vcredist140.vm" />
+    </dependencies>
+  </metadata>
+</package>
--- a/packages/pdbresym.vm/tools/chocolateyinstall.ps1
+++ b/packages/pdbresym.vm/tools/chocolateyinstall.ps1
@@ -1,0 +1,20 @@
+$ErrorActionPreference = 'Stop'
+Import-Module vm.common -Force -DisableNameChecking
+
+try {
+  $toolName = 'PDBReSym'
+  $category = 'Utilities'
+
+  $zipUrl = 'https://github.com/mandiant/STrace/releases/download/v1.3.3/PDBReSym.zip'
+  $zipSha256 = '803dfc0321581bc39001f050cdafe672e9e3247e96ffd42606fda3d641f0fd57'
+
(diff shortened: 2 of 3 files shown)


stevemk14ebr:
Looks good to me! We may want to consider the VM size as these pdbs can be big, and some people may also want the wow64 32 bit system binaries cached as well (can be done easily with flags to the tool)

Ana06:
@stevemk14ebr I agree. We are not planing to add it to the default config because of the increase in size, similar to Visual Studio.

schrodyn:
Nice! This covers exactly what I wanted. Thank you both =)

emtuls:
Should we also attempt to remove `C:\symbols` when uninstalling?

Also, should we have something that prints out something like `downloading all PDB's. This may take a few minutes...`? 
When I tested it, it was hard to tell if it was working or just taking a bit of time, and it doesn't natively spit out which PDB it's downloading or anything that indicates that it is running.

I think it also should include `<dependency id="common.vm" version="0.0.0.20240410" />` from https://github.com/mandiant/VM-Packages/pull/959. I had an older `common.vm` I think that was installed on my test setup and it failed until I updated my `common.vm` to latest.

Ana06:
@emtuls 
> Should we also attempt to remove C:\symbols when uninstalling?

I don't think we should. We don't know if the user downloaded other things in there, so I would say better we keep it simple. There is the following comment in the code: `Uninstalling only removes the tool not the downloaded symbols`



I'll add that a debugging message and the common dependency, thanks for the review @emtuls!

stevemk14ebr:
The tool does actually provide pretty extensive output. If it's not printing it's either due to the terminal you're using or a bug

emtuls:
Definitely seems to be some sort of terminal/powershell issue.

Running it on its own seems to work fine (even in powershell console), but running it in the script does not. I tried a bunch of various combinations to try to get it to display, but nothing worked (docs says it writes to Stderr: https://docs.rs/indicatif/latest/indicatif/). 

Oh well. :(

stevemk14ebr:
Looked into this for a second, stderr seems to be the preferred stream to write progress type information to, and also it's used for errors. This is so that progress type information can still get printed while stdout for final results can be easily readirected. 

I'm not sure which script you used to run it exactly, but perhaps the runner script should be adjusted to redirect output to the console for progress updates (maybe optionally) for some packages where the progress information is wanted to be displayed?

Ana06:
@stevemk14ebr 

> Looked into this for a second, stderr seems to be the preferred stream to write progress type information to, and also it's used for errors. This is so that progress type information can still get printed while stdout for final results can be easily readirected.

We are not really interested in progress information when installing a package as this would makes the log difficult to read. I think an info message as @emtuls suggested is enough in this case.

emtuls (approved):

stevemk14ebr:
@Ana06 I think this tool should always be present, but only executed in FULL configurations. As the installer file is written now it will execute the binary always, which makes it unnacceptable to add to the default config (so it will be missing from most FLAREVM builds). It being missing is bad, it should always be present, but not always run.

I'd suggest modifying the installer script to accept some kind of arguments that make it aware of the configration. And in the FULL configuration it will execute the binary. Is this possible?

PR #987, opened 2024-04-11
Update common.vm for all affected packages

emtuls:
From https://github.com/mandiant/VM-Packages/pull/959, there was an issue with packages not requiring the most up to date `common.vm` leading to many packages to fail in the `daily.yml` check.

There was also an issue with some older packages not performing `sha256 checks` due to it not being a mandatory requirement when made, which will need to be fully addressed later.

This should fix these issues.

### DIFF
--- a/packages/common.vm/tools/vm.common/vm.common.psm1
+++ b/packages/common.vm/tools/vm.common/vm.common.psm1
@@ -225,10 +225,11 @@ function VM-Install-Raw-GitHub-Repo {
     )
     try {
         if ($withoutBinFile) {
-            VM-Install-From-Zip -toolName $toolName -category $category -zipUrl $zipUrl -zipSha256 $zipSha256 -innerFolder $innerFolder -executableName $executableName -withoutBinFile -powershellCommand $powershellCommand
+            $toolDir = (VM-Install-From-Zip -toolName $toolName -category $category -zipUrl $zipUrl -zipSha256 $zipSha256 -innerFolder $innerFolder -executableName $executableName -withoutBinFile -powershellCommand $powershellCommand)[0]
         } else {
-            VM-Install-From-Zip -toolName $toolName -category $category -zipUrl $zipUrl -zipSha256 $zipSha256 -innerFolder $innerFolder -executableName $executableName -powershellCommand $powershellCommand
+            $toolDir = (VM-Install-From-Zip -toolName $toolName -category $category -zipUrl $zipUrl -zipSha256 $zipSha256 -innerFolder $innerFolder -executableName $executableName -powershellCommand $powershellCommand)[0]
         }
+        return $toolDir
     } catch {
         VM-Write-Log-Exception $_
     }
@@ -320,9 +321,10 @@ function VM-Install-Shortcut{
     }
 }
 
-# This functions returns $toolDir (outputed by Install-ChocolateyZipPackage) and $executablePath
+# This functions returns $toolDir and $executablePath
 function VM-Install-From-Zip {
     [CmdletBinding()]
+    [OutputType([System.Object[]])]
     Param
     (
         [Parameter(Mandatory=$true, Position=0)]
@@ -331,7 +333,7 @@ function VM-Install-From-Zip {
         [string] $category,
         [Parameter(Mandatory=$true, Position=2)]
         [string] $zipUrl,
-        [Parameter(Mandatory=$true, Position=3)]
+        [Parameter(Mandatory=$false, Position=3)]
         [string] $zipSha256,
         [Parameter(Mandatory=$false)]
         [string] $zipUrl_64,
@@ -397,7 +399,7 @@ function VM-Install-From-Zip {
             VM-Install-Shortcut -toolName $toolName -category $category -executablePath $executablePath -consoleApp $consoleApp -arguments $arguments
             Install-BinFile -Name $toolName -Path $executablePath
         }
-        return $executablePath
+        return ,@($toolDir, $executablePath)
     } catch {
         VM-Write-Log-Exception $_
     }
--- a/packages/adconnectdump.vm/adconnectdump.vm.nuspec
+++ b/packages/adconnectdump.vm/adconnectdump.vm.nuspec
@@ -2,11 +2,11 @@
 <package xmlns="http://schemas.microsoft.com/packaging/2015/06/nuspec.xsd">
   <metadata>
     <id>adconnectdump.vm</id>
-    <version>0.0.0.20240323</version>
+    <version>0.0.0.20240411</version>
     <authors>fox-it</authors>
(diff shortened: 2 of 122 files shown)


packages/common.vm/tools/vm.common/vm.common.psm1
```diff
@@ -289,19 +289,19 @@ function VM-Install-Shortcut{
...
         }
+
         Install-ChocolateyShortcut @shortcutArgs
 
     } else {
         $shortcutArgs = @{
             ShortcutFilePath = $shortcut
             TargetPath       = $executablePath
-            Arguments        = $arguments
-            WorkingDirectory = $executableDir
```
Ana06, line 299: why are we removing this two arguments here?

packages/common.vm/tools/vm.common/vm.common.psm1
```diff
@@ -369,34 +364,21 @@ function VM-Install-From-Zip {
             url64bit       = $zipUrl_64
             checksum64     = $zipSha256_64
         }
-        Install-ChocolateyZipPackage @packageArgs | Out-Null
```
Ana06, line 372: why are you changing this?

packages/common.vm/tools/vm.common/vm.common.psm1
```diff
@@ -639,27 +603,7 @@ function VM-Install-With-Installer {
         Install-ChocolateyInstallPackage @packageArgs
         VM-Assert-Path $executablePath
 
-        # if no icon path provided, set the shortcut icon to be the executable's icon. For MSI files, attempt to get executable icon from the installer first.
```
Ana06, line 642: why are you removing this code?

packages/common.vm/tools/vm.common/vm.common.psm1
```diff
@@ -533,22 +515,6 @@ function VM-Remove-Tool-Shortcut {
     Remove-Item $shortcut -Force -ea 0 | Out-Null
 }
 
-# Delete Desktop shortcuts
-function VM-Remove-Desktop-Shortcut {
```
Ana06, line 537: why are you removing this code?

Ana06 (commented):
It looks to me like you are removing code in common.vm by accident

emtuls:
That would be correct. Too many ctrl-z's 😬 

Thanks for catching that!

emtuls:
Recent fail was due to difference in expected return from `VM-Install-From-Zip`, which it shouldn't need to do anyway, so I updated the 2 packages that failed because of that to work in a more intended way.

Those packages were:
`file.vm`
`dnspyex.vm`

packages/common.vm/tools/vm.common/vm.common.psm1
```diff
@@ -221,7 +221,7 @@ function VM-Install-Raw-GitHub-Repo {
         # $powershellCommand = "Get-Content README.md"
         # $powershellCommand = "Import-Module module.ps1; Get-Help Main-Function"
         [Parameter(Mandatory=$false)]
-        [string] $powershellCommand
+        [string] $powershellCommand = ""
```
Ana06, line 224: why is this needed?

packages/dnspyex.vm/tools/chocolateyinstall.ps1
```diff
@@ -7,8 +7,9 @@ try {
 
   $zipUrl = "https://github.com/dnSpyEx/dnSpy/releases/download/v6.5.0/dnSpy-netframework.zip"
   $zipSha256 = "5962e3cca902e650c61050e2a879af58b78eec91288b7a7b77a7bc761424a0ed"
-  $toolDir = (VM-Install-From-Zip $toolName $category $zipUrl -zipSha256 $zipSha256)[0]
+  VM-Install-From-Zip $toolName $category $zipUrl -zipSha256 $zipSha256
```
Ana06, line 10: This function should return the `$toolDir` as documented in this comment in `$common`:
> # This functions returns $toolDir (outputed by Install-ChocolateyZipPackage) and $executablePath

If the previous PR got rid of that, we should bring it back instead of having to recalculate `$toolDir`. Other packages use this as well.

packages/file.vm/tools/chocolateyinstall.ps1
```diff
@@ -8,8 +8,10 @@ try {
     $zipUrl = "https://github.com/nscaife/file-windows/releases/download/20170108/file-windows-20170108.zip"
     $zipSha256 = "963147318f96d9345471e1a9a3943def4d95fcb3c1fe020e465ab910d0cda4a3"
 
-    $executablePath = (VM-Install-From-Zip $toolName $category $zipUrl -zipSha256 $zipSha256 -consoleApp $true)[-1]
-    $executableDir = Split-Path -Path $executablePath
+    VM-Install-From-Zip $toolName $category $zipUrl -zipSha256 $zipSha256 -consoleApp $true
```
Ana06, line 11: Same here. We should leave it as it was before and fix `VM-Install-From-Zip` to behave as expected.

packages/common.vm/tools/vm.common/vm.common.psm1
emtuls: I thought that was what led to: https://github.com/mandiant/VM-Packages/actions/runs/8651261154/job/23721615509#step:4:460

```
[adconnectdump.vm] vm.common.psm1 [+] ERROR : A parameter cannot be found that matches parameter name 'powershellCommand'.
At C:\ProgramData\_VM\vm.common\vm.common.psm1:228 char:185
+ ... utableName $executableName -withoutBinFile -powershellCommand $powers ...
+                                                ~~~~~~~~~~~~~~~~~~
ERROR: A parameter cannot be found that matches parameter name 'powershellCommand'.
The upgrade of adconnectdump.vm was NOT successful.
Error while running 'C:\ProgramData\chocolatey\lib\adconnectdump.vm\tools\chocolateyinstall.ps1'.
```

But now I'm realizing that it was just due to not having the `common.vm` version set properly in a dependency.


We do the same for `$executableName`: https://github.com/mandiant/VM-Packages/pull/987/files#diff-42c583682dc7cee57079eab399b94fbe5a60b1fcec5626a25a7092e197cb7308L217

packages/common.vm/tools/vm.common/vm.common.psm1
```diff
@@ -397,7 +399,7 @@ function VM-Install-From-Zip {
             VM-Install-Shortcut -toolName $toolName -category $category -executablePath $executablePath -consoleApp $consoleApp -arguments $arguments
             Install-BinFile -Name $toolName -Path $executablePath
         }
-        return $executablePath
+        return ,@($toolDir, $executablePath)
```
Ana06, line 402: Strange syntax with this comma. Maybe we should add a comment to clarify what this does, but this shouldn't block the PR as the packages are broken.

Ana06 (approved):

Issue #991, opened 2024-04-12
Split `pdbresym.vm` in `pdbresym.vm` and `pdbs.pdbresym.vm`


Ana06:
### Details

As per @stevemk14ebr feedback in https://github.com/mandiant/VM-Packages/pull/986#issuecomment-2051987881, I think we should split the downloading of the symbol in a separate package, something like `pdbresym.vm` (just the tool) and `pdbs.pdbresym.vm` (which requires the tool and download the symbols).

@mandiant/flare-vm opinions on the naming? Should we add the new `pdbresym.vm` to the default flare-vm config?

stevemk14ebr:
I like this idea, and I agree with Elliot this is the best way for now (per linked issue)

stevemk14ebr:
Thanks Ana!

PR #993, opened 2024-04-16
Split pdbresym.vm

Ana06:
Split pdbresym.vm into pdbresym.vm and dbs.pdbresym.vm to separate the download of the symbols (which increases the VM size considerable) and the too installation. @mandiant/flare-vm I think we should add pdbresym.vm to the default config, anything against it?

Closes https://github.com/mandiant/VM-Packages/issues/991


emtuls (approved):
Tested. Works well and good to go! Thank you for doing that, it's much better this way. :)

PR #995, opened 2024-04-16
[libraries.python3.vm] Fix dependency version

Ana06:
We pinned the common version when we meant to request a minimum version of it instead in the following PR:
https://github.com/mandiant/VM-Packages/pull/946


emtuls (approved):
Looks good! Good catch 🙂

PR #998, opened 2024-04-16
fix param bug with vm-uninstall-with-uninstaller

emtuls:
`VM-Uninstall-With-Uninstaller` used the position parameter of `1` more than once causing an issue with tools attempting to uninstall using it that did not directly pass in parameter names.


Ana06 (approved):
Thanks @emtuls!

Issue #999, opened 2024-04-19
Create package template for Node tools


Ana06:
### Details

The packages of Node tools installed using npm look very similar and a new template could be added to our create_package script and to our issue template. Current node tools examples:
- [malware-jail.vm](https://github.com/mandiant/VM-Packages/blob/38142cb7fac41a23da58c1006dae38a24b71ec5c/packages/malware-jail.vm/tools/chocolateyinstall.ps1)
- [pkg-unpacker.vm](https://github.com/mandiant/VM-Packages/blob/38142cb7fac41a23da58c1006dae38a24b71ec5c/packages/pkg-unpacker.vm/tools/chocolateyinstall.ps1)

~New tool we could add to test the template:~
- ~https://github.com/ben-sb/obfuscator-io-deobfuscator~

Ana06:
We should start by adding a function helper

Ana06:
Function helper added, still missing modifying the script, template and CI (GH action to create the package).

Ana06:
I have just realised that we do not need to download the zip to install [obfuscator-io-deobfuscator](https://github.com/ben-sb/obfuscator-io-deobfuscator), it is enough with running: `npm install -g obfuscator-io-deobfuscator`, tracked this in https://github.com/mandiant/VM-Packages/issues/1005

PR #1000, opened 2024-04-19
Add helper function for Node tools

Ana06:
Introduce a `VM-Install-Node-Tool-From-Zip` helper function to install Node tools installed with `npm` and normally downloaded from a GitHub repository as a ZIP with an inner folder.

First step to address https://github.com/mandiant/VM-Packages/issues/999

I do not understand why `PSAvoidUsingPositionalParameters` has started complaining now in the linter, as we use positional parameters in many functions:

```
packages\common.vm\tools\vm.common\vm.common.psm1
1 rule violation found.    Severity distribution:  Error = 0, Warning = 0, Information = 1

RuleName                            Severity     ScriptName Line  Message
--------                            --------     ---------- ----  -------
PSAvoidUsingPositionalParameters    Information  vm.common. 393   Cmdlet 'VM-Install-From-Zip' has positional
                                                 psm1             parameter. Please use named parameters instead of
                                                                  positional parameters when calling a command.
```

Exclude this violation as it is of `Information` severity and we wouldneed to make a lot of changes to fix it.

### DIFF
--- a/packages/common.vm/common.vm.nuspec
+++ b/packages/common.vm/common.vm.nuspec
@@ -2,7 +2,7 @@
 <package xmlns="http://schemas.microsoft.com/packaging/2015/06/nuspec.xsd">
   <metadata>
     <id>common.vm</id>
-    <version>0.0.0.20240416</version>
+    <version>0.0.0.20240419</version>
     <description>Common libraries for VM-packages</description>
     <authors>Mandiant</authors>
   </metadata>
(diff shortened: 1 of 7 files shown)


emtuls (changes requested):
I think we are missing the `Try-Catch` inside of `VM-Install-Node-Tool-From-Zip`

emtuls (approved):
Good to go! 🙂

Issue #992, opened 2024-04-15
configure windbg to find the symbols downloaded by pdbresym.vm


Ana06:
### Details

Is it possible to configure windbg to find the symbols downloaded in `pdbs.pdbresym.vm` (`File -> Settings -> Debugging settings -> Symbol path`)? It may be enought with setting the Symbol path in the windbg package to `C:\symbols`. @mandiant/flare-vm @stevemk14ebr opinions?

stevemk14ebr:
Yes it's very easy set environment variable `_NT_SYMBOL_PATH` to `srv*c:\symbols*https://msdl.microsoft.com/download/symbols`

PDBReSym carefully reimplements the cache structure windbg and other tools rely on.

Ana06:
Thanks for the info @stevemk14ebr! I have set the variable running the following command in PowerShell:
```
[System.Environment]::SetEnvironmentVariable("_NT_SYMBOL_PATH", "srv*c:\symbols*https://msdl.microsoft.com/download/symbols", "Machine")
```

But if after that I open Windbg, it seems like the settings are not being updated. Is this enough?
![image](https://github.com/mandiant/VM-Packages/assets/16052290/ab9e1164-1819-4369-8ca2-5808114fb28f)

stevemk14ebr:
Well that's lame, doesn't seem like it reads the environment variable unless I'm missing something. Quick RE shows it writes a file `%LOCALAPPDATA%\DbgX.xml` with the settings. We could pre-populate this I guess. My file after setting the sym path on fresh load is:

```
<?xml version="1.0" encoding="utf-8"?>
<Settings>
  <Setting Name="EulaState" Value="accepted" />
  <Setting Name="MainWindow.WindowPlacement" Value="44;0;1;-1;-1;-1;-1;26;26;926;726" />
  <XmlSetting Name="EngineSettings">
    <Settings Version="1">
      <Namespace Name="Sources">
        <Setting Name="SourceSearchPath" Type="VT_BSTR" Value="srv*" />
      </Namespace>
      <Namespace Name="Symbols">
        <Setting Name="Sympath" Type="VT_BSTR" Value="srv*c:\symbols" />
      </Namespace>
      <Namespace Name="Extensions">
        <Namespace Name="DisabledExtensions"></Namespace>
      </Namespace>
    </Settings>
  </XmlSetting>
  <XmlSetting Name="WindowPreferences">
[...]

PR #1003, opened 2024-04-22
[pdbs.pdbresym.vm] Set _NT_SYMBOL_PATH

Ana06:
Set _NT_SYMBOL_PATH to include `C:\symbols`, where we have downloaded the symbols using PDBReSym.

Closes https://github.com/mandiant/VM-Packages/issues/992


emtuls (approved):
Looks good! 🙂

Issue #1004, opened 2024-04-22
regcool uses URL without the version


Ana06:
### Details

regcool uses a URL without the version and breaks with updates. See the last daily run.

emtuls:
For some reason, it seems that when we don't pass in a checksum, chocolately automatically attempts to validate the checksum of the newly downloaded file against the checksum of the previous package stored in the cache in our sources.
![image](https://github.com/mandiant/VM-Packages/assets/35854880/e6a9bd11-95c4-4492-855c-f71ee80dc5d6)


Because we don't want to force all packages to ignore checksums, we could attempt to add `$env:ChocolateyIgnoreChecksums = $true` to the top of the code (inside the `try-catch`) and this will act as if we set `--ignore-checksums` for the specific package, and allows it to work as intended.

Ana06:
@emtuls did you want to add this package to the default config? :thinking:

PR #1007, opened 2024-04-23
Ignore Checksum for Signature Validated Files

emtuls:
This attempts to fix https://github.com/mandiant/VM-Packages/issues/1004

Issue and explanation copied here:
For some reason, it seems that when we don't pass in a checksum for certain choco functions, chocolatey automatically attempts to validate the checksum of the newly downloaded file against the checksum of the previous package stored in the cache in our sources.
image

Because we don't want to force all packages to ignore checksums, we can add `$env:ChocolateyIgnoreChecksums = $true` to the top of the code (inside the `try-catch`) and this will act as if we set `--ignore-checksums` for this specific package, allowing it to install correctly.

EDIT: This seems to not get pushed through unless we remove the current choco package in our myget cache.

### DIFF
--- a/packages/common.vm/common.vm.nuspec
+++ b/packages/common.vm/common.vm.nuspec
@@ -2,7 +2,7 @@
 <package xmlns="http://schemas.microsoft.com/packaging/2015/06/nuspec.xsd">
   <metadata>
     <id>common.vm</id>
-    <version>0.0.0.20240419</version>
+    <version>0.0.0.20240424</version>
     <description>Common libraries for VM-packages</description>
     <authors>Mandiant</authors>
   </metadata>
--- a/packages/common.vm/tools/vm.common/vm.common.psm1
+++ b/packages/common.vm/tools/vm.common/vm.common.psm1
@@ -128,20 +128,99 @@ function VM-Assert-Path {
     }
 }
 
-# Raise an exception if the Signature of $file_path is invalid
+# Raise an exception if the Signature of $filePath is invalid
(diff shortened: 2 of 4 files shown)


emtuls:
Hmmm. It seems that our `test_upload` uses the `test_install.ps1` script which only does a `upgrade` rather than a forced `install`, which leads to the package not actually attempting to install the updated packed?

The top powershell is the old package installing with failure, the bottom is attempting to install the new version with `test_install.ps1`, which also leads to a failure.
image

But, if I pack it manually and for an install with `choco install "regcool.vm" --source="C:\Users\Emtuls\FlareVM\VM-Packages\packages\regcool.vm" -y --force`, then it works just fine:
image

Ana06:
@emtuls 

> For some reason, it seems that when we don't pass in a checksum for certain choco functions, chocolatey automatically attempts to validate the checksum of the newly downloaded file against the checksum of the previous package stored in the cache in our sources.

I have checked the Chocolatey code and this is not the case. As explained in https://github.com/mandiant/VM-Packages/issues/1004#issuecomment-2124413300, the problem with `regcool.vm` is that we decreased the package version without deleting the older versions from MyGet. The issue is fixed now.

But I do like some of the ideas in this PR, like checking or at least rending who is signing the package. I think we should discuss what part of this implementation we want to keep and adapt the other packages using signature validated files as well.

Ana06:
https://github.com/mandiant/VM-Packages/pull/1157 improves how we verify the tools using signatures. So we can close this PR.

Issue #974, opened 2024-04-03
VM-Clean-Up - fix sort Desktop icons


Ana06:
### Details

I have encountered two issues with `VM-Clean-Up`:
- When sorting the Desktop icons by type, if they are already sorted, they are sorted by type in reverse order. This causes the VM to look different depending how the icons were sorted and it is a problem to distribute builds that look the same.
- When running `VM-Clean-Up` from a PowerShell tab in the Windows Terminal, the icons are not sorted. Does maybe increasing the time we wait fixes this? Do we need to document to run this from PowerShell and not from the Windows Terminal?

@emtuls ideas?

emtuls:
After many hours of testing various things, I finally found a solution that works for both standard Powershell and Windows-Terminal Powershell. 😊

Regular Powershell is a little bit easier, since when we perform `(New-Object -ComObject Shell.Application).toggleDesktop();`, it automatically brings the focus to the desktop, whereas with Windows-Terminal, it doesn't.

The general sorting solution ended up just being to sort by Name first, then by Item Type. Testing various ways does seem to work how we would like.

The only maybe more ideal way would be to have `Tools` folder just under the `Recycle Bin`, but this was the best I could get.

PR #1009, opened 2024-04-24
Force same icon layout when sorting

emtuls:
This fixes https://github.com/mandiant/VM-Packages/issues/974

I also moved the icon sorting function call outside of the `VM-Remove-DesktopFiles` call since it felt a little out of place as it's not removing desktop files, rather, it's sorting icons on the desktop, and it also made it more clear what individual steps were being performed when calling `VM-Clean-Up`

### DIFF
--- a/packages/common.vm/tools/vm.common/vm.common.psm1
+++ b/packages/common.vm/tools/vm.common/vm.common.psm1
@@ -1354,12 +1354,23 @@ public class Shell {
 }
 
 # Sort Desktop icons by item type using WScript.Shell to replicate the manual steps
+# This is written in a way to that ensures things will be sorted properly for multiple possible states.
 function VM-Sort-Desktop-Icons {
     VM-Write-Log "INFO" "Sorting Desktop icons"
+    # toggleDesktop sets the focus to the desktop for regular Powershell, but not fully for Windows-Terminal Powershell.
     (New-Object -ComObject Shell.Application).toggleDesktop();
     Start-Sleep -Milliseconds 100;
+    # Start Windows Run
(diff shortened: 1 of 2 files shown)


packages/common.vm/tools/vm.common/vm.common.psm1
```diff
@@ -1358,8 +1358,18 @@ function VM-Sort-Desktop-Icons {
     VM-Write-Log "INFO" "Sorting Desktop icons"
     (New-Object -ComObject Shell.Application).toggleDesktop();
     Start-Sleep -Milliseconds 100;
+    # Start Windows Run
```
Ana06, line 1361: why do we need to start Windows Run?

packages/common.vm/tools/vm.common/vm.common.psm1
```diff
@@ -1358,8 +1358,18 @@ function VM-Sort-Desktop-Icons {
...
+    $shell = New-Object -ComObject Shell.Application
+    $shell.FileRun()
```
Ana06, line 1363: For consistency:
```suggestion
    $shell = New-Object -ComObject Shell.Application;
    $shell.FileRun();
```

Ana06 (commented):
Thanks for doing some extra research on this @emtuls!

packages/common.vm/tools/vm.common/vm.common.psm1
emtuls: With that pop-up, I'm about to use `Alt-F4` to simply close it, which will put the focus onto the desktop.  There may be other ways, but I couldn't find a better solution. 😅 

It may also work with other window pop-ups, but this was a simple one to start. Other ones I tried needed the Windows key + <something> to open and that take as easy to emulate with the SendKeys functionality.

Issue #1001, opened 2024-04-19
py shortcut icon still broken


Ana06:
### Details

I have just installed ida.diaphora.vm 3.2.0 and the  icon of the .py file looks broken:

![image](https://github.com/mandiant/VM-Packages/assets/16052290/3474b5ac-45ea-41ac-8089-f73549e920d3)

@emtuls hadn't this been addressed in https://github.com/mandiant/VM-Packages/issues/675?

We are planing to change how diaphora is installed (without shortcut): https://github.com/mandiant/VM-Packages/issues/994, so this issue is about ensuring that the .py shortcut look nice and not about fixing diaphora specifically. But I noticed the issue while testing diaphora.

emtuls:
Strange...I just tried it on a fresh setup and it looks fine for me? 🤔
![image](https://github.com/mandiant/VM-Packages/assets/35854880/b37a50ce-4aaf-4b1b-8e07-b1d68f590f30)

emtuls:
Ah, so I was mistaken. It appeared to work because of my vscode I had installed.

So, this issue is because we opted to go the route of requiring tools that were python based to use python as the `executablePath` (which will make python the `Target` for the shortcut), which will in turn make python the icon due to `iconLocation` being set to `executablePath`: https://github.com/mandiant/VM-Packages/blob/main/packages/common.vm/tools/vm.common/vm.common.psm1#L228

The downside that this had was that we don't default all Python files to use the Python icon as a shortcut if they are something like ida.diaphora, since it's just a python plugin and we weren't planning to execute it directly with Python.

The other downside is that we have to call `VM-Install-Shortcut` directly, like such: https://github.com/mandiant/VM-Packages/blob/main/packages/didier-stevens-beta.vm/tools/chocolateyinstall.ps1#L22-L25

An alternative we could do, that I think I may have suggested somewhere, would be to add inside of `VM-Install-From-Zip`, a check for if the `$executableName` has an extension of `.py`, and then we could do:
```
if ([System.IO.Path]::GetExtension($executableName) -eq ".py") {
    $executablePath = (Get-Command python).Source
    $filePath = Join-Path $toolDir "$toolName.py"
    $arguments = $filePath + $arguments
    $consoleApp = $true
    VM-Install-Shortcut -toolName $toolName -category $category -executablePath $executablePath -consoleApp $consoleApp -arguments $arguments
}
```

I've just tested that and it would work, if you want to go that route.

Ana06:
[...]

PR #1011, opened 2024-04-25
Fix Python Shortcuts and ida.diaphora

emtuls:
This fixes https://github.com/mandiant/VM-Packages/issues/1001


Ana06 (changes requested):
See https://github.com/mandiant/VM-Packages/issues/1001#issuecomment-2076852323

PR #1012, opened 2024-04-25
Make FakeNet not close on ctrl+c

emtuls:
This makes it so that the FakeNet command prompt that it opens up in stays open after you hit `ctrl+c` to stop FakeNet.


Ana06 (approved):
Thanks @emtuls

Issue #695, opened 2023-10-17
Install flare-capa in ida.plugin.capa.vm instead of libraries.python3.vm


Ana06:
We should install flare-capa in ida.plugin.capa.vm instead of libraries.python3.vm.

Discussed with @MalwareMechanic 

Related https://github.com/mandiant/VM-Packages/issues/673

PR #1014, opened 2024-04-25
Install flare-capa in ida.plugin.capa.vm

Ana06:
Install the Python library `flare-capa` in ida.plugin.capa.vm instead of `libraries.python3.vm`. This is an easy change thanks to the introduction of the `VM-Pip-Install` function in https://github.com/mandiant/VM-Packages/pull/935.

Closes https://github.com/mandiant/VM-Packages/issues/695


emtuls (approved):
Looks good!

Issue #1015, opened 2024-04-25
Create Function for File Association for "Open With"


emtuls:
### Details
After diving into something that @binjo brought up, it appears to me that currently, we do not have any way to add a default file/executable to open specific file types.

For example, while `7zip` associates itself with certain file types for a right click option (https://github.com/mandiant/VM-Packages/blob/main/packages/7zip-nsis.vm/tools/chocolateyinstall.ps1#L38C107-L38C124), it does not make itself the default executable to open those file types, leading to the file types having either no associated icon for the file or other default icons.

It would be nice for file types that we want to associate 7zip with, like those listed https://github.com/mandiant/VM-Packages/blob/main/packages/7zip-nsis.vm/tools/chocolateyinstall.ps1#L36, to also have 7zip as the icon for those files as well.

This would require mimicing how 7zip does this action when you manually click "Open With" and select `7zFM`.

To do this, we will need to add `HKEY_CURRENT_USER\SOFTWARE\Classes\7z_auto_file\shell\open\command` to have `(Default): "C:\Program Files\7-Zip\7zFM.exe" "%1"`

This provides a central `7zip` "Open With" that we can use for other file types to use. For each of those file types, we would just need to do something like: `HKEY_CURRENT_USER\SOFTWARE\Classes\.7z` and add `(Default): 7z_auto_file`, which is an example of just the `.7z` file type.

I propose we make a function that can do this for us inside of `common.vm` and call it something like `VM-Set-Open-With-Association`.

PR #1016, opened 2024-04-25
Add Open With File Association

emtuls:
This addresses https://github.com/mandiant/VM-Packages/issues/1015

It adds the ability to give a file extension a default executable to open it with. I also updated the 7zip packages to use the new function.

If you already have a `.7z` file (or other associated file type), this may require logging out and back in or performing another form of File Explorer refresh, such as `Stop-Process -Name explorer; Start-Process explorer`.


packages/common.vm/tools/vm.common/vm.common.psm1
```diff
@@ -886,6 +886,82 @@ function VM-Remove-From-Right-Click-Menu {
...
+            New-Item -Path $commandKey -Force
+            New-ItemProperty -Path $commandKey -Name '(Default)' -Value "`"$executablePath`" `"%1`""
+
+            # Create/update the file extension key
+            $extKey = "${hive}\Software\Classes\$extension"
+            New-Item -Path $extKey -Force
+            New-ItemProperty -Path $extKey -Name '(Default)' -Value "${exeName}_auto_file"
+
+            # Add to OpenWithProgids for visibility in "Open with..." menu
+            if ((Get-ItemProperty -Path $extKey -Name 'OpenWithProgids' -ErrorAction Ignore)) {
```
Ana06, line 914:
```suggestion
            if (Get-ItemProperty -Path $extKey -Name 'OpenWithProgids' -ErrorAction Ignore) {
```

packages/googlechrome.vm/tools/chocolateyinstall.ps1
```diff
@@ -22,14 +22,21 @@ try {
...
 }
 
 # Try to set configuration, but do not fail the package if it fails
 $ErrorActionPreference = 'Continue'
 
+# Delete Desktop shortcut created during installation
+$desktopShortcut = Join-Path ${Env:Public} "Desktop\Google Chrome.lnk"
+if (Test-Path $desktopShortcut) {
+    Remove-Item $desktopShortcut -Force -ea 0
+}
```
Ana06, line 38:
```suggestion
VM-Remove-Desktop-Shortcut "Google Chrome"
```

Ana06 (commented):
It works nicely! I have just 2 minor comments :smile:

Ana06:
@emtuls I think you don't have time in the next days to address the minor changes I pointed out. I will address the feedback myself so that we can merge this PR. :wink:

Ana06:
I have address the feedback and added more details to the commit messages. I think it is good to go!

Ana06 (approved):
I have address the feedback and added more details to the commit messages. I think it is good to go!

Ana06:
@emtuls could we remove now the Chrome dependency in Cyberchef and the code to open it with chrome and set the icon? I think with this PR removing that code would keep the current behavior if both cyberchef.vm and chrome.vm are installed and remove the chrome dependency would allow to have a different browser as default (and to open cyberchef)

emtuls:
> @emtuls could we remove now the Chrome dependency in Cyberchef and the code to open it with chrome and set the icon? I think with this PR removing that code would keep the current behavior if both cyberchef.vm and googlechrome.vm are installed and remove the chrome dependency would allow to have a different browser as default (and to open cyberchef) if only cyberchef.vm is installed.

I'll look into this! I had a thought about it earlier, but I got tied up in some things, so I'll have to give it some thought again soon. :)

emtuls:
Noted issues here: https://github.com/mandiant/VM-Packages/issues/1021#issuecomment-2127748033

We still run into the issue where a popup displays asking if we want to use Chrome (other other browser) as the default browser even after the association. The icon and expected executable for the association changes, but it's not enough to bypass the popup unless we use something like `SetDefaultBrowser`: https://community.chocolatey.org/packages/setdefaultbrowser

Issue #996, opened 2024-04-16
Add issue template for IDA plugins && support automate package creation


Ana06:
### Details

Introduce a new helper for IDA plugins and an issue template so that we can create packages that install ida plugins automatically.

Split https://github.com/mandiant/VM-Packages/issues/719 using the new template

stevemk14ebr:
You can use the IDAUSR folder for partial configs and plugin folder merging without needing to put them in the IDA directory. Works for python and binary plugins

Ana06:
> You can use the IDAUSR https://hex-rays.com/blog/igors-tip-of-the-week-33-idas folder for partial configs and plugin folder merging without needing to put them in the IDA directory

This is how we are doing it already (for the only plugin we have: capa plugin) :wink:  `IDAUSR` is shared between IDA versions.

Ana06:
I am adding a helper function in https://github.com/mandiant/VM-Packages/pull/1013 that downloads a single file to the plugins directory. This seems to be how most plugins distributed as a DLL are installed. The following Python plugins can also be installed in this way:
- https://github.com/hasherezade/ida_ifl
- https://github.com/airbus-cert/comida

I have checked several IDA plugins and the other common case seems to be copying a folder to the plugins directory (normally downloaded from GH in a ZIP). Plugins that can be installed in this way:
- https://github.com/gaasedelen/tenet (plugins folder)
- https://github.com/gaasedelen/lighthouse (plugins folder)
- https://github.com/danigargu/deREferencing (root folder, we may want to ask the maintainer for a subfolder that does not include things like the license and the readme and we can directly copy)


I suggest extending the helper to support this second case after #1013 has been merged and create a new issue template for automation for both cases and that allow to provide installation details for the rest. After that we can trigger automation to generate the packages mentioned above!

Ana06:
https://github.com/gaasedelen/tenet doesn't work with Python 3.10: https://github.com/gaasedelen/tenet/issues/15#issuecomment-2084743150

PR #1020, opened 2024-04-30
Add ZIP support in VM-Install-IDA-Plugin & Add ida.plugin.dereferencing.vm

Ana06:
`VM-Install-IDA-Plugin` now in addition of supporting downloading an IDA plugin file to the plugins directory, it also supports ZIPs containing a plugin (and supporting files/directories). For ZIPs, we check if there is an inner folder (this is the case for GH ZIPs) and if there is a directory called 'plugins'.  We copy all files in this directory with the exception of the `README` and the `LICENSE` file (often present in GH repos). The copied files must include `$pluginName`, which is used for uninstallation.

Use the IDA plugin template in `create_package_template.py` to generate `ida.plugin.dereferencing.vm` thank to the new ZIP support:

```
python3 scripts/utils/create_package_template.py --type IDA_PLUGIN --pkg_name "ida.plugin.dereferencing" --version "0.0.0.20240430" --authors "danigargu" --description "IDA Pro plugin that implements new registers and stack views." --tool_name "dereferencing.py" --target_url "https://github.com/danigargu/deREferencing/archive/c5c606a9e70bff48214ce5286a37b15752fd8d1b.zip" --target_hash "3ddec5c7569bc53883c5feaeb36d1145e2dde1c67491d14929af05938870dc1e"
```

Partially addresses https://github.com/mandiant/VM-Packages/issues/996

I would like to add this plugin to the default FLARE-VM installation, anything against it @mandiant/flare-vm ?

### DIFF
--- a/packages/common.vm/common.vm.nuspec
+++ b/packages/common.vm/common.vm.nuspec
@@ -2,7 +2,7 @@
 <package xmlns="http://schemas.microsoft.com/packaging/2015/06/nuspec.xsd">
   <metadata>
     <id>common.vm</id>
-    <version>0.0.0.20240425</version>
+    <version>0.0.0.20240429</version>
     <description>Common libraries for VM-packages</description>
     <authors>Mandiant</authors>
   </metadata>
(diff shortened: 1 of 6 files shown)


williballenthin:
this is awesome!

PR #1023, opened 2024-05-07
Update package to fix daily failure

naacbin:
- dcode.vm
- dotnet-6.vm
- recmd.vm
- rla.vm
- timeline_explorer.vm

### DIFF
--- a/packages/dotnet-6.vm/dotnet-6.vm.nuspec
+++ b/packages/dotnet-6.vm/dotnet-6.vm.nuspec
@@ -2,13 +2,13 @@
 <package xmlns="http://schemas.microsoft.com/packaging/2015/06/nuspec.xsd">
   <metadata>
     <id>dotnet-6.vm</id>
-    <version>0.0.0.20231224</version>
+    <version>0.0.0.20240507</version>
     <authors>Microsoft</authors>
     <description>Metapackage for .NET6 to ensure all packages use the same version.</description>
     <dependencies>
       <dependency id="common.vm" />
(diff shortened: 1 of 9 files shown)


naacbin:
Dokan is in failure (which cause memprocfs too) but I haven't got the time to investigate the error

packages/dotnet-6.vm/dotnet-6.vm.nuspec
```diff
@@ -2,13 +2,13 @@
...
     <id>dotnet-6.vm</id>
-    <version>0.0.0.20231224</version>
+    <version>0.0.0.20240507</version>
     <authors>Microsoft</authors>
     <description>Metapackage for .NET6 to ensure all packages use the same version.</description>
     <dependencies>
       <dependency id="common.vm" />
-      <dependency id="dotnet-6.0-desktopruntime" version="(6.0.14, 6.0.26]" />
-      <dependency id="dotnet-6.0-runtime" version="(6.0.14, 6.0.26]" />
+      <dependency id="dotnet-6.0-desktopruntime" version="[6.0.14, 6.1)" />
```
Ana06, line 10: why do we need to increase the .NET version?

Ana06 (commented):
Thanks for fixing the packages @naacbin! :cupid:

packages/dotnet-6.vm/dotnet-6.vm.nuspec
naacbin: This isn't needed but the syntax was incorrect "(]" instead of "[)" and increasing minor version won't be a problem.

Ana06 (approved):

PR #1037, opened 2024-05-13
Add ida.plugin.hashdb.vm

Ana06:
Add ida.plugin.hashdb.vm. We couldn't use the issues automation as we need to install the requirements (the `requests` Python library), but being able to use the script/template/helper functions made it very easy to create it manually.


mr-tz (approved):
Cool, the required Internet connection is a bummer though.

Issue #851, opened 2024-01-22
Document local testing setup


Ana06:
### Details

Several people has asked me what is my local testing setup.

For testing new packages I recommend to:
- using an empty windows
- install flare-vm with a configuration with just one simple package (for example hashmyfiles.vm)
- Copy the script `test_install.ps1` to the Desktop and create a folder `packages` in the Desktop
- Take a snapshot `testing`
- Every time you test a new package, revert to `testing`, copy the package to test to the `packages` folder and run the `test_install.ps1` script

This way you ensure that all needed dependencies are installed. 

Document also how to test pinging of icons to taskbar.

@vm-packages anything else regarging local testing that should be documented?

Related https://github.com/mandiant/VM-Packages/issues/728

naacbin:
LGTM ! Nothing to add on my side

emtuls:
Looks good! Thanks for documenting all of that.

Ana06:
Closing this issue then! @mandiant/commando-vm do you have any commando VM specific testing setup? If so, it would be a good idea to open an issue to add some documentation to this new page when you have time. :wink:

day1player:
LGTM!

PR #1041, opened 2024-05-15
Improve test_install script

Ana06:
Improve `test_install.ps1` script:
- Allow the `packages` folder in the same directory to be a link to the actual directory. This is useful when testing packages in a shared folder. Related: https://github.com/mandiant/VM-Packages/issues/851
- Delete previously built packages in `built_pkgs` to avoid testing old packages.
- Simplify restoring location: `-PassThru` displays the location and `| Out-Null` discard the output, so we can remove both.


emtuls (approved):
Awesome. Looks good! Good quality of life updates. :)

PR #1049, opened 2024-05-16
:robot: Add js-deobfuscator.vm

vm-packages:
Automated PR from https://github.com/mandiant/VM-Packages/issues/1048 :rocket:

PR #1050, opened 2024-05-17
[common] Print directory in Powershell shortcuts

Ana06:
Fix bug I noticed while testing https://github.com/mandiant/VM-Packages/pull/1049#pullrequestreview-2061233129: Make Powershell shortcuts consistent with the cmd one by printing the execution directory.

### Before
image
### After 
image


emtuls (approved):
Awesome! Good update! Makes it more like our CMD one. :)

Issue #972, opened 2024-04-02
VSCode Package not updating correct


emtuls:
### What's the problem?

As noted here: https://github.com/mandiant/VM-Packages/pull/947#issuecomment-2030533590

VSCode versioning seems to be out of sync and there is also an issue preventing it from updating properly that I'm not sure is something in our control.

Attempting to perform the update correctly led to an issue with resolving `chocolatey-core.extension >= 1.3.3` for some reason:
![image](https://github.com/mandiant/VM-Packages/assets/35854880/d65ad7e1-5b93-490b-bc9e-19138e77e62c)

### Steps to Reproduce

Change `vscode.vm.nuspec` package to use `1.87.2` for `vscode.install`. Maybe also want to change the version for the overall package (`vscode.vm`) to be `1.87.2.20240402`

### Environment

Environment agnostic

### Detected by test suite

No

Ana06:
I'll take a look into this because it has the potential to break many packages

Ana06:
It is not a bug in the update scrip as the version is not out of sync. The dependency is not called the same as the package, which means our test suite doesn't consider it to be a metapackage (this also follows our documentation). Options:

1. Change the version of this package to 0.0.0.DATE and not consider it a metapackage (similar to packages with a URL without version)
2. Change the name of this package to `vscode.install.vm` so that it is considered a metapackage
3. Change the dependency to `vscode` instead `vscode.install`, which seems to just be a wrapper, so that the the package is considered a metapackage and has the same name.

If there nothing against it, I would go for option 3. @emtuls opinions?

Ana06:
The package doesn't fail to install for me locally and it is also not failing in the daily. Can you please add detailed instructions how to reproduce the `chocolatey-core.extension` error?

Ana06:
@emtuls any updates on the error when installing this package? we still need to fix the version

PR #1051, opened 2024-05-17
[vscode] Fix metapackage version missmatch

Ana06:
Change the dependency to `vscode` instead `vscode.install`, which seems to just be a wrapper, so that the package and the dependency have the same name and the package is considered a metapackage. The CI will now update the version following the metapackage versioning convention.

Closes https://github.com/mandiant/VM-Packages/issues/972


emtuls (approved):
Glad it was a simple fix! Thank you for getting to it before me. :)

Issue #966, opened 2024-04-01
Use CyberChef icon instead of Chrome icon for CyberChef?


emtuls:
### Details

Currently we use the Chrome icon for opening Cyberchef which works well and is dual purpose since it also acts as a chrome shortcut, but is there a necessity for this case?

People in the community may assume that it is only Chrome Browser shortcut and not realize that it is also a CyberChef shortcut and due to this being a malware analysis environment, it might make sense to use the CyberChef icon versus a Chrome one. The functionality wouldn't change of course, just the icon.

Not sure if this is something we should consider or not and I definitely don't have strong feelings one way or another. I do like the CyberChef icon due to it being more explicit about what it is primarily for, but I also know that a Chrome shortcut may make a lot of sense.

Opening for discussion and thoughts on possible preference. :)

Ana06:
I would prefer the CyberChef icon in the Tools folder, but I am not sure if it may be confusing in the taskbar. :thinking:

emtuls:
Hmmm, if our goal is to have `CyberChef` pinned to the taskbar, it might be more clear to show an official `CyberChef` icon rather than Google Chrome, which people might assume may just be a Google Chrome browser shortcut.

From the perspective of someone who might not be too familiar with FlareVM, this is what I could imagine:
If someone wants to use Chrome, I would imagine they would likely click the current icon in the taskbar, which theoretically does work, but is not our intention for it. 
And if someone wants to use CyberChef, they may not realize that the shortcut is for CyberChef (since it just looks like a Google Chrome shortcut) and they would likely search for "CyberChef" and find it in the Tools folder rather than use the shortcut. That is, until they maybe one day realize that the shortcut in the taskbar leads them to opening CyberChef directly.

stevemk14ebr:
I would love a cyberchef icon pinned to the dash. I do this myself with a little bash script and then change the icon on that to cyberchef

Ana06:
@stevemk14ebr we are using [CustomStartLayout.xml](https://github.com/mandiant/flare-vm/blob/main/CustomStartLayout.xml) to ping icons to the taskbar, so the implementation itself is not difficult. :wink: It seems we agree that we prefer having the CyberChef icon in the Tools folder and the taskbar. I think this could be implemented together with 
[...]

PR #1052, opened 2024-05-22
:robot: Package update

vm-packages:
Automated package update

Issue #1021, opened 2024-04-30
Remove Chrome dependency in cyberchef.vm thank to new changes


Ana06:
### Details

I think with the changes in https://github.com/mandiant/VM-Packages/pull/1016 we could now remove  the Chrome dependency in Cyberchef and the code to open it with chrome and set the icon (simplifying the installer code). I think we would keep the current behavior if both cyberchef.vm and googlechrome.vm are installed and remove the chrome dependency would allow to have a different browser as default (and to open cyberchef) if only cyberchef.vm is installed. This would need to add googlechrome.vm explicitly in [flare-vm](https://github.com/mandiant/flare-vm/blob/main/config.xml).

@emtuls what do you think?

Ana06:
`VM-Set-Open-With-Association $exePath ".html"` changes the icon but not the default icon, right? I think this is confusing, we should change the icon of html + default browser or remove the icon association. Let's discuss in https://github.com/mandiant/VM-Packages/issues/822 if we can change the default browser.

emtuls:
I may have a working solution for this. 😊 I still need to test how the new shortcut would work if chrome was not installed, but at least for the chrome case...After using `VM-Set-Open-With-Association $exePath ".html"`, we just need to create the following registry key and then restart `explorer.exe`:
```
[HKEY_LOCAL_MACHINE\SOFTWARE\Policies\Microsoft\Windows\Explorer]
“NoNewAppAlert”=dword:00000001
```

**NOTE**: This does not make `chrome` the default browser still, which also makes the popup asking to make `chrome` the default browser still appear. I can look into seeing if I can bypass that as well, without actually setting the default browser, but it may not be possible.

Ana06:
Because of the changes in https://github.com/mandiant/VM-Packages/pull/1148, we can't remove the dependency

Issue #822, opened 2024-01-11
Set Chrome as default browser


Ana06:
### Details

I think we should add a `googlechrome.defaultbrowser.vm` package to add Chrome as default browser as proposed in https://github.com/mandiant/VM-Packages/issues/16. 

One idea would be to use the following PowerShell script
```
Add-Type -AssemblyName System.Windows.Forms
Start ms-settings:defaultapps
Sleep 2
[System.Windows.Forms.SendKeys]::SendWait("{TAB 5}{ENTER}")
Sleep 1
[System.Windows.Forms.SendKeys]::SendWait("{TAB 2}{ENTER}");
```

But this only works if Google Chrome is the third browser in the list, what I am not sure that is always the case. This is how it looks like for me:
![Screenshot 2024-01-11 at 17 39 56](https://github.com/mandiant/VM-Packages/assets/16052290/e8e5147d-5840-4464-8ec6-64d0814a46ce)

How does it look for others? Does anyone have a better idea to implement it? 🤔

emtuls:
Did we determine if https://github.com/DanysysTeam/PS-SFTA from https://github.com/mandiant/VM-Packages/issues/823 is something we can use or at least take a part of the code to use in our own way?

If not, could we use the alternative from chocolatey? https://community.chocolatey.org/packages/setdefaultbrowser

Ana06:
I have asked legal and they have advised that we send an issue/PR to request the addition of the LICENSE file if we want to use the code.

Ana06:
@emtuls which of the two options ([`PS-SFTA`](https://github.com/DanysysTeam/PS-SFTA) or the `[setdefaultbrowser](https://community.chocolatey.org/packages/setdefaultbrowser) browser) would you prefer to use? If we prefer to use [`PS-SFTA`](https://github.com/DanysysTeam/PS-SFTA) I can ask them to include the license file.

emtuls:
Whichever is easier imo, but `setdefaultbrowser` seems pretty easy with it being in chocolatey already. :)

PR #1054, opened 2024-05-23
Update CyberChef to use CyberChef icon

emtuls:
This converts the CyberChef PNG file that is provided with the download and converts it to a `.ico` so that Chocolatey can use it for an icon for the shortcut. 

Closes https://github.com/mandiant/VM-Packages/issues/966

### DIFF
--- a/packages/cyberchef.vm/cyberchef.vm.nuspec
+++ b/packages/cyberchef.vm/cyberchef.vm.nuspec
@@ -2,7 +2,7 @@
 <package xmlns="http://schemas.microsoft.com/packaging/2015/06/nuspec.xsd">
   <metadata>
     <id>cyberchef.vm</id>
-    <version>10.19.0</version>
+    <version>10.19.0.20240913</version>
     <authors>GCHQ</authors>
     <description>The Cyber Swiss Army Knife - a web app for encryption, encoding, compression, data analysis, and more.</description>
     <dependencies>
--- a/packages/cyberchef.vm/tools/chocolateyinstall.ps1
+++ b/packages/cyberchef.vm/tools/chocolateyinstall.ps1
@@ -25,7 +25,7 @@ try {
   $htmlPath = Join-Path $toolDir "CyberChef_v10.19.0.html" -Resolve
   $arguments = "start chrome $htmlPath && exit"
   $executableArgs = "/C $arguments"
-  $iconLocation = "%ProgramFiles%\Google\Chrome\Application\chrome.exe"
+  $iconLocation = VM-Create-Ico (Join-Path $toolDir "images\cyberchef-128x128.png") # Create .ico for cyberchef icon
 
   Install-ChocolateyShortcut -ShortcutFilePath $shortcut -TargetPath $executableCmd -Arguments $executableArgs -WorkingDirectory $toolDir -WindowStyle 7 -IconLocation $iconLocation
 
--- a/packages/common.vm/common.vm.nuspec
+++ b/packages/common.vm/common.vm.nuspec
@@ -2,7 +2,7 @@
 <package xmlns="http://schemas.microsoft.com/packaging/2015/06/nuspec.xsd">
   <metadata>
     <id>common.vm</id>
-    <version>0.0.0.20240826</version>
+    <version>0.0.0.20240913</version>
     <description>Common libraries for VM-packages</description>
     <authors>Mandiant</authors>
   </metadata>
--- a/packages/common.vm/tools/vm.common/vm.common.psm1
+++ b/packages/common.vm/tools/vm.common/vm.common.psm1
@@ -1800,3 +1800,19 @@ function VM-Set-Legal-Notice {
     New-ItemProperty -Path $RegistryPath -Name legalnoticetext -Value $legalnoticetext -Force
 }
 
+# Converts image file to .ico needed for file icons
+function VM-Create-Ico {
+    param (
+        [string]$imagePath
+    )
+    Add-Type -AssemblyName System.Drawing
+    $imageDirPath = Split-Path -Path $imagePath -Parent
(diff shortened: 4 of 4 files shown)


mr-tz (approved):

packages/cyberchef.vm/tools/chocolateyinstall.ps1
```diff
@@ -19,13 +19,23 @@ try {
...
+  $fs.Close()
+
   $shortcutDir = Join-Path ${Env:TOOL_LIST_DIR} $category
   $shortcut = Join-Path $shortcutDir "$toolName.lnk"
   $executableCmd  = Join-Path ${Env:WinDir} "system32\cmd.exe" -Resolve
   $htmlPath = Join-Path $toolDir "CyberChef_v10.18.3.html" -Resolve
   $arguments = "start chrome $htmlPath && exit"
   $executableArgs = "/C $arguments"
-  $iconLocation = "%ProgramFiles%\Google\Chrome\Application\chrome.exe"
+  $iconLocation = $icoPath
```
Ana06, line 38: can we rename `$icoPath` to `$iconLocation` and remove this assignation?

packages/cyberchef.vm/tools/chocolateyinstall.ps1
```diff
@@ -19,13 +19,23 @@ try {
   Install-ChocolateyZipPackage @packageArgs
   VM-Assert-Path $toolDir
 
+  # Create .ico for cyberchef icon
+  Add-Type -AssemblyName System.Drawing
```
Ana06, line 23: [nitpick] Could we need this feature (converting a .png to .ico) for other packages? Maybe worth to move this code to a helper in common. :thinking:

Ana06 (changes requested):

Ana06 (changes requested):
Thanks for the changes @emtuls! After some more testing, I am not sure that converting the image to icon is the best approach. The current code make the shortcut looks like:
image
But we could downloa and use directly the icon in https://raw.githubusercontent.com/gchq/CyberChef/18159ce806e63286019a5a6828422a992fbb253b/src/web/static/images/favicon.ico which I think looks a bit better and removes the need for the extra code:
image

It is also confusing and inconsistent with other apps (like capa or fakenet) that the icon in the taskbar is a different one. This is related to https://github.com/mandiant/VM-Packages/issues/1021 and https://github.com/mandiant/VM-Packages/issues/822, but I think we should address then in the same PR so that the icon change takes place everywhere.

packages/cyberchef.vm/cyberchef.vm.nuspec
```diff
@@ -2,7 +2,7 @@
 <package xmlns="http://schemas.microsoft.com/packaging/2015/06/nuspec.xsd">
   <metadata>
     <id>cyberchef.vm</id>
-    <version>10.18.3</version>
+    <version>10.18.3.20240523</version>
```
Ana06, line 5: It needs update because https://github.com/mandiant/VM-Packages/pull/1052 has been merged

Ana06:
@emtuls :eyes:

packages/cyberchef.vm/tools/chocolateyinstall.ps1
```diff
@@ -19,13 +19,13 @@ try {
   Install-ChocolateyZipPackage @packageArgs
   VM-Assert-Path $toolDir
 
+  $iconLocation = VM-Create-Ico (Join-Path $toolDir "images\cyberchef-128x128.png") # Create .ico for cyberchef icon
```
Ana06, line 22: @emtuls have you forgotten to push the updated changes? :thinking: This code is still creating the `.ico` instead of using https://raw.githubusercontent.com/gchq/CyberChef/18159ce806e63286019a5a6828422a992fbb253b/src/web/static/images/favicon.ico that seems to looks better and does not require the extra function

emtuls:
@mandiant/flare-vm Any preference or thoughts on which icon to use for Cyberchef? 
One icon uses the robot-chef mix, but has a slight blurry look when viewed in the Tools folder in small/detail view (looks good on taskbar and desktop though).
cyberchef
image

The other icon is just a Chef hat (also provided by Cyberchef as a `.ico`, so not a random chef hat), which looks a little better in small/detail view, but maybe slightly blurry when on Desktop and okay in taskbar
image
image

Another thing to note is that the Robot-Chef mix icon requires some additional code to convert an image file to a `.ico` file.

Ana06:
@d35ha can you test this locally and share your opinion on this? :smile:

Ana06:
@stevemk14ebr @mr-tz @binjo @d35ha any preference on the icon? The icons look a bit different (see screenshots in the previous comments). @emtuls prefers to generate Robot-Chef icon (requires some additional code to convert an image file to a `.ico` file), while I prefer the Chef hat.

stevemk14ebr:
I don't mind either, I vote for whichever is easiest to maintain

mr-tz:
> I don't mind either, I vote for whichever is easiest to maintain

I second this. FWIW taskbars may be configured in other colors (mine is black). I agree that gray on gray is a bit hard to see.

Ana06:
I think the Chef hat is easier to maintain because it requires less code. But as it seems nobody else has a strong opinion, I'll let @emtuls decide. @emtuls can you please rebase the final version so that we can merge this change? :smile:

emtuls:
After googling CyberChef and seeing most websites/videos use the Robot-Chef icon, I'm inclined to keep that style for now. :)

Ana06 (approved):
LGTM! :+1: 

I find it confusing to have the CyberChef icon in the taskbar and that when clicking it, it opens Chrome as a different taskbar icon:

image

But I think this is already part of https://github.com/mandiant/VM-Packages/issues/1100 and we can continue to discuss how to address it there.

PR #1056, opened 2024-05-27
NetworkMiner 2.9 Released

erik4711:
* TZSP support
* StealC extractor
* Improved Modbus parser
* JA4 support
* GTP decapsulation
https://www.netresec.com/?page=Blog&month=2024-05&post=NetworkMiner-2-9-Released


Ana06 (approved):
Thanks for the hash update @erik4711! :smiling_face_with_three_hearts:  It would be better to use a URL with the version so that it doesn't break with updates, but I guess there is no such a URL.

[nitpick] For next time: It would be appreciated if you could squash your commits to avoid `fixup` or `update nuspec` commits in order to keep a clean commit history. :wink:

PR #1057, opened 2024-05-27
Fix wrong version for NetworkMiner

naacbin:
When merging #1056 a wrong version format was pushed. This fixes the version, however it may be necessary to remove the version `2.9.20240411` from MyGet.


packages/networkminer.vm/networkminer.vm.nuspec
```diff
@@ -2,7 +2,7 @@
 <package xmlns="http://schemas.microsoft.com/packaging/2015/06/nuspec.xsd">
   <metadata>
     <id>networkminer.vm</id>
-    <version>2.9.20240411</version>
```
Ana06, line 5: The new version is more correct, but this version would be ok too, as the date is only use to ensure the version is increased. I'll remove the version from MyGet and merge the change.

Ana06 (approved):

Issue #997, opened 2024-04-16
Simplify Windbg code && fix icon


Ana06:
### Details

Use `$executablePath = "$(where.exe $toolName)"` to simplify the code of windbg.vm by executing WinDbg directly instead as a consoleApp. This fixes the icon too.

naacbin:
As mentioned in  #1040 we could use ` https://windbg.download.prss.microsoft.com/dbazure/prod/1-2402-24001-0/windbg.msixbundle` to download Windbg as a msi which is handle more easily instead of using AppxPackage that will download the installer. Moreover, this allow versioning of Windbg.

naacbin:
@Ana06, if you agree I can make the change.

naacbin:
I had some free time so I did a PR.

PR #1058, opened 2024-05-27
Refactor WinDbg install

naacbin:
I rework WinDbg install to improve it (versioned + icon support #997)

Closes https://github.com/mandiant/VM-Packages/issues/997

### DIFF
--- a/packages/windbg.vm/tools/chocolateyinstall.ps1
+++ b/packages/windbg.vm/tools/chocolateyinstall.ps1
@@ -5,17 +5,25 @@ try {
     $toolName = 'WinDbg'
     $category = 'Debuggers'
 
-    # It seems WinDbg is now distributed as an .appinstaller and we need to install it using Add-AppxPackage
-    Add-AppxPackage -AppInstallerFile 'https://windbg.download.prss.microsoft.com/dbazure/prod/1-0-0/windbg.appinstaller'
+    $bundleVersion = "1-2402-24001"
+    $bundleUrl = "https://windbg.download.prss.microsoft.com/dbazure/prod/$bundleVersion-0/windbg.msixbundle"
+    $bundleSha256 = "e941076cb4d7912d32a22ea87ad2693c01fa465227b4d1ead588283518de428f"
 
-    $shortcutDir = Join-Path ${Env:TOOL_LIST_DIR} $category
-    $shortcut = Join-Path $shortcutDir "$toolName.lnk"
-    $executableCmd  = Join-Path ${Env:WinDir} "system32\cmd.exe"
-    # Use `start` to close the open console
-    $executableArgs = "/C start WinDbgX.exe"
-    $executableDir  = Join-Path ${Env:UserProfile} "Desktop"
-    Install-ChocolateyShortcut -shortcutFilePath $shortcut -targetPath $executableCmd -Arguments $executableArgs -WorkingDirectory $executableDir -RunAsAdmin
+    $packageArgs = @{
+        packageName   = ${Env:ChocolateyPackageName}
+        url           = $bundleUrl
+        checksum      = $bundleSha256
+        checksumType  = "sha256"
+        fileFullPath  = Join-Path ${Env:TEMP} "$toolName.msixbundle"
+    }
+    Get-ChocolateyWebFile @packageArgs
+    Add-AppxPackage -Path $packageArgs.fileFullPath
+
+    $toolPackage = Get-AppxPackage -Name "Microsoft.$toolName"
+    $iconLocation = Join-Path $toolPackage.InstallLocation "DbgX.Shell.exe" -Resolve
+    $executablePath = "shell:AppsFolder\$($toolPackage.PackageFamilyName)!$($toolPackage.Name)"
+
+    VM-Install-Shortcut -toolName $toolName -category $category -executablePath $executablePath -iconLocation $iconLocation -RunAsAdmin
 } catch {
     VM-Write-Log-Exception $_
 }
-
--- a/packages/windbg.vm/windbg.vm.nuspec
+++ b/packages/windbg.vm/windbg.vm.nuspec
@@ -2,7 +2,7 @@
 <package xmlns="http://schemas.microsoft.com/packaging/2015/06/nuspec.xsd">
   <metadata>
     <id>windbg.vm</id>
-    <version>0.0.0</version>
+    <version>1.2402.24001.20250124</version>
     <authors>Microsoft</authors>
     <description>WinDbg is a debugger that can be used to analyze crash dumps, debug live user-mode and kernel-mode code, and examine CPU registers and memory.</description>
     <dependencies>


packages/windbg.vm/tools/chocolateyinstall.ps1
```diff
@@ -5,16 +5,22 @@ try {
...
-    Install-ChocolateyShortcut -shortcutFilePath $shortcut -targetPath $executableCmd -Arguments $executableArgs -WorkingDirectory $executableDir -RunAsAdmin
+    $packageArgs = @{
+        packageName   = ${Env:ChocolateyPackageName}
+        url           = $exeUrl
+        checksum      = $exeSha256
+        checksumType  = "sha256"
+        fileFullPath  = Join-Path ${Env:TEMP} "$toolName.msixbundle"
+    }
+    Get-ChocolateyWebFile @packageArgs
+    Add-AppxPackage -Path $packageArgs.fileFullPath
```
Ana06, line 19: What about simplifying this code using the MSIX bundle URL directly in `Add-AppxPackage` instead of downloading it? As the URL includes the version, the change still allows versioning of Windbg, but with a simpler code.
```PowerShell
Add-AppxPackage "https://windbg.download.prss.microsoft.com/dbazure/prod/1-2402-24001-0/windbg.msixbundle"
```

packages/windbg.vm/tools/chocolateyinstall.ps1
```diff
@@ -5,16 +5,22 @@ try {
...
+
+    $installDir = (Get-AppxPackage -Name "Microsoft.$toolName").InstallLocation
+    $executablePath = Join-Path $installDir "DbgX.Shell.exe"
```
Ana06, line 22: have you tested this code locally? I get an error when trying to run the shortcut:
image

The following code works:
```suggestion
    $installDir = (Get-AppxPackage -Name "Microsoft.$toolName").InstallLocation
    $iconLocation  = Join-Path $installDir "DbgX.Shell.exe" -Resolve
    $executablePath = "$(where.exe WinDbgX.exe)"
    VM-Install-Shortcut -toolName $toolName -category $category -executablePath $executablePath -iconLocation $iconLocation -RunAsAdmin
```

Note I have also added a `-Resolve` to `Join-Path` to ensure the `DbgX.Shell.exe` exists.

packages/windbg.vm/tools/chocolateyinstall.ps1
```diff
@@ -5,16 +5,22 @@ try {
...
+    VM-Install-Shortcut -toolName $toolName -category $category -executablePath $executablePath
```
Ana06, line 23: I think we want to run `WinDbg` as Admin :thinking: 
```suggestion
    VM-Install-Shortcut -toolName $toolName -category $category -executablePath $executablePath -RunAsAdmin
```

Ana06 (changes requested):
Thanks for working on this @naacbin. It is nice that the shortcut has the WinDbg icon! :rocket:! 
image

packages/windbg.vm/tools/chocolateyinstall.ps1
naacbin: Yes, I have tested locally strange...

packages/windbg.vm/tools/chocolateyinstall.ps1
naacbin: On my system it run as admin by default without this option, but I will add it to force the start as admin.

packages/windbg.vm/tools/chocolateyinstall.ps1
naacbin: How, did you test ? I have tried from `Tools` directory and the Windows bar, it seems to work. Maybe it was the `RunAsAdmin` that was needed. Can you try again ?

> I have added the `-Resolve` parameter

naacbin: Weird, the only difference between our setup is that my empty VM only has `chocolatey`, I don't have a snapshot with `install.ps1` installed (yet). If I get some time, I will investigate on this, but anyway, your proposal seems to work for me too.

packages/windbg.vm/tools/chocolateyinstall.ps1
naacbin: For me, this doesn't mean you shouldn't check the hash value. Connection failure could have an impact on receiving the full file or someone could carry out a MITM attack. I can remove this if you think it's necessary but for me, the code overhead isn't that much, in addition to being consistent with other packages.

packages/windbg.vm/tools/chocolateyinstall.ps1
naacbin: It seems that the CI of windows-2019 doesn't appreciate `"$(where.exe WinDbgX.exe)"`.

naacbin: > Weird, the only difference between our setup is that my empty VM only has `chocolatey`, I don't have a snapshot with `install.ps1` installed (yet).

After creating a snapshot as [documented in the wiki](https://github.com/mandiant/VM-Packages/wiki/Local-testing) but with my current VM (which is build 22621) it still works (without the `where.exe`. So the problem, comes from Windows version.

packages/windbg.vm/tools/chocolateyinstall.ps1
Ana06: You are right that the extra check ensures we are installing what we expected and the code is not very complicated. So you have convinced me we should leave this code :wink: 

But I want to clarify that we are not checking the hash for all packages:
- For some packages that use a URL without the version, get updated often and install tools that are important from a "trusted" URL, we are using `VM-Assert-Signature` to ensure we can install the package without checking the hash (which breaks the package often). `VM-Assert-Signature` which checks that there is a valid signature and that the certificate was not signed by a publisher not trusted on the system. This is not ideal and @emtuls is proposing checking the signing authority in https://github.com/mandiant/VM-Packages/pull/1007, but we need to discuss how we want to handle this in more detail. Please feel free to share your thoughts and ideas on this!
- packages installed with `pip` and `npm`

Any ideas to improve the installation/verification of tools is very appreciated. :smile:

packages/windbg.vm/tools/chocolateyinstall.ps1
```diff
@@ -5,16 +5,23 @@ try {
...
+        checksum      = $exeSha256
+        checksumType  = "sha256"
+        fileFullPath  = Join-Path ${Env:TEMP} "$toolName.msixbundle"
+    }
+    Get-ChocolateyWebFile @packageArgs
+    Add-AppxPackage -Path $packageArgs.fileFullPath
+
+    $installDir = (Get-AppxPackage -Name "Microsoft.$toolName").InstallLocation
+    $iconLocation  = Join-Path $installDir "DbgX.Shell.exe" -Resolve
+    $executablePath = "$(where.exe WinDbgX.exe)"
```
Ana06, line 23: it seems like the windows-2019 run can't find `WinDbgX.exe` which likely mean it hasn't been installed... (likely it was also not working before, but we were not checking). The GH runner is using Windows `10.0.17763`, locally I am using `10.0.19045` for testing.

Searching online about issues installing WinDbg, I see information about conflicts with the Windows Software Development Kit. For example: https://superuser.com/questions/1524982/windbg-repeatedly-downloads-but-doesnt-install. Maybe requiring `windows-sdk-10.0` as dependency fixes the issue. :thinking: 

@mandiant/vms any other ideas how to fix this?

naacbin: Unfortunately, I won't be able to test it myself as I don't have any Windows 10 VM and can't get one easily to test for now.

Ana06: I have Windows 10, but newer than the one in GH actions and it works for me. I am also not sure how to get an older Windows 10 for local testing. We can use GH for testing by pushing new changes to this PR (or running the test in a fork to avoid the noise in the PR).

@mandiant/vms any ideas to fix or test this problem?

naacbin: > Searching online about issues installing WinDbg, I see information about conflicts with the Windows Software Development Kit. For example: https://superuser.com/questions/1524982/windbg-repeatedly-downloads-but-doesnt-install. Maybe requiring `windows-sdk-10.0` as dependency fixes the issue. 🤔

I tried to add `windows-sdk-10.0` as dependency but it still failed.

Ana06: If we do not figure out what the issue is, we can keep running it as a console app with:
```
VM-Install-Shortcut -toolName $toolName -category $category -executablePath "start $toolName.exe" -consoleApp $true -runAsAdmin -iconLocation $iconLocation
```

naacbin: I have tested this, however I get `'"start WinDbgX.exe"' is not recognized as an internal or external command, operable program or batch file.` because of the double quote added inside `VM-Install-Shortcut`.

`VM-Install-Shortcut -toolName $toolName -category $category -executablePath "start WinDbgX.exe" -consoleApp $true -RunAsAdmin -iconLocation $iconLocation`

naacbin: @Ana06, I have tried another idea by catching the exception, it seems to work !

Ana06: I do not like that we use `$executablePath = Join-Path $installDir "DbgX.Shell.exe"` as executing `DbgX.Shell.exe` does not work for me, so potentially this is broken for others too. I think without the `start` it should work and be equivalent to the code we had before:

VM-Install-Shortcut -toolName $toolName -category $category -executablePath "$toolName.exe" -consoleApp $true -runAsAdmin -iconLocation $iconLocation

Ana06: I have noticed I can't even accces `C:\Program Files\WindowsApps\` where `DbgX.Shell.exe` is, likely why I can't execute it
image

image

packages/windbg.vm/tools/chocolateyinstall.ps1
```diff
@@ -5,17 +5,32 @@ try {
...
-    $executableDir  = Join-Path ${Env:UserProfile} "Desktop"
-    Install-ChocolateyShortcut -shortcutFilePath $shortcut -targetPath $executableCmd -Arguments $executableArgs -WorkingDirectory $executableDir -RunAsAdmin
+    $packageArgs = @{
+        packageName   = ${Env:ChocolateyPackageName}
+        url           = $exeUrl
+        checksum      = $exeSha256
+        checksumType  = "sha256"
+        fileFullPath  = Join-Path ${Env:TEMP} "$toolName.msixbundle"
+    }
+    Get-ChocolateyWebFile @packageArgs
```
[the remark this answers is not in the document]
naacbin: This is not necessary, as `Get-ChocolateyWebFile` will throw an error with "incorrect hash" if the file has not been downloaded.

packages/windbg.vm/tools/chocolateyinstall.ps1
naacbin: > I have noticed I can't even accces `C:\Program Files\WindowsApps\` where `DbgX.Shell.exe` is, likely why I can't execute it


I don't think it's related, at least not directly. Since you should be able to access`C:\Program Files\WindowsApps\Microsoft.WinDbg_1.2402.24001.0_x64__8wekyb3d8bbwe`.  in my installation (which works with `DbgX.Shell.exe`) I don't have access to WinDbg either.


I don't think it's related at least directly. Since you should be able to access `C:\Program Files\WindowsApps\Microsoft.WinDbg_1.2402.24001.0_x64__8wekyb3d8bbwe\`. Besides, in my setup (which works with `DbgX.Shell.exe`) I don't have access to `C:\Program Files\WindowsApps\` either.

naacbin: > I do not like that we use `$executablePath = Join-Path $installDir "DbgX.Shell.exe"` as executing `DbgX.Shell.exe` does not work for me, so potentially this is broken for others too. 

I only use it when `which` fails, so only for older versions of Windows 10, it seems. According to the current  `windbg.vm` code, it shouldn't work for these older versions of Windows either because `WinDbgX` is hard-coded without `VM-Assert-Path`.

> I think without the `start` it should work and be equivalent to the code we had before:
> 
> `VM-Install-Shortcut -toolName $toolName -category $category -executablePath "$toolName.exe" -consoleApp $true -runAsAdmin -iconLocation $iconLocation `

If I do this, it will still not work for older versions of Windows 10, I think my solution might give it a chance to actually work.

emtuls:
So I'm looking into this and it looks like the same issue I ran into when I was trying to do the `windows-terminal` package for `wt.exe` (https://github.com/mandiant/VM-Packages/issues/934#issuecomment-1977815578) where it installed to the correct location for `Windows Server 2022`, but not for `Windows Server 2019`. Unfortunately, I never found a solution to do it properly for Windows Apps, and instead, I went the route of using a Github package, which doesn't seem to exist for `WindbgX`.

It seems that the `Windows Server 2019` either installs the Windows Apps to a different location than the standard one (or installs it differently in general), or it's failing to install for some reason. But the fact that `$iconLocation  = Join-Path $installDir "DbgX.Shell.exe" -Resolve` actually resolves leads me to believe that it is installing, just not how we expect.

Now the issue with not being able use that as the executable location is because it's a highly protected folder, so @Ana06 is likely correct in that the shortcut to that as the `executablePath` in the `catch` statement is very unlikely to actually be working and I'm not sure if that's the route we want to take. There seems to be ways that we can possibly change the ownership of the directory and be able to potentially make it run, as noted [here](**url**), but I don't think that's the best route to take.

I'm thinking that the better route might be what @Ana06 suggested with https://github.com/mandiant/VM-Packages/pull/1058#discussion_r1644253183 where we use a Console App.

I will try to do a little more testing to see if I can find an alternative, but that's my current take at the moment. :)

emtuls:
Judging from this post, it sounds like the `WinDbgX.exe` and similar executables in `Microsoft\WindowsApps` are considered an "Execution Alias", which is not supported in Windows Server 2019: https://stackoverflow.com/questions/67495731/how-do-i-make-an-execution-alias-work-on-windows-server-2019
This is why it's not created when we do `Add-AppxPackage` for the `windows-2019` github runner.

A suggested alternative that may work for us would be to create a command similar to `Explorer.Exe Shell:AppsFolder:<ApplicationID>!<AppID>` which looks like it can potentially be done programmatically: https://www.andreasnick.com/106-starting-msix-and-appx-packages-with-powershell-from-the-command-line.html

Or a powershell command `Invoke-CommandInDesktopPackage`: https://learn.microsoft.com/en-us/powershell/module/appx/invoke-commandindesktoppackage?view=windowsserver2019-ps

I'm not sure what the best solution really is, but for it to truly work on Windows Server 2019 we'd likely need one of these options, as `VM-Install-Shortcut -toolName $toolName -category $category -executablePath "$toolName.exe" -iconLocation $iconLocation -RunAsAdmin` is merely making a shortcut to something that may not actually be working. Unfortunately I do not have this sort of setup at the moment to test this.

I'd be willing to accept `VM-Install-Shortcut -toolName $toolName -category $category -executablePath "$toolName.exe" -iconLocation $iconLocation -RunAsAdmin`, as it seems that there have not been any complaints as of yet for it. 

@Ana06 thoughts?

packages/windbg.vm/tools/chocolateyinstall.ps1
```diff
@@ -5,17 +5,32 @@ try {
...
-    VM-Write-Log-Exception $_
+    if ($_.Exception.Message -match "INFO: Could not find files for the given pattern\(s\).")
+    {
+        $executablePath = Join-Path $installDir "DbgX.Shell.exe"
+        VM-Install-Shortcut -toolName $toolName -category $category -executablePath $executablePath -iconLocation $iconLocation -RunAsAdmin
+    }
+    else
+    {
+        VM-Write-Log-Exception $_
+    }
```
emtuls, line 34: I'm not sure if this is the route we want to take, as noted in my comments on the PR.

I think our current route of just making a shortcut may be the better alternative for the moment, unless we want to take the time to properly test and confirm one of the possible working solutions I noted in my comments.

emtuls (dismissed):

naacbin:
> Now the issue with not being able use `DbgX.Shell.exe` as the executable location is because it's in a highly protected folder, so @Ana06 is likely correct in that the shortcut to that as the `executablePath` in the `catch` statement is very unlikely to actually be working and I'm not sure if that's the route we want to take. There seems to be ways that we can possibly change the ownership of the directory and be able to potentially make it run, as noted [here](**url**), but I don't think that's the best route.

I don't think changing permission of the directory will solve anything, because on my side it's working with `DbgX.Shell.exe` without modifying the permission of the protected folder. However, I agree that it's unlikely to work on Windows Server 2019.

naacbin:
> I'd be willing to accept `VM-Install-Shortcut -toolName $toolName -category $category -executablePath "$toolName.exe" -iconLocation $iconLocation -RunAsAdmin`, as it seems that there have not been any complaints as of yet for it.

So we merge this PR with the line provided and create an issue to find a better way with `Explorer.Exe Shell:AppsFolder:<ApplicationID>!<AppID>` or `Invoke-CommandInDesktopPackage` ?

packages/windbg.vm/tools/chocolateyinstall.ps1
naacbin: At least we need to continue on this error, or `windows-2019` will fail on `where.exe WinDbgX.exe`.

packages/windbg.vm/tools/chocolateyinstall.ps1
```diff
@@ -5,17 +5,32 @@ try {
...
+        checksum      = $bundleSha256
+        checksumType  = "sha256"
+        fileFullPath  = Join-Path ${Env:TEMP} "$toolName.msixbundle"
+    }
+    Get-ChocolateyWebFile @packageArgs
+    Add-AppxPackage -Path $packageArgs.fileFullPath
+
+    $installDir = (Get-AppxPackage -Name "Microsoft.$toolName").InstallLocation
+    $iconLocation  = Join-Path $installDir "DbgX.Shell.exe" -Resolve
+    $executablePath = "$(where.exe WinDbgXA.exe)"
```
d35ha, line 23: Actually `where.exe` doesn't throw exceptions, it returns `1` in case of failure and since `$ErrorActionPreference` is set to `Stop`; the non-zero return code is treated as an exception with the `stderr` as exception message.

It's a way better to use the powershell function `Get-Command` with the `ErrorAction` set to `Continue` and to check the returned path if empty or not instead of relying on exceptions.

It would be like this:
```powershell
$executablePath = (Get-Command WinDbgXA.exe -ea 0).Path
if (!$executablePath)
{
    $executablePath = Join-Path $installDir DbgX.Shell.exe
}
```

d35ha (changes requested):

packages/windbg.vm/tools/chocolateyinstall.ps1
```diff
@@ -5,17 +5,32 @@ try {
...
+        packageName   = ${Env:ChocolateyPackageName}
+        url           = $bundleUrl
+        checksum      = $bundleSha256
+        checksumType  = "sha256"
+        fileFullPath  = Join-Path ${Env:TEMP} "$toolName.msixbundle"
+    }
+    Get-ChocolateyWebFile @packageArgs
+    Add-AppxPackage -Path $packageArgs.fileFullPath
+
+    $installDir = (Get-AppxPackage -Name "Microsoft.$toolName").InstallLocation
```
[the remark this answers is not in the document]
Ana06: This is the error I got as well.

packages/windbg.vm/tools/chocolateyinstall.ps1
Ana06: @d35ha your proposal could cause that `$executablePath` is set to `Join-Path $installDir DbgX.Shell.exe` which doesn't fail the package, but we know fails to open in Windows 10 because of being in a protected directory. This results in a broken shortcut without noticing during the installation, or am I missing something?

Ana06:
@naacbin sorry that it took that long to review the PR, I am not sure which the best solution is. @d35ha thanks for sharing your feedback, hopefully we can soon find a good solution. :wink:

d35ha (approved):

d35ha:
> @naacbin sorry that it took that long to review the PR, I am not sure which the best solution is. @d35ha thanks for sharing your feedback, hopefully we can soon find a good solution. 😉

@Ana06 
A simple solution is to create a shortcut that points to the application directly under `shell:AppsFolder`, in that case the shortcut target type is set to `PackageFamilyName!PackageName` effectively running the windows application directly without any middlewares (ex. `cmd.exe`).

Ana06 (approved):
Smart solution @d35ha! :bowtie: thanks for taking this over so that we can finally merge it.

@naacbin sorry, you put a lot of effort into implementing this enhancement and we were not very fast responding/reviewing your changes. I was busy working on other priority tasks and I didn't have the time to investigate and issue for which I didn't have a good solution. @naacbin can you please let us know if you are ok with @d35ha proposal? :pray:

PR #1059, opened 2024-05-28
Added static download link for NetworkMiner 2.9

erik4711:


erik4711:
It feels a bit awkward to set the NetworkMiner version to something other than `2.9.0` in [networkminer.vm.nuspec](https://github.com/mandiant/VM-Packages/blob/main/packages/networkminer.vm/networkminer.vm.nuspec). Is it possible to set the nuspec version to something like `2.9.0-static` or similar to indicate that the version is the same, but the source URL has changed?

Ana06:
We use the forth segment when changing a package but keeping the same tools version, as documented in the [wiki](https://github.com/mandiant/VM-Packages/wiki/Chocolatey-packages#version). In this case: `2.9.0.20240621`.

emtuls:
Hi @erik4711! 

Wanted to follow up to see if you would be able to make the change in the `.nuspec` from `<version>2.9.0</version>` to `<version>2.9.0.20240621</version>`? 😊

erik4711:
@emtuls @Ana06 See pull request #1115 with updated URL and nuspec version.

PR #1065, opened 2024-06-01
Fix trustedsec-remote-ops-bof package name

naacbin:
Fix the name of package reference in nuspec file


Ana06 (approved):
I thought our lint checked that the names matched :confused:  I'll open an issue for this, good catch!

Issue #728, opened 2023-10-27
Re-structure & improve documentation


Ana06:
### Details

We have a lot of documentation in the Wiki, but I find it difficult to navigate. Reorganise the wiki documentation (and maybe remove/summarize some part).

Ana06:
I have just done in a big update to the documentation, adding some important details that come up often in discussions but also simplifying it and restructuring to hopefully make it easier to use. @mandiant/vms @naacbin @MalwareMechanic any feedback? Can we close this issue?

naacbin:
I have read both [Chocolatey-packages](https://github.com/mandiant/VM-Packages/wiki/Chocolatey-packages) and [Coding-Conventions](https://github.com/mandiant/VM-Packages/wiki/Coding-Conventions#small-pull-requests) documentation, it's really good. I have nothing to add. Thanks for this work !

Ana06:
Thanks for the review @naacbin!

PR #1066, opened 2024-06-01
Fix openjdk.vm dependency version format

naacbin:
`openjdk.vm` dependency version wasn't respecting the convention, it should be `[x,y)` not `(x,y]`.  The version `21.0.1` need to be pickup as it is hard-coded in `openjdk.vm`.

### DIFF
--- a/packages/openjdk.vm/openjdk.vm.nuspec
+++ b/packages/openjdk.vm/openjdk.vm.nuspec
@@ -2,12 +2,12 @@
 <package xmlns="http://schemas.microsoft.com/packaging/2015/06/nuspec.xsd">
   <metadata>
     <id>openjdk.vm</id>
-    <version>0.0.0.20240202</version>
+    <version>0.0.0.20240531</version>
     <authors>Oracle</authors>
     <description>Metapackage for OpenJDK to ensure all packages use the same OpenJDK version.</description>
     <dependencies>
       <dependency id="common.vm" />
-      <dependency id="openjdk" version="(21.0.0, 21.0.1]" />
+      <dependency id="openjdk" version="[21.0.1, 21.0.2)" />
     </dependencies>
   </metadata>
 </package>


Ana06 (commented):
> openjdk.vm dependency version wasn't respecting the convention, it should be [x,y) not (x,y]

I am not sure what convention do you mean. I think we used this syntax to avoid our bot to update the dependency. So I think both versions are equivalent.

I think even relaxing it a bit should be ok, as I do not expect minor updates in openjdk to break anything and it would allow more flexibility (problems with dependencies incompatibilities): `[21.0.0, 21.1.0)`.

I think we do not have documentation on what is expected and should probably add some to the wiki. :thinking:

naacbin:
> I am not sure what convention do you mean. I think we used this syntax to avoid our bot to update the dependency. So I think both versions are equivalent.

I thought I had read in the documentation that we were using "[x,y)" and not '(x,y]', but I can't find it anywhere. I must be wrong.

naacbin:
> > openjdk.vm dependency version wasn't respecting the convention, it should be [x,y) not (x,y]
> 
> I am not sure what convention do you mean. I think we used this syntax to avoid our bot to update the dependency. So I think both versions are equivalent.
> 
> I think even relaxing it a bit should be ok, as I do not expect minor updates in openjdk to break anything and it would allow more flexibility (problems with dependencies incompatibilities): `[21.0.0, 21.1.0)`.
> 
> I think we do not have documentation on what is expected and should probably add some to the wiki. 🤔

We can't as currently the version is hard-coded in the PowerShell script.

Ana06 (approved):
I think the code is equivalent, but may be a bit easier to read, so I would say let's get it in!

Ana06:
Related: I have added some documentation about dependencies in https://github.com/mandiant/VM-Packages/wiki/Chocolatey-packages#dependencies @naacbin we have had some discussion recently about dependency versions, could you check it out and let me know if you have any feedback (either in https://github.com/mandiant/VM-Packages/issues/728 or opening a new issue as you see it fit).

PR #1067, opened 2024-06-01
Github version change on update_package.py

naacbin:
As commented in #1053:
`hayabusa` as well as other packages such as `upx` have a problem when the version for example `v2.15.0` is different of the version in the filename `hayabusa-2.15.0-win-x64.zip` because of the `v`. The update works only because it will fallback to `update_version_url` and the updated version is in one of the case of `get_increased_version`.

From my testing adding `v?` in the url of `update_github_url` will allow the update via github without breaking any other package auto update (but I think a deeper review is required).

### DIFF
--- a/scripts/utils/update_package.py
+++ b/scripts/utils/update_package.py
@@ -89,12 +89,12 @@ def update_github_url(package):
     # Use findall as some packages have two urls (for 32 and 64 bits), we need to update both
     # Match urls like https://github.com/mandiant/capa/releases/download/v4.0.1/capa-v4.0.1-windows.zip
     matches = re.findall(
-        "[\"'](?P<url>https://github.com/(?P<org>[^/]+)/(?P<project>[^/]+)/releases/download/(?P<version>[^/]+)/[^\"']+)[\"']",
+        "[\"'](?P<url>https://github.com/(?P<org>[^/]+)/(?P<project>[^/]+)/releases/download/v?(?P<version>[^/]+)/[^\"']+)[\"']",
         content,
     )
     # Match also urls like https://github.com/joxeankoret/diaphora/archive/refs/tags/3.0.zip
     matches += re.findall(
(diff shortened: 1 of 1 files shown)


scripts/utils/update_package.py
```diff
@@ -89,12 +89,12 @@ def update_github_url(package):
     # Use findall as some packages have two urls (for 32 and 64 bits), we need to update both
     # Match urls like https://github.com/mandiant/capa/releases/download/v4.0.1/capa-v4.0.1-windows.zip
     matches = re.findall(
-        "[\"'](?P<url>https://github.com/(?P<org>[^/]+)/(?P<project>[^/]+)/releases/download/(?P<version>[^/]+)/[^\"']+)[\"']",
+        "[\"'](?P<url>https://github.com/(?P<org>[^/]+)/(?P<project>[^/]+)/releases/download/v?(?P<version>[^/]+)/[^\"']+)[\"']",
```
Ana06, line 92: I like the idea, but this change breaks updates for tools which don't include `v` in the release name, for example:
- `https://github.com/myfreeer/7z-build-nsis/releases/download/23.01/7z2301-x86.exe`
- `https://github.com/SnaffCon/Snaffler/releases/download/1.0.150/Snaffler.exe`
- `https://github.com/winsiderss/si-builds/releases/download/3.0.7645/systeminformer-3.0.7645-release-bin.zip`

what about leaving the regular expression as it was and deleting the `v` afterward if contained in the matched version? I think that should work for all cases. :thinking:

naacbin: This shouldn't be the case. Checking on online regex it looks good :
image. How did you find that it breaks updates ?
If it's too difficult to solve with regex, I will do as you suggest.

Ana06: Sorry, I didn't realized the `v` has a `?` which makes it optional. The change LGTM, thanks!

Ana06 (approved):
This is a nice improvement, thanks @naacbin!

Issue #1063, opened 2024-05-29
dokan fails to install in the daily run


Ana06:
### Details

dokan fails to install in the daily run with the following error:
```
ERROR: Running ["C:\Windows\System32\msiexec.exe" /i "C:\Tools\Dokan\Dokan_x64.msi" /qn /norestart ] was not successful. Exit code was '1603'. Exit code indicates the following: Generic MSI Error. This is a local environment error, not an issue with a package or the MSI itself - it could mean a pending reboot is necessary prior to install or something else (like the same version is already installed). Please see MSI log if available. If not, try again adding '--install-arguments="'/l*v c:\dokan.vm_msi_install.log'"'. Then search the MSI Log for "Return Value 3" and look above that for the error..
The upgrade of dokan.vm was NOT successful.
```
Although it was working on https://github.com/mandiant/VM-Packages/pull/1053. I think the issue is that we are missing the [`validExitCodes` we use in `VM-Install-With-Installer`](https://github.com/mandiant/VM-Packages/blob/165ec34682f7cb5356d0e8e51a29e0cc2f9df54b/packages/common.vm/tools/vm.common/vm.common.psm1#L732C12-L732C44) in the dokan code. @naacbin would you like to try to fix this (as it is related to your latests PRs)?

naacbin:
I can check when I have some time, but I'm not sure its related to `validExitCodes`, even if this could help. I think the problem comes from the forced install of `dokan.vm`. In the pipeline `dokan.vm` is first installed as dependency for `arsenalimagemounter.vm` (which works from the daily pipeline) and then the installation is forced for `dokan.vm` again. As the package is already present the MSI throw the error you highlighted.

naacbin:
Closing as #1068 as been merged.

PR #1068, opened 2024-06-01
Add 1603 to validExitCodes

naacbin:
This should avoid exiting in error when dokan is already installed. #1063

### DIFF
--- a/packages/common.vm/common.vm.nuspec
+++ b/packages/common.vm/common.vm.nuspec
@@ -2,7 +2,7 @@
 <package xmlns="http://schemas.microsoft.com/packaging/2015/06/nuspec.xsd">
   <metadata>
     <id>common.vm</id>
-    <version>0.0.0.20240516</version>
+    <version>0.0.0.20240531</version>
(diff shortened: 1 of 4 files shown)


Ana06 (commented):
The dokan packages doesn't use the helper in common or am I missing something? If it is possible to adapt dokan to use the helper would be good.

I think adding the exit code is also useful for other packages as I remember using a workaround for another MSI installer because of the same issue (I think the chrome one).

Ana06 (approved):
LGTM, thanks @naacbin!

Ana06:
@emtuls I think we can add also the exit codes to googlechrome.vm and remove the `-ErrorAction SilentlyContinue` and `VM-Assert-Path $exePath` which we added because the MSI installer failed to install if the tool was already installed (which should fail with code `1603`):  https://github.com/mandiant/VM-Packages/blob/715d55a3d23dbbc1181a3fc782f898751b5a786a/packages/googlechrome.vm/tools/chocolateyinstall.ps1#L23C51-L23C80

Issue #1080, opened 2024-06-06
Template + automation for Python packages installed with pip


Ana06:
### Details

Code used by packages with tools installed with `pip`:
```
    VM-Pip-Install $toolName

    $executablePath = "$(where.exe $toolName)"
    $arguments = "--help"

    VM-Install-Shortcut $toolName $category $executablePath -consoleApp $true -arguments $arguments
```

Example: [magika.vm](https://github.com/mandiant/VM-Packages/blob/36c12f6727921dbe5fa83bd0e7ae3d711b56c821/packages/magika.vm/tools/chocolateyinstall.ps1#L8-L13)

Create a package template, issue template & automation for this package type.

Ana06:
With #1083, the installation code is now very simple:
```
VM-Install-With-Pip -toolName $toolName -category $category
```

The function supports `arguments`, that can be used if https://github.com/mandiant/VM-Packages/issues/1082 is implemented.

Ana06:
I think we should start fixing the version of the tools installed with Pip (as we have done with magika in https://github.com/mandiant/VM-Packages/pull/1179 and plan to do with ida.plugin.capa.vm as discussed in https://github.com/mandiant/VM-Packages/issues/1173). This should be addressed as part of this issue by adding the version to the package and issue templates.

PR #1083, opened 2024-06-07
Add autoit-ripper.vm & helper functions to install tools with Pip

Ana06:
Add helper functions in `vm.common.psm1` to simplify installation of Python tools installed with pip. Related: https://github.com/mandiant/VM-Packages/issues/1080

Use the new helpers in magika.vm, also fixing the following issues:
- Fix uninstallation which wrongly used `VM-Uninstall`, as it is only  needed to remove the shortcut in the Tools category.
- Remove unused `$iconLocation` in installation. This change does not have any visible effect, but simplifies the code.

Note the version of magika has been **decreased** because it was incorrect. I have just deleted the previous package from MyGet to ensure the change is tested. **We should merge this PR quick as the package will be broken till then**.

Add [`autoit-ripper`](https://github.com/nazywam/AutoIt-Ripper) which has helped me analysing a compiled AutoIt script recently. The extracted script is similar to the one extracted with `UnAutoIt`, which seems to have been removed from GitHub.

### DIFF
--- a/packages/autoit-ripper.vm/autoit-ripper.vm.nuspec
+++ b/packages/autoit-ripper.vm/autoit-ripper.vm.nuspec
@@ -1,0 +1,13 @@
+<?xml version="1.0" encoding="utf-8"?>
+<package xmlns="http://schemas.microsoft.com/packaging/2015/06/nuspec.xsd">
+  <metadata>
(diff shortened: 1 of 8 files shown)


mr-tz (approved):
lgtm

PR #1084, opened 2024-06-07
[CI] Rename GH workflows

Ana06:
The number of workflows has increased and some of the names are too long for the interface or not very descriptive. Improve the names to make it to look nicer and easier to navigate.

### Before
image

### After
The screenshot is for my fork so it is missing CodeQL, but I think it helps seeing why this is an improvement. :wink: 
image

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -1,4 +1,4 @@
-name: CI
+name: Test & Push
 
 on:
   push:
--- a/.github/workflows/daily.yml
+++ b/.github/workflows/daily.yml
@@ -1,4 +1,4 @@
-name: daily
+name: Daily Test
 
 on:
   workflow_dispatch:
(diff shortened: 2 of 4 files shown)


mr-tz (approved):
neat

emtuls:
Looks good except one typo that seems to be from somewhere else. Might be worth fixing it in this PR as well:
- `mixmatches` is spelled incorrectly as `mitmatches`

Ana06:
@emtuls 

> `mixmatches` is spelled incorrectly as `mitmatches`

is it `mixmatches` or `mismatches`? [The Cambridge dictionary says the second is correct](https://dictionary.cambridge.org/dictionary/english/mismatch), which means we only have to change it in the action name..

mr-tz (approved):

emtuls:
Whoops. Even I misspelled it trying to type it out! Thank you @Ana06. :)

Issue #1074, opened 2024-06-05
VS Code Python Extension


mwilliams31:
### Details

Include the [Python extension](https://marketplace.visualstudio.com/items?itemName=ms-python.python) for VS Code. This installs  two extensions: [Pylance](https://marketplace.visualstudio.com/items?itemName=ms-python.vscode-pylance) and [Python Debugger](https://marketplace.visualstudio.com/items?itemName=ms-python.debugpy). Without the Python Debugger extension (or similar), Python scripts cannot be debugged in VS Code.

Another extension to consider including is [Jupyter](https://marketplace.visualstudio.com/items?itemName=ms-toolsai.jupyter).

Ana06:
thanks for the issue @mwilliams31! I suggest we implement two additional packages `vscode.python.vm` and `jupyter.python.vm` and add at least the first one to the FLARE-VM default configuration. @mandiant/vms opinions? @mandiant/flare-vm do you think we should add jupyter to the default configuration?

day1player:
there are some existing packages for this it looks like?
https://community.chocolatey.org/packages/vscode-pylance
https://community.chocolatey.org/packages/vscode-python
https://community.chocolatey.org/packages/vscode-jupyter

emtuls:
I don't think it hurts to have the `jupyter` one in the default as well. I know it can be useful in `vscode`, so I imagine it could be for that as well, and that's a tool that many people use when doing Python & Data type work, which FlareVM makes sense for that situation when it comes to Data Analysis on Malware.

Ana06:
@emtuls @emtuls is there something against using the community packages @day1player proposed so that we do not have to maintain the two extra packages? (I have just seen his proposal)

We should add either the community packages or the .vm packages to the flare-vm default configuration.

d35ha:
@Ana06 
I don't see a problem in  using VS marketplace directly, it enables us to have a better version control.

PR #1087, opened 2024-06-11
Add python and jupyter vscode extension packages

d35ha:
Closes https://github.com/mandiant/VM-Packages/issues/1074.

### DIFF
--- a/packages/vscode.extension.jupyter.vm/tools/chocolateyinstall.ps1
+++ b/packages/vscode.extension.jupyter.vm/tools/chocolateyinstall.ps1
@@ -1,0 +1,8 @@
+$ErrorActionPreference = 'Stop'
+Import-Module vm.common -Force -DisableNameChecking
+
+$extensionId = 'ms-toolsai.jupyter'
+$extensionVersion = '2024.6.2024060601'
+$executablePath = Join-Path ${Env:ProgramFiles} "\Microsoft VS Code\bin\code.cmd" -Resolve
+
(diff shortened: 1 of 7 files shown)


emtuls (changes requested):
I don't think we need to add all of these extensions separately. Just adding the `python extension` alone adds both `Pylance` and `Python Debugger` (https://marketplace.visualstudio.com/items?itemName=ms-python.python). :)

Also, I think we should name the packages `vscode.python.vm` and `vscode.jupyter.vm` instead, as the extension piece makes the name a bit longer than necessary.

Overall, they work great though! Just those changes and I think it'll be good to go.

d35ha:
> Also, I think we should name the packages `vscode.python.vm` and `vscode.jupyter.vm` instead, as the extension piece makes the name a bit longer than necessary.

@emtuls 
Renaming it to `vscode.<ext>.vm` would make it harder to exclude these extensions at `lint.py` (see: https://github.com/mandiant/VM-Packages/blob/aca19f0dd7238dc8c9920de8e96a88cd5a285d15/scripts/test/lint.py#L322), also it's I think it's consistent with `notepadpp.plugin.<plugin>.vm` and `ida.plugin.<plugin>.vm`.

d35ha:
@emtuls 
Updated.

Ana06 (approved):
Thanks for the work @d35ha and for the review @emtuls, LGTM!

emtuls (approved):
Looks good! Awesome job @d35ha, thank you!

Issue #1071, opened 2024-06-03
Linter idea: check file/folder name matches the package name in nuspec


Ana06:
### Details

Our linter should check the package folder and nuspec file names match the package name in nuspec to avoid naming mismatches like the fixed in https://github.com/mandiant/VM-Packages/pull/1065

PR #1089, opened 2024-06-11
Check the package name against its folder and nuspec file

d35ha:
Closes https://github.com/mandiant/VM-Packages/issues/1071.


scripts/test/lint.py
```diff
@@ -253,12 +253,27 @@ def check(self, path):
         return True
 
 
+class PackageIdNotMatchingFolderOrNuspecName(Lint):
+    name = "package ID doesn't match package folder name or nuspec file name"
+    recommendation = "make sure the package ID is the same as the package folder name and the nuspec file name"
+
+    def check(self, path):
+
```
emtuls, line 261: Minor nitpick. Could we remove the extra space between `check` and `dom` as well as `folder` and `return`?

emtuls (changes requested):
Minor change, but outside of that, it looks good! Thank you @d35ha!

scripts/test/lint.py
d35ha: Updated

scripts/test/lint.py
```diff
@@ -253,12 +253,26 @@ def check(self, path):
         return True
 
 
+class PackageIdNotMatchingFolderOrNuspecName(Lint):
+    name = "package ID doesn't match package folder name or nuspec file name"
+    recommendation = "make sure the package ID is the same as the package folder name and the nuspec file name"
+
+    def check(self, path):
+        dom = minidom.parse(str(path))
+        pid = dom.getElementsByTagName("id")[0].firstChild.nodeValue
```
Ana06, line 262: What about using a more readable name? we use in this file `pkg_id` for the same. Also `nodeValue` is an alias for `data`, I would use `data` to keep it consistent with the rest of the code.
```suggestion
        pkg_id = dom.getElementsByTagName("id")[0].firstChild.data
```

Ana06 (commented):
This looks great, thanks @d35ha! Just a minor improvement idea.

Ana06 (approved):
Great work!

emtuls (approved):

Issue #965, opened 2024-04-01
Add Windows Time Travel Debugging Cli


schrodyn:
### Details

Microsoft ships a standalone utility to collect time travel debugging traces. Information is available here https://learn.microsoft.com/en-us/windows-hardware/drivers/debuggercmds/time-travel-debugging-ttd-exe-command-line-util

This allows a person to collect a trace without the WinDBG GUI. I think this would be a great addition to the FLARE-VM builds.

The documentation provides a Powershell script that could be easily integrated into the FLARE-VM builds. https://learn.microsoft.com/en-us/windows-hardware/drivers/debuggercmds/time-travel-debugging-ttd-exe-command-line-util#download-the-ttdexe-command-line-utility-package-and-extract-the-files-manually

Ana06:
I think this would be a great addition, thank you for opening this issue!

It looks like we do not have a link which includes the version. Which means the download will change and we can't use the hash for the installation. We would need to use the `VM-Assert-Signature` as we do in sysinternals and googlechrome, which is a bit more elaborate as we can't use automation to create the packages in this case. @schrodyn a PR would be much appreciated.

Do we know which version the linked PorwerShell script uses? We would need to check we can use that code if we need to use that script.

schrodyn:
Not sure if this helps. There is a versioned msxi bundle that can be downloaded, the URL for it lives inside the TTP.appinstaler file. This could be downloaded and the file TTD-x64.msix extracted from it and instaleld by FLARE-VM.

# Downloading TTD.appinstaller
```
[quack (13:43) Windows]
➜ wget https://aka.ms/ttd/download
--2024-04-04 13:44:36--  https://aka.ms/ttd/download
Resolving aka.ms (aka.ms)... 23.54.202.151
Connecting to aka.ms (aka.ms)|23.54.202.151|:443... connected.
HTTP request sent, awaiting response... 301 Moved Permanently
Location: https://windbg.download.prss.microsoft.com/dbazure/prod/1-0-0/TTD.appinstaller [following]
--2024-04-04 13:44:37--  https://windbg.download.prss.microsoft.com/dbazure/prod/1-0-0/TTD.appinstaller
Resolving windbg.download.prss.microsoft.com (windbg.download.prss.microsoft.com)... 152.199.21.175
Connecting to windbg.download.prss.microsoft.com (windbg.download.prss.microsoft.com)|152.199.21.175|:443... connected.
HTTP request sent, awaiting response... 200 OK
[...]

PR #1092, opened 2024-06-13
Add the Time Travel Debugging CLI package

d35ha:
Closes https://github.com/mandiant/VM-Packages/issues/965.

### DIFF
--- a/packages/ttd.vm/tools/chocolateyinstall.ps1
+++ b/packages/ttd.vm/tools/chocolateyinstall.ps1
@@ -1,0 +1,26 @@
+$ErrorActionPreference = 'Stop'
+Import-Module vm.common -Force -DisableNameChecking
+
+try {
+    $toolName = "ttd"
+    $category = "Debuggers"
+
+    # From https://aka.ms/ttd/download
+    $bundleUrl = "https://windbg.download.prss.microsoft.com/dbazure/prod/1-11-319-0/TTD.msixbundle"
+    $bundleSha256 = "f7b80731c3a6994b3763c4100073b101965327d6556fa4bfb553d70ce49be366"
+
+    $packageArgs = @{
+        packageName   = ${Env:ChocolateyPackageName}
+        url           = $bundleUrl
+        checksum      = $bundleSha256
+        checksumType  = "sha256"
+        fileFullPath  = Join-Path ${Env:TEMP} "$toolName.msixbundle"
+    }
+
+    Get-ChocolateyWebFile @packageArgs
+    Add-AppxPackage -Path $packageArgs.fileFullPath
+
+    VM-Install-Shortcut -toolName $toolName -category $category -executablePath "$toolName.exe" -consoleApp $true -runAsAdmin
+} catch {
+    VM-Write-Log-Exception $_
+}
--- a/packages/ttd.vm/tools/chocolateyuninstall.ps1
+++ b/packages/ttd.vm/tools/chocolateyuninstall.ps1
@@ -1,0 +1,9 @@
+$ErrorActionPreference = 'Continue'
+Import-Module vm.common -Force -DisableNameChecking
+
+$toolName = "ttd"
+$category = "Debuggers"
+
+VM-Remove-Tool-Shortcut $toolName $category
+
+Get-AppxPackage *TimeTravelDebugging* | Remove-AppPackage
--- a/packages/ttd.vm/ttd.vm.nuspec
+++ b/packages/ttd.vm/ttd.vm.nuspec
@@ -1,0 +1,12 @@
+<?xml version="1.0" encoding="utf-8"?>
+<package xmlns="http://schemas.microsoft.com/packaging/2015/06/nuspec.xsd">
+  <metadata>
+    <id>ttd.vm</id>
+    <version>1.11.319.20240614</version>
+    <authors>Microsoft</authors>
+    <description>Time travel debugging command line utility.</description>
+    <dependencies>
+      <dependency id="common.vm" />
+    </dependencies>
+  </metadata>
+</package>


emtuls:
Installs correctly and works well! 

I want to make sure @Ana06 checks it over to agree that it will work with this method rather than needing to use `VM-Assert-Signature` like she had suggested here: https://github.com/mandiant/VM-Packages/issues/965#issuecomment-2034160190

If this URL format works for future updates, then I think it should be good to go. :)

packages/ttd-cli.vm/ttd-cli.vm.nuspec
```diff
@@ -0,0 +1,12 @@
+<?xml version="1.0" encoding="utf-8"?>
+<package xmlns="http://schemas.microsoft.com/packaging/2015/06/nuspec.xsd">
+  <metadata>
+    <id>ttd-cli.vm</id>
+    <version>1.11.319</version>
```
Ana06, line 5: [nitpick] It look like the tool version uses the 4th segment, so strictly following [our documentation](https://github.com/mandiant/VM-Packages/wiki/Chocolatey-packages#non-metapackages), we should add the date as a forth segment. As the tool forth segment is zero, I have no strong opinion here.
```suggestion
    <version>1.11.319.20240614</version>
```

packages/ttd-cli.vm/ttd-cli.vm.nuspec
```diff
@@ -0,0 +1,12 @@
+<?xml version="1.0" encoding="utf-8"?>
+<package xmlns="http://schemas.microsoft.com/packaging/2015/06/nuspec.xsd">
+  <metadata>
+    <id>ttd-cli.vm</id>
```
Ana06, line 4: [nitpick] Why not using just the tool name: `ttd.vm`?

packages/ttd-cli.vm/tools/chocolateyinstall.ps1
```diff
@@ -0,0 +1,37 @@
+$ErrorActionPreference = 'Stop'
+Import-Module vm.common -Force -DisableNameChecking
+
+try {
+    $toolName = "ttd"
+    $category = "Debuggers"
+    $toolVersion = "1-11-319-0"
```
Ana06, line 7: Using the version in an env variable break automatic updates. We removed some of these variables in https://github.com/mandiant/VM-Packages/pull/1053.

packages/ttd-cli.vm/tools/chocolateyinstall.ps1
```diff
@@ -0,0 +1,37 @@
...
+        unzipLocation  = $tempDownloadDir
+        url            = $zipUrl
+        checksum       = $zipSha256
+        checksumType   = 'sha256'
+    }
+    Install-ChocolateyZipPackage @packageArgs | Out-Null
+    VM-Assert-Path $tempDownloadDir
+
+    # Get the unzipped x64 msix (also there are "TTD-ARM64.msix" and "TTD-x86.msix")
+    $unzippedInstaller = Join-Path $tempDownloadDir "TTD-x64.msix" -Resolve
```
Ana06, line 28: You don't need to unzip and select the version manually, you can use `Add-AppxPackage` with the `msixbundle` directly. The code of this package should look very similar to https://github.com/mandiant/VM-Packages/pull/1058 which we haven't merged because of an issue in the CI. This also shows that the introduction of a helper function for these kind of packages would be useful.

packages/ttd-cli.vm/tools/chocolateyinstall.ps1
```diff
@@ -0,0 +1,37 @@
...
+    Add-AppxPackage -Path $unzippedInstaller
+
+    # Attempt to remove temporary directory
+    Remove-Item $tempDownloadDir -Recurse -Force -ea 0
+
+    VM-Install-Shortcut -toolName $toolName -category $category -executablePath "$toolName.exe" -consoleApp $true -runAsAdmin
```
Ana06, line 34: What about adding an icon? :smile:

Ana06 (changes requested):

packages/ttd-cli.vm/ttd-cli.vm.nuspec
d35ha: I think it would be too short for a package name.

packages/ttd-cli.vm/tools/chocolateyinstall.ps1
d35ha: This tool doesn't have an icon, it's using the default executable icon.

packages/ttd-cli.vm/tools/chocolateyinstall.ps1
d35ha: Done.

packages/ttd-cli.vm/ttd-cli.vm.nuspec
d35ha: Done.

packages/ttd-cli.vm/tools/chocolateyinstall.ps1
d35ha: Done.

packages/ttd-cli.vm/ttd-cli.vm.nuspec
emtuls: Doesn't matter which way we go with it for me, but `ttd` alone would be okay imo. We already have other short names like `die`, `c3`, `rla`, `upx`, `map`, `isd`, `file`, `nasm`, `nmap`, etc

d35ha: @emtuls 
Done.

packages/ttd.vm/tools/chocolateyinstall.ps1
```diff
@@ -0,0 +1,24 @@
...
+
+try {
+    $toolName = "ttd"
+    $category = "Debuggers"
+
+    # From https://aka.ms/ttd/download
+    $bundleUrl = "https://windbg.download.prss.microsoft.com/dbazure/prod/1-11-319-0/TTD.msixbundle"
+    $bundleSha256 = "f7b80731c3a6994b3763c4100073b101965327d6556fa4bfb553d70ce49be366"
+    $bundlePath = Join-Path ${Env:TEMP} "$($toolName)_1-11-319-0.msixbundle"
+    Get-ChocolateyWebFile -PackageName $toolName -FileFullPath $bundlePath -Url $bundleUrl -Checksum $bundleSha256 -ChecksumType "sha256"
```
Ana06, line 12: [nitpick] I would prefer to use the `$packageArgs = @{}` syntax for consistency with other packages

Ana06 (approved):

Ana06:
@naacbin can you take a look at this PR? It is similar to windbg that you are refactoring in @1058, so ideally the code of the th two packages should look very similar.

packages/ttd.vm/tools/chocolateyinstall.ps1
```diff
@@ -0,0 +1,24 @@
...
+Import-Module vm.common -Force -DisableNameChecking
+
+try {
+    $toolName = "ttd"
+    $category = "Debuggers"
+
+    # From https://aka.ms/ttd/download
+    $bundleUrl = "https://windbg.download.prss.microsoft.com/dbazure/prod/1-11-319-0/TTD.msixbundle"
+    $bundleSha256 = "f7b80731c3a6994b3763c4100073b101965327d6556fa4bfb553d70ce49be366"
+    $bundlePath = Join-Path ${Env:TEMP} "$($toolName)_1-11-319-0.msixbundle"
```
naacbin, line 11: I don't think it's necessary to specify the version here. `Join-Path ${Env:TEMP} "$toolName.msixbundle"` is enough.

packages/ttd.vm/tools/chocolateyinstall.ps1
naacbin: In #1058 I have done it like below :
```powershell
$packageArgs = @{
    packageName   = ${Env:ChocolateyPackageName}
    url           = $bundleUrl
    checksum      = $bundleSha256
    checksumType  = "sha256"
    fileFullPath  = Join-Path ${Env:TEMP} "$toolName.msixbundle"
}

Get-ChocolateyWebFile @packageArgs
Add-AppxPackage -Path $packageArgs.fileFullPath
```

packages/ttd.vm/tools/chocolateyinstall.ps1
Ana06: I agree, it will need less changes when updating

d35ha: @naacbin Done.

packages/ttd.vm/tools/chocolateyinstall.ps1
d35ha: @naacbin Done.

Ana06 (approved):
Thanks for the work @d35ha!

Issue #973, opened 2024-04-03
Change new tabs default in Windows terminal to CMD


Ana06:
### Details

The Windows Terminal defaults to CMD when you open it. The default to create a new tab is PowerShell. Is it possible to change the default? And if it is, do we want to change it to CMD? @mandiant/vms  :eyes: 

Related https://github.com/mandiant/VM-Packages/issues/971

emtuls:
This is unfortunately not possible for Windows 10 in an automated manner at the moment. It can be done manually after installing Windows Terminal, but that isn't what we are aiming for.

See https://github.com/mandiant/VM-Packages/pull/1093 for more information.

Issue #971, opened 2024-04-02
Make Windows Terminal the default to run console apps


Ana06:
### Details

Since https://github.com/mandiant/flare-vm/pull/580 and https://github.com/mandiant/VM-Packages/pull/962, Windows Terminal is the default console in FLARE-VM, being adding to the right-click and to the taskbar. @mandiant/vms should we also make Windows Terminal the default console to run tools as console apps? Currently the default is cmd for most tool and powershell for ps1 scripts.

d35ha:
@Ana06 
This feature is only [available in windows 11 or an updated version of windows 10](https://learn.microsoft.com/en-us/windows/terminal/install#:~:text=This%20feature%20is%20available%20in%20all%20versions%20of%20Windows%2011%20and%20versions%20of%20Windows%2010%2022H2%20after%20the%20installation%20of%20the%20May%2023%2C%202023%20update%2C%20KB5026435.).
It's not available in the currently enrolled version (as of `FLARE-VM.20240604`).
If available, it can be done via the [Windows terminal >=1.12 settings](https://devblogs.microsoft.com/commandline/windows-terminal-preview-1-12-release/) or the [traditional terminal settings](https://www.makeuseof.com/set-reset-default-terminal-app-windows/#:~:text=2.%20How%20to%20Change%20the%20Default%20Terminal%20App%20Using%20Windows%20Console%20Host%20Properties).

Ana06:
What do you mean with "this feature"? We want to open the shortcuts in the Tools directory with the Windows terminal, this is something we should be able to do in all versions.

I am using `10.0.19045.0` (downloaded from the recommended link in FLARE-VM), so we need to support this version as well.

d35ha:
@Ana06 
The default terminal host in windows is `conhost.exe`, for any application that requires a console (Ex. console apps like `cmd.exe` or GUI apps using [AllocConsole](https://learn.microsoft.com/en-us/windows/console/allocconsole)) windows, by default, assigns the traditional console host to it.
Since our shortcuts are basically wrapper for `cmd.exe` commands, the used console is the old `conhost.exe`. `cmd.exe` itself is not a console, it's an app that requires a console.
To change the default console host in windows, the OS build should >= `19045.3031` .

emtuls:
This is unfortunately not possible for Windows 10 in an automated manner at the moment. It can be done manually after installing Windows Terminal, but that isn't what we are aiming for.

[...]

PR #1093, opened 2024-06-14
Set the Windows Terminal as the default terminal

d35ha:
Closes https://github.com/mandiant/VM-Packages/issues/971.
Closes https://github.com/mandiant/VM-Packages/issues/973.

### DIFF
--- a/packages/windows-terminal.vm/tools/chocolateyinstall.ps1
+++ b/packages/windows-terminal.vm/tools/chocolateyinstall.ps1
@@ -40,6 +40,7 @@ try {
   $defaultSettings = @"
 {
   "`$schema": "https://aka.ms/terminal-settings-schema-v1.1",
+  "defaultProfile": "Command Prompt",
   "profiles": {
     "defaults": {
       "elevate": true
--- a/packages/windows-terminal.vm/tools/chocolateyinstall.ps1
+++ b/packages/windows-terminal.vm/tools/chocolateyinstall.ps1
@@ -54,6 +54,12 @@ try {
   $label = "Open Terminal here"
   $icon = "$executablePath"
   VM-Add-To-Right-Click-Menu -menuKey $toolName -menuLabel $label -command $command -menuIcon $icon -type "directory" -background
+
+  # Set windows terminal as the default terminal (effective only when OS Build >= 19045.3031)
+  $registryPath = 'HKCU:\Console\%%Startup'
+  New-Item $registryPath -Force | Out-Null
+  Set-ItemProperty $registryPath -Name "DelegationConsole" -Value "{2EACA947-7F5F-4CFA-BA87-8F7FBEEFBE69}" -Force | Out-Null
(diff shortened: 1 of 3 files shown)


Ana06:
I am not sure this works as proposed in #971. @d35ha can you provide more details or how this works and how it looks like (maybe adding a screenshot)?

d35ha:
@Ana06 
I've created a console app, here is it how its execution looks before and after setting these registry values:

before

after

Also, running `cmd` or `powershell` opens it in Windows Terminal.

Ana06:
@d35ha is proposing using the windows terminal to open the shortcuts in the Tools directory, is this PR achieving this?

d35ha:
@Ana06 
Any application that requires a console would be assigned the Windows terminal console instead of the the old conhost.exe (that is including the shortcuts under the Tools directory).

packages/windows-terminal.vm/tools/chocolateyinstall.ps1
```diff
@@ -54,6 +55,12 @@ try {
   $label = "Open Terminal here"
   $icon = "$executablePath"
   VM-Add-To-Right-Click-Menu -menuKey $toolName -menuLabel $label -command $command -menuIcon $icon -type "directory" -background
+
+  # Set windows terminal as the default terminal (effective only when OS Build >= 19045.3031)
```
Ana06, line 59: I am using `10.0.19045.0` (downloaded from the recommended link in FLARE-VM), so this approach does not work

Ana06 (changes requested):

emtuls:
I looked into this a little bit more.

I wanted to try and find a way to have any "console app" shortcut open with `Windows Terminal` (if it exists) or `cmd.exe` if not, but the way to make that happen would likely require an external script to launch the correct one. I tried creating a singular command that can be done within a shortcut target cmd, but it ended up being too long.

An alternative would be that we force `Windows Terminal` to be installed, but I believe this would require modification of our install script so that it is always installed if I'm not mistaken?

Let me know if that's the preferred route and I can work on that.

emtuls:
I got something mostly working for Command Prompt, but it seems like Powershell is going to be a lot more of an issue due to a long standing bug with Windows Terminal and how it interprets semicolons: https://github.com/microsoft/terminal/issues/13264

This makes it very difficult, if not impossible to do our powershell shortcuts. I was able to make one work, but it would require removing the "Write-Host" portion (echoing the command we write into the console).

For reference, I modified `powersploit` to use this as a shortcut target: `"C:\Tools\Windows Terminal\terminal-1.19.10573.0\wt.exe" -p "PowerShell" "powershell.exe -NoExit -Command ""Get-Command -Module (Import-Module .\Recon\Recon.psd1)""`

This still seems to work as expected, just a re-write of the original command we used: `Import-Module .\Recon\Recon.psd1 ; Get-Command -Module Recon`

Ana06:
Thanks for the research @emtuls! I suggest we merge this PR as it is, as it seems it is not easy to support it in our Windows version.

Ana06 (approved):
I think it is nice that previous versions benefit from the improvement even if it does not work in older version. I have tested it in `10.0.19045.0` and it does not seem to introduce any issues (it also does modify the behavior).

PR #1094, opened 2024-06-18
Update lecmd

naacbin:


Ana06 (approved):
Almost no failures in the daily run! thanks @naacbin!!! :rocket:

PR #1076, opened 2024-06-05
:robot: Package update

vm-packages:
Automated package update

PR #1098, opened 2024-06-26
Fix missing dependency in hashdb plugin

emtuls:
This fixes the issue noted [here](https://github.com/mandiant/VM-Packages/pull/1076#issuecomment-2190501954) about the ida plugin `hashdb` missing a dependency for `python3` since it uses `pip` to install `requests`.

### DIFF
--- a/packages/ida.plugin.hashdb.vm/ida.plugin.hashdb.vm.nuspec
+++ b/packages/ida.plugin.hashdb.vm/ida.plugin.hashdb.vm.nuspec
@@ -2,11 +2,12 @@
 <package xmlns="http://schemas.microsoft.com/packaging/2015/06/nuspec.xsd">
   <metadata>
     <id>ida.plugin.hashdb.vm</id>
-    <version>1.9.1</version>
+    <version>1.9.1.20240526</version>
     <authors>OALabs</authors>
     <description>Malware string hash lookup plugin for IDA Pro</description>
     <dependencies>
       <dependency id="common.vm" version="0.0.0.20240509" />
+      <dependency id="libraries.python3.vm" />
     </dependencies>
   </metadata>
 </package>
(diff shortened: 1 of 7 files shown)


packages/ida.plugin.hashdb.vm/ida.plugin.hashdb.vm.nuspec
```diff
@@ -2,11 +2,12 @@
 <package xmlns="http://schemas.microsoft.com/packaging/2015/06/nuspec.xsd">
   <metadata>
     <id>ida.plugin.hashdb.vm</id>
-    <version>1.9.1</version>
+    <version>1.9.1.20240526</version>
     <authors>OALabs</authors>
     <description>Malware string hash lookup plugin for IDA Pro</description>
     <dependencies>
       <dependency id="common.vm" version="0.0.0.20240509" />
+      <dependency id="python3.vm" />
```
Ana06, line 10: I think we should use `libraries.python3.vm` instead to ensure there are no dependency issues.

Ana06 (changes requested):
Thanks for fixing this @emtuls! 

[nitpick] I think we should add `libraries.python3.vm` as dependency to the rest of our Python IDA plugins (I think that's all of them except `ida.plugin.sigmaker.vm`) even if they do not use `pip` to install dependencies. Can you address that as well @emtuls

packages/ida.plugin.hashdb.vm/ida.plugin.hashdb.vm.nuspec
emtuls: Sure thing! Updated. :)

Ana06 (approved):
> [nitpick] I think we should add libraries.python3.vm as dependency to the rest of our Python IDA plugins (I think that's all of them except ida.plugin.sigmaker.vm) even if they do not use pip to install dependencies. Can you address that as well @emtuls

I am ok with fixing this here or in a separate PR. @emtuls are you planing to fix it for the other packages as well? If you do not have time, can you at least create an issue that we do not forget it?

Ana06 (approved):

Issue #1101, opened 2024-07-04
Open with CFF explorer is missing icon


Ana06:
### Details

Open with CFF explorer is missing icon:
![image](https://github.com/mandiant/VM-Packages/assets/16052290/b34073e5-1b03-4211-9c71-06e32ddf2dbc)

PR #1108, opened 2024-07-17
Add the CFF Explorer icon to the right click & fix uninstallation

Ana06:
Add the CFF Explorer icon to the right click menu option. Installing CFF Explorer adds it to the right click menu for several file extensions (without icon). Delete the registry subkeys for the concrete file extensions to ensure the one we create for all extensions applies.

image

Closes https://github.com/mandiant/VM-Packages/issues/1101

When uninstalling CFF Explorer, the right click menu option was not being removed because the
wrong name was being used. Correct the name to fix it.


binjo (approved):
LGTM

Issue #1086, opened 2024-06-10
Issue with strings


Ana06:
### Details

After installing FLARE-VM, the first `strings` in path is not the sysinternals strings, but:
![image](https://github.com/mandiant/VM-Packages/assets/16052290/d8c4c384-d5ea-44a3-b453-41b80ed0e043)

I had expected the sysinternals one. Should we add the sysinternals path at the beginning of PATH in the package? Are there any better solutions? @vm-packages

Also, should we add a shortcut for `strings` in the Tools directory?

mr-tz:
I'd also expect `strings.exe`. I don't think we need a shortcut (there are other strings tools integrated).

PR #1109, opened 2024-07-17
Ensure strings from sysinternals is the first in PATH

Ana06:
I expect the strings from sysinternals to be the first one in the PATH (and not others like the ones in DidierStevensSuite). Use `Install-BinFile` to creates a shim for `strings.exe` in `$env:ChocolateyInstall\bin` that we ensure is on the top of the PATH after installing Python.

Closes https://github.com/mandiant/VM-Packages/issues/1086

image


binjo (approved):
LGTM

PR #1113, opened 2024-07-26
[libraries.python3.vm] Add pycryptodomex library

Ana06:
Add `pycryptodomex` Python library, the new PyCryptodome. The old PyCrypto and PyCryptodome can coexist, keep the old one for better compatibility.


emtuls (approved):
Looks good! 

I was slightly confused at first with how the repo read in regards to which libraries can coexist, but this seems to be the correct way. :)

Thanks for the addition!

PR #1115, opened 2024-08-09
Added static download link for NetworkMiner 2.9

erik4711:
Also added a date suffix to the version in nuspec to make sure the new zipUrl update in chocolateyinstall.ps1 gets processed


emtuls (approved):
Awesome. Looks good!

Thank you @erik4711!

PR #1119, opened 2024-08-21
[common] Add VM-Set-Legal-Notice helper function

Ana06:
Add `VM-Set-Legal-Notice` helper function to set the legal notice to the desired text. The legal notice caption used is `Terms and Conditions`.


emtuls (approved):
Looks good! Thanks for the added function. Will make it easier to automate things!

PR #1121, opened 2024-08-24
Add internet detector utility

emtuls:
This adds a script to detect if internet connection is active or not by adding an icon in the taskbar and also changing the background based on if internet is detected or not.

This will also require an addition of a new python libraries to install:
`pywin32`


scripts/utils/internet_detect_tool.py
```diff
@@ -0,0 +1,246 @@
...
+CONNECT_TEST_URL = "https://www.msftconnecttest.com/connecttest.txt"
+EXPECTED_RESPONSE = "Microsoft Connect Test"
+SPI_SETDESKWALLPAPER = 20
+SPIF_UPDATEINIFILE = 0x01
+SPIF_SENDWININICHANGE = 0x02
+COLOR_DESKTOP = 1
+ICON_RED_X = "red_circle.ico"
+ICON_FLARE = "flare.ico"
+DEFAULT_BACKGROUND = os.path.join(os.environ.get('VM_COMMON_DIR'), "background.png")
+INTERNET_BACKGROUND = os.path.join(os.environ.get('VM_COMMON_DIR'),"background-internet.png")
```
Ana06, line 28: We need to add this image to flare-vm and copy as we do with the background image.

scripts/utils/internet_detect_tool.py
```diff
@@ -0,0 +1,246 @@
...
+
+# Define constants
+CHECK_INTERVAL = 2  # Seconds
+CONNECT_TEST_URL = "https://www.msftconnecttest.com/connecttest.txt"
+EXPECTED_RESPONSE = "Microsoft Connect Test"
+SPI_SETDESKWALLPAPER = 20
+SPIF_UPDATEINIFILE = 0x01
+SPIF_SENDWININICHANGE = 0x02
+COLOR_DESKTOP = 1
+ICON_RED_X = "red_circle.ico"
```
Ana06, line 25: where are the icons stored?

scripts/utils/internet_detect_tool.py
```diff
@@ -0,0 +1,246 @@
...
+ICON_FLARE = "flare.ico"
```
Ana06, line 26: @mandiant/commando-vm is there interest in commando VM to use a tool like this? If so, we should keep it generic.

scripts/utils/internet_detect_tool.py
```diff
@@ -0,0 +1,246 @@
...
+import struct
+import signal
+import ctypes
+import time
+import os
+
+
+# Define constants
+CHECK_INTERVAL = 2  # Seconds
+CONNECT_TEST_URL = "https://www.msftconnecttest.com/connecttest.txt"
```
Ana06, line 19: I think we should have a hash with TEST_URLs and their EXPECTED_RESPONSE, as if this concrete URL is down, the internet test would fail.

Ana06: what about testing the adapter in addition?

scripts/utils/internet_detect_tool.py
```diff
@@ -0,0 +1,246 @@
...
+tray_icon = None
+stop_event = threading.Event()  # To signal the background thread to stop
+hwnd = None     # We'll assign the window handle here later
+# Win32 API icon handles
+hicon_flare = None
+hicon_red_x = None
+
+def signal_handler(sig, frame):
+    print("Ctrl+C detected. Resetting background and exiting...")
+    reset_background()
```
Ana06, line 41: I know this was my idea, but I am thinking now it may be a bad idea:  If internet is enabled and you stop the tool, the internet is still enabled. :thinking:

mr-tz:
How does this script interact with FakeNet? Should FakeNet be handled specifically?

Ana06:
> How does this script interact with FakeNet? Should FakeNet be handled specifically?

If fakenet response is the one we would get with internet, our tool (as other programs like malware) thinks the internet is on. The default fakenet response shouldn't match the ones we are expecting, so without modifying the default fakenet configuration, it will think there is no internet.

@mr-tz do you have a proposal to improve the implementation? What do you exactly mean with handling fakenet specifically?

Note: I think this tool could be hosted in the fakenet repository and even been part of fakenet, it would need to be discussed with the fakenet maintainers to see if it is feasible once we have a working and tested version.

mr-tz:
I'd expect it to not warn users that the Internet is reachable if FakeNet is running.

emtuls:
This raises a good thought.

While the tool works with FakeNet active, you would see a lot of requests generated in order for the tool to function. We can potentially reduce the request interval, but this would cause a delay in detecting internet which is a trade-off we need to consider.

What interval would be fair enough to detect a real internet connection in enough time for a user to be aware while also not being too intrusive in a FakeNet output? Also note, due to adding backup internet connectivity URL's, there are now multiple requests that are issued consecutively, rather than a single one.

scripts/utils/internet_detect_tool.py
emtuls: Whoops! Thank you for catching that. Added here: https://github.com/mandiant/flare-vm/pull/608

emtuls:
I was able to remove the `PyQt5` dependency. :)

emtuls:
Added nicer tooltips that tell explain what the icons mean: "Internet Detected" vs "Internet Not Detected"
image
image

mr-tz:
I remember some filtering work FakeNet did on Internet connectivity checks to reduce noise. Maybe @tinajohnson can chime in on this?

I'm not sure what a good interval would be, maybe 20 seconds?

emtuls:
Chatting with FakeNet team, it sounds like we can add the tool to "Diverter's ProcessBlackList and HTTPListener's ProcessBlackList": 
see the second paragraph here ("You may also want to enable..." : https://github.com/mandiant/flare-fakenet-ng?tab=readme-ov-file#anonymous-listener

emtuls:
I also updated the tool to use a pink icon when internet is active and gray when not, as well as convert everything over to a VM Package and have it install a scheduled task to run every 2 minutes.

It is a `.pyw` file in order to not have a terminal interface popup for it to run.
image
image

emtuls:
Also note, this will fail to pass the `test_upload` due to requiring certain files to be existent in the main branch already:
https://github.com/mandiant/VM-Packages/pull/1121/files#diff-bb3df9f8d616b7a9b00f92220dafa6cc20883cdd518f04b8e2c3e3993dfe9fe6R15

Ana06:
@emtuls 

> Also note, this will fail to pass the test_upload due to requiring certain files to be existent in the main branch already:
https://github.com/mandiant/VM-Packages/pull/1121/files#diff-bb3df9f8d616b7a9b00f92220dafa6cc20883cdd518f04b8e2c3e3993dfe9fe6R15

We need to split this PR as the package need to succeed to be pushed to MyGet: one PR only for the tool, one PR for the package. Sorry for the confusion.

packages/internet_detector.vm/tools/chocolateyinstall.ps1
```diff
@@ -0,0 +1,30 @@
...
+$toolName = 'internet_detector'
+$category = 'Utilities'
+
+# Install dependency for windows api
+VM-Pip-Install "pywin32"
+
+$toolDir = "${Env:RAW_TOOLS_DIR}\$toolName"
+if (-not (Test-Path $toolDir)) {  # Check if the directory doesn't exist
+    New-Item -Path $toolDir -ItemType Directory -Force
+}
```
Ana06, line 13:
```suggestion
New-Item -Path $toolDir -ItemType Directory -Force -ea 0
```

packages/internet_detector.vm/tools/chocolateyinstall.ps1
```diff
@@ -0,0 +1,30 @@
...
+
+# Download the script
+(New-Object net.webclient).DownloadFile('https://raw.githubusercontent.com/mandiant/VM-Packages/main/packages/internet_detector.vm/tools/internet_detector.pyw', $toolDir)
```
Ana06, line 16: use it directly is inside the package

packages/internet_detector.vm/tools/chocolateyinstall.ps1
```diff
@@ -0,0 +1,30 @@
...
+VM-Install-Shortcut $toolName $category $executablePath -arguments $filePath
+
+# Define the action to be performed
+$action = New-ScheduledTaskAction -Execute $executablePath -Argument "$filePath"
+
+# Define the trigger to run every 2 minutes
+$trigger = New-ScheduledTaskTrigger -Once -At (Get-Date) -RepetitionInterval (New-TimeSpan -Minutes 2)
+
+# Create the scheduled task
+Register-ScheduledTask -Action $action -Trigger $trigger -TaskName 'Internet Detector' -Force
```
Ana06, line 30:
```suggestion
# Define the action to be performed in the scheduled task
$action = New-ScheduledTaskAction -Execute $executablePath -Argument "$filePath"
# Define the trigger to run every 2 minutes
$trigger = New-ScheduledTaskTrigger -Once -At (Get-Date) -RepetitionInterval (New-TimeSpan -Minutes 2)
# Create the scheduled task
Register-ScheduledTask -Action $action -Trigger $trigger -TaskName 'Internet Detector' -Force
```

packages/internet_detector.vm/tools/chocolateyuninstall.ps1
```diff
@@ -0,0 +1,7 @@
+$ErrorActionPreference = 'Continue'
+Import-Module vm.common -Force -DisableNameChecking
+
+$toolName = 'internet_detector'
+$category = 'Utilities'
+
+VM-Uninstall $toolName $category
```
Ana06, line 7: I think we should remove the scheduled task and maybe the toolDir as well

packages/internet_detector.vm/tools/internet_detector.pyw
```diff
@@ -0,0 +1,239 @@
...
+                    win32con.WM_USER + 20, self.icon, self.tooltip)
+        win32gui.Shell_NotifyIcon(win32gui.NIM_MODIFY, self.nid)
+
+    def __del__(self):
+        # Remove the icon from the system tray when the object is destroyed
+        win32gui.Shell_NotifyIcon(win32gui.NIM_DELETE, self.nid)
+
+# Attempt to extract a known good value in response.
+def extract_title(data):
+    match = re.search(r'<title>(.*?)</title', data)
```
Ana06, line 91:
```suggestion
    match = re.search(r'<title>(.*?)</title>', data)
```

packages/internet_detector.vm/tools/internet_detector.pyw
```diff
@@ -0,0 +1,239 @@
+import threading
+import requests
+import win32api
+import win32gui
+import win32con
+import urllib3
+import hashlib
```
Ana06, line 7:
```suggestion
```

packages/internet_detector.vm/tools/internet_detector.pyw
```diff
@@ -0,0 +1,239 @@
...
+    "https://www.msftconnecttest.com/connecttest.txt": "Microsoft Connect Test", # HTTPS Test #1
+    "http://www.google.com": 'Google',          # HTTP Test
+    "https://www.wikipedia.com": 'Wikipedia',   # HTTPS Test #2
+    "https://www.youtube.com": 'YouTube'        # HTTPS Test #3
+    }
+SPI_SETDESKWALLPAPER = 20
+SPIF_UPDATEINIFILE = 0x01
+SPIF_SENDWININICHANGE = 0x02
+COLOR_DESKTOP = 1
+ICON_INDICATOR_ON = os.path.join(os.environ.get('VM_COMMON_DIR'), "indicator_on.ico")
```
Ana06, line 28: add icons + `background-internet.png`

packages/internet_detector.vm/tools/internet_detector.pyw
```diff
@@ -0,0 +1,239 @@
...
+ICON_INDICATOR_OFF = os.path.join(os.environ.get('VM_COMMON_DIR'), "indicator_off.ico")
+DEFAULT_BACKGROUND = os.path.join(os.environ.get('VM_COMMON_DIR'), "background.png")
+INTERNET_BACKGROUND = os.path.join(os.environ.get('VM_COMMON_DIR'),"background-internet.png")
+
+# Global variables
+tray_icon = None
+stop_event = threading.Event()  # To signal the background thread to stop
```
Ana06, line 35: [nitpick] run `black -l 120` for consistent styling

packages/internet_detector.vm/tools/internet_detector.pyw
```diff
@@ -0,0 +1,239 @@
...
+    else:
+        return None
+
+def check_internet():
+    for url, expected_response in CONNECT_TEST_URL_AND_RESPONSES.items():
+        try:
+            # Perform internet connectivity tests
+            response = requests.get(url, timeout=5, verify=False)
+            if expected_response in (extract_title(response.text) or response.text):
+                print("Internet connectivity detected via URL: {}".format(url))
```
Ana06, line 103: [nitpick] Use the most important Python >3.6 future (f-strings):
```suggestion
                print(f"Internet connectivity detected via URL: {url}")
```

Ana06 (changes requested):

packages/internet_detector.vm/tools/chocolateyuninstall.ps1
emtuls: toolDir should be removed via `VM-Uninstall` I believe? https://github.com/mandiant/VM-Packages/blob/921ab89f37a2513fa77571f695415b2160970b03/packages/common.vm/tools/vm.common/vm.common.psm1#L607

tinajohnson:
Apologies for missing the ping here but I did have a conversation about using FakeNet's blacklist config with @emtuls privately. A correction to my original comment on the same - adding the process name to just the Diverter's ProcessBlackList config should do the trick as that will act as a global blacklist.

packages/internet_detector.vm/tools/internet_detector.pyw
```diff
@@ -0,0 +1,238 @@
...
+    tray_icon = SysTrayIcon(hwnd, hicon_indicator_off, "Internet Detector")
+    
+    while not stop_event.is_set():
+        msg = win32gui.PeekMessage(hwnd, 0, 0, 0)
+        if msg and len(msg) == 6:
+            win32gui.TranslateMessage(msg)
+            win32gui.DispatchMessage(msg)
+        time.sleep(0.1)
+
+def enable_transparency_effects():
```
Ana06, line 151: I think the transparency effect does not work in a dark taskbar in non-activated Windows (at least I see no difference). What about in addition **setting the taskbar to rose color in the dark mode**? 

I am ok with keeping the transparency if you like it (I will never see it anyway :rofl:).

Ana06 (changes requested):
I have tested it locally and it is very fancy! :tada: Great work @emtuls!

packages/internet_detector.vm/tools/chocolateyinstall.ps1
```diff
@@ -0,0 +1,33 @@
...
+
+# Move images to common directory
+$imagesPath = Join-Path $packageToolDir "Images"
+Move-Item "$imagesPath\*" ${Env:VM_COMMON_DIR} -Force
+
+# Get path to pythonw for executing python without terminal popup
+$executablePath = (Get-Command pythonw).Source
+$filePath = Join-Path $packageToolDir "$toolName.pyw"
+
+VM-Install-Shortcut $toolName $category $executablePath -arguments $filePath
```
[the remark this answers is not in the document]
Ana06: after doing some more testing, it seems to work without admin rights. :confused:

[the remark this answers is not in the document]
Ana06: We should keep it consistent with how we run the scheduled task, is it running with admin rights?

packages/internet_detector.vm/tools/internet_detector.pyw
```diff
@@ -0,0 +1,421 @@
...
+    return False
+
+
+def check_internet_and_update_tray_icon():
+    global tray_icon, hicon_indicator_off, hicon_indicator_on, default_color
+    if check_internet():
+        tray_icon.set_icon(hicon_indicator_on)
+        tray_icon.set_tooltip("Internet Connection: Detected")
+        set_transparency_effects(1)
+        set_taskbar_accent_color(ROSE_COLOR, ROSE_ACCENT_PALETTE, ACCENT_PREVALENCE)  # Rose Color = #ff5200c3
```
Ana06, line 295: I would move this `# Rose Color = #ff5200c3` comment to where the `ROSE_ACCENT_PALETTE` constant is defined

Ana06 (commented):
Great work @emtuls :tada: 

image

Just some minor improvement suggestions :wink: 
- can we change the pink icon to have more contrast with the pink taskbask? (for example with a more intense/lighter pink) it is a bit difficult to see when it is hidden and it does not hurt when it is not:
image
image

- it would be nice that the icon is always visible. It seems I can move it manually to the taskbar and then it stays there. It would be nice if the tool could place it there. I propose you open an issue your this and research if it is possible after this PR has been merged. As it can be placed manually in the taskbar, I do not consider this a blocker and I would like to start using the tool! :tada:

emtuls:
> Great work @emtuls 🎉
> 
> image
> 
> Just some minor improvement suggestions 😉
> 
> * can we change the pink icon to have more contrast with the pink taskbask? (for example with a more intense/lighter pink) it is a bit difficult to see when it is hidden and it does not hurt when it is not:
>   image
>   image
> * it would be nice that the icon is always visible. It seems I can move it manually to the taskbar and then it stays there. It would be nice if the tool could place it there. I propose you open an issue your this and research if it is possible after this PR has been merged. As it can be placed manually in the taskbar, I do not consider this a blocker and I would like to start using the tool! 🎉

I have updated it to just use the colored version, which I think looks great with the pink taskbar.

Unfortunately, I don't think there's a way to force the icon to stay on the tray in the taskbar...For me, it starts there, but sometimes gets moved, but it can be moved back down manually. Couldn't find any way to force it to stay on the bottom.

packages/internet_detector.vm/tools/default.ini
```diff
@@ -0,0 +1,352 @@
...
+ProcessWhiteList: ncat.exe, nc.exe
+HostBlackList: 5.5.5.5
+Hidden:      False
+
+[DNS Server]
+Enabled:     True
+Port:        53
+Protocol:    UDP
+Listener:    DNSListener
+ProcessBlackList: pythonw.exe
```
Ana06, line 264: we can not exclude `pythonw.exe`. It is too generic and it could be needed to run malware.

Two options:
### duplicate `pythonw.exe` with a different name [ugly temporary workaround]
In the package, duplicate `pythonw.exe` in the same directory with name `pythonw_internet_detector.exe`. Use `pythonw_internet_detector.exe` to run the tool and in the fakenet log. It is a bit ugly, but we could use it as a temporary workaround to be able to merge this PR and start testing this tool.

### build the tool
We could build the tool into an executable in the package. I see [capa doing just: `python -m build`](https://github.com/mandiant/capa/blob/master/.github/workflows/publish.yml), so it may be also not difficult to do. @mr-tz and @williballenthin may be able to help with how to best implement it. :pray:

packages/internet_detector.vm/tools/chocolateyinstall.ps1
```diff
@@ -0,0 +1,37 @@
...
+
+# Create scheduled task for tool to run every 2 minutes.
+$action = New-ScheduledTaskAction -Execute $executablePath -Argument "$filePath"
+$trigger = New-ScheduledTaskTrigger -Once -At (Get-Date) -RepetitionInterval (New-TimeSpan -Minutes 2)
+Register-ScheduledTask -Action $action -Trigger $trigger -TaskName 'Internet Detector' -Force
+
+# Move updated `default.ini` with tool added to ProcessBlackList to FakeNet Config Directory
```
Ana06, line 32:
```suggestion
# Move updated `default.ini` with tool added to ProcessBlackList to FakeNet Config Directory
# TODO: Contribute this change to the fakenet default configuration
```

packages/internet_detector.vm/tools/chocolateyinstall.ps1
```diff
@@ -0,0 +1,37 @@
...
+$fakenetConfigDir = Join-Path ${Env:UserProfile} "Desktop\fakenet_logs.lnk"
+if (Test-Path $fakenetConfigDir) { # Check if the shortcut exists
+    $targetPath = (New-Object -ComObject WScript.Shell).CreateShortcut($fakenetConfigDir).TargetPath
+    Move-Item "$packageToolDir\default.ini" -Destination $targetPath
```
Ana06, line 36: I think you are doing something wrong moving the file, as the file I see in my fakenet default configuration after installing this package does not have your changes. Please test this out @emtuls!

Also, I would copy instead of moving, so that we have a copy of the file in case it is needed it.

Ana06 (commented):
tool looks good! we need to figure out

packages/internet_detector.vm/tools/chocolateyinstall.ps1
```diff
@@ -0,0 +1,37 @@
+$ErrorActionPreference = 'Stop'
+Import-Module vm.common -Force -DisableNameChecking
+
+$toolName = 'internet_detector'
+$category = 'Utilities'
```
Ana06, line 5: Wondering if `Networking` is a better category than `Utilities`. @mandiant/flare-vm opinions?

Ana06 (changes requested):
If I have internet in the VM and I start fakenet, I expect the tool to keep showing me there is internet. But this is not the case at the moment (both with and without the config file). Is this the fakenet bug reported in https://github.com/mandiant/flare-fakenet-ng/issues/190?

packages/internet_detector.vm/tools/chocolateyinstall.ps1
```diff
@@ -0,0 +1,37 @@
...
+# # Create scheduled task for tool to run every 2 minutes.
+# $action = New-ScheduledTaskAction -Execute $rawToolPath
+# $trigger = New-ScheduledTaskTrigger -Once -At (Get-Date) -RepetitionInterval (New-TimeSpan -Minutes 2)
+# Register-ScheduledTask -Action $action -Trigger $trigger -TaskName 'Internet Detector' -Force
+
+# Copy updated `default.ini` with tool added to ProcessBlackList to FakeNet Config Directory
+$fakenetConfigDir = Join-Path ${Env:UserProfile} "Desktop\fakenet_logs.lnk"
+if (Test-Path $fakenetConfigDir) { # Check if the shortcut exists
+    $targetPath = (New-Object -ComObject WScript.Shell).CreateShortcut($fakenetConfigDir).TargetPath
+    Copy-Item "$packageToolDir\default.ini" -Destination $targetPath
```
Ana06, line 36: **[IMPORTANT]**As mentioned in a previous comment, this is not working. You are copying the config file to `C:\Tools\fakenet\fakenet3.2-alpha` instead of `C:\Tools\fakenet\fakenet3.2-alpha\configs`, so the config is not being applied.

I also think we should use the tool directory directly instead of the shortcut, as if the user removes the shortcut in the desktop everything should continue working.

```suggestion
$fakenetConfigDir = Get-ChildItem "C:\Tools\fakenet\*\configs"
if ($fakenetConfigDir) {
    Copy-Item "$packageToolDir\default.ini" -Destination $fakenetConfigDir
```

Ana06 (changes requested):

packages/internet_detector.vm/tools/chocolateyinstall.ps1
```diff
@@ -0,0 +1,37 @@
...
+
+VM-Install-Shortcut -toolName $toolName -category $category -executablePath "$toolDir/$toolName.exe"
+
+# TODO - Uncomment when FakeNet BlackList for DNS is fixed/addressed. https://github.com/mandiant/flare-fakenet-ng/issues/190
+# # Create scheduled task for tool to run every 2 minutes.
+# $action = New-ScheduledTaskAction -Execute $rawToolPath
+# $trigger = New-ScheduledTaskTrigger -Once -At (Get-Date) -RepetitionInterval (New-TimeSpan -Minutes 2)
+# Register-ScheduledTask -Action $action -Trigger $trigger -TaskName 'Internet Detector' -Force
+
+# Copy updated `default.ini` with tool added to ProcessBlackList to FakeNet Config Directory
```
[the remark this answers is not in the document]
Ana06: Alternatively (better more robust solution): Move this to our fakenet package in this PR.

We can still open the PR/issue in fakenet afterwards to discuss if porting the change upstream is an option, as otherwise we will have to keep this file in sync in our package which is not ideal.

packages/fakenet-ng.vm/tools/chocolateyinstall.ps1
```diff
@@ -34,6 +35,13 @@ try {
   VM-Install-Shortcut -toolName $toolName -category $category -executablePath $executablePath -executableDir $toolDir -consoleApp $true
   Install-BinFile -Name $toolName -Path $executablePath
 
+  # Replace `default.ini` with our modified one that includes change for 'internet_detector'.
+  # TODO: Keep our modified `default.ini` in-sync on updates to package.
```
Ana06, line 39: [nitpick]
```suggestion
  # IMPORTANT: Keep our modified `default.ini` in-sync on updates to package.
```

There is nothing TODO if no update is done :wink:

Ana06 (approved):
Thanks for all the work @emtuls, two minor improvement suggestions, but I think we are good to merge this! :wink:

Ana06 (approved):
We are ready to finally get this in and get my pink dynamic analysis VM! :pink_heart:

PR #1122, opened 2024-08-26
Fix daily test

Ana06:
The daily test hasn't run correctly in the last weeks. The testing of the packages seem to be aborted because the default maximum execution time (6 hours) is exceeded. Increase the timeout to 12 hours.

The failure caused the README badge to be updated wrongly: 
image
Change the conclusion status of the package testing step to success even if packages fail, as the action is still
running as expected (we want to count the failures, so we expect failures). This way we detect other errors early avoiding the wiki and the badge to be updated.

### DIFF
--- a/.github/workflows/daily.yml
+++ b/.github/workflows/daily.yml
@@ -10,6 +10,8 @@ on:
 jobs:
   test:
     runs-on: ${{ matrix.os }}
+    # Testing all packages takes several hours
+    timeout-minutes: 720
     strategy:
       fail-fast: false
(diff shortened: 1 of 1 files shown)


mr-tz (approved):

Ana06:
Thanks for the review @mr-tz! I have just triggered the daily test manually :wink:

emtuls:
Seems we are getting an ERROR still after the latest run:
image


It looks like there's an issue with `git.install` dependency, where git isn't installing correctly, which then leads to a bunch of other installation failures: https://github.com/mandiant/VM-Packages/actions/runs/10559091787/job/29249826630#step:3:4884
image

PR #1137, opened 2024-09-13
Fix Custom Taskbar from breaking on Win 11

emtuls:
This is step 2 for fixing the Windows 11 installation error as noted in https://github.com/mandiant/flare-vm/issues/603

This works in tandem with https://github.com/mandiant/flare-vm/pull/617 and needs to be merged alongside it in order for installations not to break.

### DIFF
--- a/packages/installer.vm/tools/chocolateyinstall.ps1
+++ b/packages/installer.vm/tools/chocolateyinstall.ps1
@@ -32,15 +32,6 @@ try {
     }
     VM-Write-Log "INFO" "Packages installation complete"
 
-    ## Configure taskbar with custom Start Layout if it exists.
-    $customLayout = Join-Path ${Env:VM_COMMON_DIR} "CustomStartLayout.xml"
-    if (Test-Path $customLayout) {
-        Import-StartLayout -LayoutPath $customLayout -MountPath "C:\"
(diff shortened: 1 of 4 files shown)


packages/installer.vm/tools/chocolateyinstall.ps1
```diff
@@ -32,14 +32,7 @@ try {
...
 
-    ## Configure taskbar with custom Start Layout if it exists.
-    $customLayout = Join-Path ${Env:VM_COMMON_DIR} "CustomStartLayout.xml"
-    if (Test-Path $customLayout) {
-        Import-StartLayout -LayoutPath $customLayout -MountPath "C:\"
-        Stop-Process -Name explorer -Force  # This restarts the explorer process so that the new taskbar is displayed.
-    } else {
-        VM-Write-Log "WARN" "CustomStartLayout.xml missing. No items will be pinned to the taskbar."
-    }
+    Stop-Process -Name explorer -Force  # This restarts the explorer process so that the new taskbar is displayed.
```
Ana06, line 35: [nitpick] As we are not setting the taskbar here anymore, this code looks out of place now. What about adding this code to the `VM-Refresh-Desktop` function or moving this line after the call to `VM-Refresh-Desktop`?

Ana06 (approved):
LGTM!

PR #1146, opened 2024-10-02
Support the valid return codes at googlechrome.vm

### LINKED ISSUE #1138: googlechrome.vm fails when Github action image updates to latest chrome
### What's the problem?
As noted in [this](https://github.com/mandiant/VM-Packages/issues/1111#issuecomment-2353439350) comment, when the image that is used by Github Actions updates to use the latest Google Chrome, an error occurs with the `googlechrome.vm` installation because the same version is already installed, thus causing the package to fail to install and any subsequent package that depends on it.

Second, when this package fails to install, it also causes an error for other packages that depend on `googlechrome.vm`, one of those being `cyberchef.vm`, and this raises the error seen in https://github.com/mandiant/VM-Packages/issues/1111. 
When `googlechrome.vm` is not installed, the path to the browser is not in the location we expect, which also causes a failure for finding the icon in the path we are using.
This particular part of the issue will be resolved when we change the icon for CyberChef to be something other than Google Chrome in this PR: https://github.com/mandiant/VM-Packages/pull/1054

See error below for what I mean by `googlechrome.vm` failing to install in the latest run. The image was recently updated to use the latest chrome browser (https://github.com/actions/runner-images/releases/tag/win22%2F20240912.1) and this issue has been noted in other areas as an issue that can occur: https://github.com/yarnpkg/yarn/issues/1647#issuecomment-339732324
image

### Steps to Reproduce

View our daily logs when Github Action images are updated to latest chrome as noted here: https://github.com/actions/runner-images/releases.

### Environment

Environment Agnostic

Depends on Github Actions Images and when they update to latest chrome.

### Additional Information

_No response_

### Detected by test suite

No

d35ha:
Closes #1138.

### DIFF
--- a/packages/googlechrome.vm/googlechrome.vm.nuspec
+++ b/packages/googlechrome.vm/googlechrome.vm.nuspec
@@ -2,7 +2,7 @@
 <package xmlns="http://schemas.microsoft.com/packaging/2015/06/nuspec.xsd">
   <metadata>
     <id>googlechrome.vm</id>
-    <version>0.0.0.20240425</version>
+    <version>0.0.0.20241002</version>
     <authors>Google LLC.</authors>
     <description>Chrome is a popular web browser.</description>
     <dependencies>
--- a/packages/googlechrome.vm/tools/chocolateyinstall.ps1
+++ b/packages/googlechrome.vm/tools/chocolateyinstall.ps1
@@ -1,6 +1,10 @@
 $ErrorActionPreference = 'Stop'
 Import-Module vm.common -Force -DisableNameChecking
 
+$fileType = "MSI"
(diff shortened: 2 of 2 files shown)


Ana06 (commented):
Is it possible to use the `VM-Install-With-Installer` helper? It would make the code simpler, consistent ans easier to maintain. :thinking:

d35ha:
> Is it possible to use the `VM-Install-With-Installer` helper? It would make the code simpler, consistent ans easier to maintain. 🤔

@Ana06 
Using ` VM-Install-With-Installer` would require a SHA256 hash, therefore a versioned URL. The used MSI URL is not versioned.

Ana06:
@d35ha 

> Using VM-Install-With-Installer would require a SHA256 hash, therefore a versioned URL. The used MSI URL is not versioned.

You are right, how could I forgot this? :see_no_evil:

Ana06 (approved):
Thanks for the fix @d35ha!

Issue #1100, opened 2024-07-01
duplication when opening application from taskbar


Ana06:
### Details

when starting fakenet, chrome, and processExplorer and IDA from the taskbar, I get a duplication which is confusing: 
![image](https://github.com/mandiant/VM-Packages/assets/16052290/17dadc4a-f0a8-4f84-8475-a46d9d2994ad)

@emtuls I think this is related to some things you have worked on. Do you have any ideas how to fix it?

emtuls:
@Ana06 
I am not sure if this can be completely fixed unfortunately.

**Unable to be fixed:**
- For the `IDA` shortcut, because our shortcut is to a launcher, it is technically not the same program that is being run (it launches the correct IDA), so that is why it's creating a whole new icon.
- For `Cyberchef`, we are using a shortcut that launches `chrome`, and because the shortcut itself is not actually `chrome`, it will create a new icon as well, as it is not the actual program that is being run.
  - **Alternatively:** I can make the shortcut be just to open up a `.html` page, which gets us closer (once the default application for `.html` is properly done), but this would not allow us to pin a shortcut to the taskbar because Microsoft does not allow pinning of a `.html` pages to a taskbar, but it does work for a desktop icon and in the `Tools` directory.

**Fixable:**
- For `Procmon` and `Procexp`, I can fix them, since it seems that our shortcuts are to the 32 bit versions and they must detect that it is running on a 64 bit system and then spawns the 64 bit version, which is a separate application and thus, creates a new icon. Changing these to just run the 64 bit programs directly mitigates this.

**Unsure**
-  For `Fakenet`, I did not experience this issue on my system, so I would need to investigate further. Would you be able to provide me  the Target your `fakenet` is using for the shortcut?

Ana06:
@emtuls can you please send a PR to fix procmon and procexp now (that seem easy to fix) and we can continue discussing the rest here?

the chrome issue is related to https://github.com/mandiant/VM-Packages/pull/1054, as this would again make it look a bit different. 

Have you tested the latest version of fakenet?

Ana06:
procmon and procexp should be fixed now.

Ana06:
# Proposal to fix Chrome/CyberChef

## Option 1: Use the CyberChef icon in the taskbar for Chrome

Use the CyberChef icon in the taskbar for Chrome that by default opens CyberChef:

![image](https://github.com/user-attachments/assets/945682ff-4438-4999-b9b4-0e06ec617452)

[...]

PR #1148, opened 2024-10-02
[cyberchef] Improve CyberChef shortcut

Ana06:
Improve the CyberChef shortcut to avoid duplication of CyberChef and Chrome in the taskbar. 
image

After installing the package, restart the system to ensure it is working.

It partially addresses https://github.com/mandiant/VM-Packages/issues/1100. See https://github.com/mandiant/VM-Packages/issues/1100#issuecomment-2388335265 for other considered options and more details.

### DIFF
--- a/packages/cyberchef.vm/cyberchef.vm.nuspec
+++ b/packages/cyberchef.vm/cyberchef.vm.nuspec
@@ -2,7 +2,7 @@
 <package xmlns="http://schemas.microsoft.com/packaging/2015/06/nuspec.xsd">
   <metadata>
     <id>cyberchef.vm</id>
-    <version>10.19.0.20240913</version>
+    <version>10.19.0.20241002</version>
     <authors>GCHQ</authors>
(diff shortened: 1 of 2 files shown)


emtuls:
Doesn't seem to work consistently for me. 🙁

First time I installed it, it didn't work at first, then after a few tries (of doing nothing different), it started to work. But then I reverted my snapshot and tried again, and couldn't get it to work at all, even with a restart of the system and also even just with restarting explorer.exe.

It sort of seems to work when I manually make a chrome shortcut, set the icon to cyberchef, then pin it and restart explorer.exe
Which is weird, because that's basically what you're doing in the package...

Ana06:
Thanks for testing it @emtuls, @mandiant/vms can someone else test it?

Even if does not improve the CyberChef + Chrome icons in the taskbar, do we want to merge the modification as it simplifies the code and uses the `VM-Install-Shortcut` helper?

Issue #1144, opened 2024-10-01
Improve signature verification in VM-Assert-Signature


Ana06:
### Details

The current implementation of `VM-Assert-Signature` uses `Get-AuthenticodeSignature` status. Reading [Microsoft documentation](https://learn.microsoft.com/en-us/dotnet/api/system.management.automation.signaturestatus?view=powershellsdk-7.4.0), I understand that this only checks that the file has a syntactically syntactically valid signature, I think we should instead verify the signing authority, for example using `signtool.exe`:
```
"C:\Program Files (x86)\Windows Kits\10\bin\10.0.17763.0\x86\signtool.exe" verify /pa /all /tw /v /d googlechromestandaloneenterprise64.msi
```

This works well for Google Chrome, Sysinternals and Metasplot, but not for RegCool. I suggest using hashes again in RegCool. Should we also replace it in the configuration by total-registry to avoid that updates break the tool leaving FLARE-VM without a registry tool?

Ana06:
It may also be possible to validate the signature of RegCool using signtool, but I am not sure how. Can someone help here? @mandiant/vms

Ana06:
I am also not sure which package we need to install to ensure `signtool.exe` is installed. It is not installed in my VM before installing FLARE-VM, but it is there after. :female_detective:

Ana06:
It seems `vcbuildtools.vm` installs it. We could require `vcbuildtools.vm` as dependency in the packages using this function and access the tool with:
```PowerShell
 $signtoolPath = Get-ChildItem -Path "C:\Program Files*\Windows Kits\10\bin\*\x86\signtool.exe" | Select -Last 1
& $signtoolPath
```

PR #1157, opened 2024-11-06
Improve VM-Assert-Signature & restore hash installation for RegCool

Ana06:
The current implementation of `VM-Assert-Signature` uses `Get-AuthenticodeSignature` status, that only checks that the file has a syntactically syntactically valid signature. Verify the signing authority using `signtool.exe`.

This new signature verification using `signtool.exe` does not work for RegCool. Restore the previous installation checking the SHA256. As the tool is using a URL that does not include the version, the hash will change with every update, breaking the package. If the package is updated often, we will need to remove it from the default configuration (replacing it total-registry) or researching if it is possible to use `signtool.exe` to verify the package. I suggest we discuss this outside this PR.

Note that the current `regcool.vm` package implementation being replaced in this PR had an important bug: `VM-Assert-Signature` was not called at all, as `Get-ChildItem -Path "$toolDir\*.{exe,dll}"` does not match any file.

Closes https://github.com/mandiant/VM-Packages/issues/1144

### DIFF
--- a/packages/common.vm/common.vm.nuspec
+++ b/packages/common.vm/common.vm.nuspec
@@ -2,7 +2,7 @@
 <package xmlns="http://schemas.microsoft.com/packaging/2015/06/nuspec.xsd">
   <metadata>
     <id>common.vm</id>
-    <version>0.0.0.20241029</version>
+    <version>0.0.0.20241106</version>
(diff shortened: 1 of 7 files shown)


emtuls (approved):
Awesome. Looks good! Thank you @Ana06 !

Issue #1152, opened 2024-10-09
IDA 9: open with IDA menu option 


Ana06:
### Details

The `open with IDA` menu option appears to not work with IDA 9.0 installed. @emtuls could you take a look?

PR #1161, opened 2024-11-07
Fix IDA Pro 9.0+ Right Click Open

emtuls:
This fixes https://github.com/mandiant/VM-Packages/issues/1152

The lastest IDA Pro 9.0 uses a `IDA Professional` rather than `IDA Pro` for its display name in the registry.


Ana06 (approved):
LGTM, we still need to update idafree to use the newer version, but you are going to run into https://github.com/mandiant/VM-Packages/issues/1166

Ana06:
@emtuls IDA free is fixed: https://github.com/mandiant/VM-Packages/issues/1166#issuecomment-2490761124. Please send a PR to update the IDA package after merging this PR.

Issue #1142, opened 2024-09-30
Bot hasn't updated sclauncher


Ana06:
### Details

As @jstrosch mentioned in https://github.com/mandiant/VM-Packages/issues/181#issuecomment-2241176682, sclauncher has a [new release](https://github.com/jstrosch/sclauncher/releases) and our bot does not seem to update it. Investigate why and improve update logs.

PR #1162, opened 2024-11-08
fix bug on update_package.py

sara-rn:
fix issue #1142 for  when --update-type is GITHUB_URL

### DIFF
--- a/scripts/utils/update_package.py
+++ b/scripts/utils/update_package.py
@@ -37,7 +37,10 @@ def get_latest_version(org, project, version):
         print(f"GitHub API response not ok: {response.status_code}")
         return None
     latest_version = response.json()["tag_name"]
-    return latest_version
+    if latest_version.startswith('v'):
+    	return latest_version.replace('v','',1)
+    else:
+    	return latest_version
 
 
 # Get url response's content hash (SHA256)


scripts/utils/update_package.py
```diff
@@ -37,7 +37,10 @@ def get_latest_version(org, project, version):
         print(f"GitHub API response not ok: {response.status_code}")
         return None
     latest_version = response.json()["tag_name"]
-    return latest_version
+    if latest_version.startswith('v'):
```
Ana06, line 40: @naacbin introduced the support for packages whose version is preceded by `v` in https://github.com/mandiant/VM-Packages/pull/1067, but it seems the `v` was only taken into account when matching the url in the file and not to get the latest version, so it seems like the fix has never worked.

What is really getting me confused is that I see upx being correctly updated by our bot in https://github.com/mandiant/VM-Packages/commit/acec181a0574c4c85f0de99d48f875093b3e23a7 before https://github.com/mandiant/VM-Packages/pull/1067, how is this possible? :female_detective:

scripts/utils/update_package.py
```diff
@@ -37,7 +37,10 @@ def get_latest_version(org, project, version):
...
+    	return latest_version.replace('v','',1)
```
Ana06, line 41: A bit more elegant :wink: 
```suggestion
    	return latest_version[1:]
```

scripts/utils/update_package.py
```diff
@@ -104,6 +107,7 @@ def update_github_url(package):
 
     latest_version = None
     for url, org, project, version in matches:
+        # version excludes `v` from the capturing group in the regex therefore latest_version_match mustn't include `v` if the version starts with `v`
```
Ana06, line 110: Can you move this comment (maybe with a bit rewording needed) just before the line `if latest_version.startswith('v'):`? Here it is difficult to understand what it is refering to :confused:

sara-rn: done in the last commit

Ana06 (commented):
@sara-rn can you please squash the commits? :smile:

scripts/utils/update_package.py
Ana06: I still don't understand how it could work in acec181, but I think the change should fix the issue, so I think we can merge it. But I would still be interested in understanding how this broke, so please if someone (maybe @naacbin)  has any clue let me know!

Ana06 (commented):
@sara-rn you have kept all commit messages in the squashed commit. Can you please reword the commit message to explain why the change is needed? For example:
```
Fix version bug in update_package.py

The regex to match the current tool version in `update_github_url ` does not 
include the `v`. However the function `get_latest_version` does include the `v` 
producing a URL with a duplicated `v` when replacing the current by the latest
version. Remove the `v` in `get_latest_version` to fix the issue.
```

You can do this with `git commit --amend` (easier in this case) or with `git rebase -i HEAD~1` + `r` (which can be used to change the commit message of commits messages that are not the last one too).

I recommend reading https://cbea.ms/git-commit for more info on best practices to write good commit messages. :wink:

Ana06 (approved):
Thanks @sara-rn

PR #1167, opened 2024-11-19
Fixed broken URL for dcode.vm

micahflack:
dcode.vm package is broken and throws error 404 when downloading installer.

```
ERROR: The remote file either doesn't exist, is unauthorized, or is forbidden for url 'https://www.digital-detective.net/download/download.php?downcode=ae2znu5994j1lforlh03'. Exception calling "GetResponse" with "0" argument(s): "The remote server returned an error: (404) Not Found."
The install of dcode.vm was NOT successful.
Error while running 'C:\ProgramData\chocolatey\lib\dcode.vm\tools\chocolateyinstall.ps1'.
 See log for details.
```

Following the original link in browser returns...

`https://www.digital-detective.net/download/downloadbac.php?downcode=ae2znu5994j1lforlh03`

`...downloadbac.php?...` instead of `...download.php?...`

Does the command, Get-WebFIle, have problems following redirects? Hope this is helpful.

### DIFF
--- a/packages/dcode.vm/dcode.vm.nuspec
+++ b/packages/dcode.vm/dcode.vm.nuspec
@@ -2,7 +2,7 @@
 <package xmlns="http://schemas.microsoft.com/packaging/2015/06/nuspec.xsd">
   <metadata>
     <id>dcode.vm</id>
-    <version>5.6.24123.20240507</version>
+    <version>5.6.24123.20241121</version>
     <authors>Digital Detective Group</authors>
     <description>Utility for converting data found on desktop and mobile devices into human-readable timestamps.</description>
     <dependencies>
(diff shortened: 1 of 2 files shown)


Ana06 (commented):
@micahflack thanks for the PR! Both URLs work for me in the browser, but I see the error in the CI. In order to trigger the tests in this PR (so that we can see if the change fixes the issue) you need to update the package version in the nuspec as documented in [our wiki](https://github.com/mandiant/VM-Packages/wiki/Chocolatey-packages#version):
> The version needs to be increased every time you modify an existing package

micahflack:
@Ana06 thank you! I made the changes like you suggested, but it looks like my fix is not what is needed; both URLs produce the error. I tried downgrading chocolatey to 2.3.0 from the current 2.4.0 on my install of Flare - but that did not work either. I thought that maybe it could have been caused by the recent release.

micahflack:
@Ana06 sure - I'll share whatever logs from choco as well

PR #1177, opened 2024-11-25
Use updated ida_launcher

emtuls:
This finalizes the fix for https://github.com/mandiant/VM-Packages/issues/1152 by making `idafree.vm` use the updated launcher.


Ana06 (approved):
Thanks @emtuls! I tested locally and it is working nicely :wink:

Issue #1174, opened 2024-11-22
enable internet detector by default


Ana06:
### Details

Enable internet detector by default ensuring this is not noisy in fakenet. This should be supported by the new fakenet version 3.3, that we have just updated to.

PR #1178, opened 2024-11-25
Add internet_detector scheduled task

emtuls:
This is step 1 for fixing https://github.com/mandiant/VM-Packages/issues/1174, which adds `internet_detector.exe` to a scheduled task so that it is always running, now that `fakenet-ng` has been updated to version `3.3` which allows blacklisted processes to not be displayed in the terminal interface.

Step 2 is https://github.com/mandiant/flare-vm/pull/630, which is to add the modification of the `DNSCache` registry key to our `flarevm` install script so that DNS requests will display as the asking process rather than `svchost`.

### DIFF
--- a/packages/internet_detector.vm/tools/chocolateyinstall.ps1
+++ b/packages/internet_detector.vm/tools/chocolateyinstall.ps1
@@ -23,8 +23,7 @@ Copy-Item "$imagesPath\*" ${Env:VM_COMMON_DIR} -Force
 
 VM-Install-Shortcut -toolName $toolName -category $category -executablePath "$toolDir/$toolName.exe"
 
-# TODO - Uncomment when FakeNet BlackList for DNS is fixed/addressed. https://github.com/mandiant/flare-fakenet-ng/issues/190
-# # Create scheduled task for tool to run every 2 minutes.
-# $action = New-ScheduledTaskAction -Execute $rawToolPath
-# $trigger = New-ScheduledTaskTrigger -Once -At (Get-Date) -RepetitionInterval (New-TimeSpan -Minutes 2)
-# Register-ScheduledTask -Action $action -Trigger $trigger -TaskName 'Internet Detector' -Force
+# Create scheduled task for tool to run every 2 minutes.
+$action = New-ScheduledTaskAction -Execute "$toolDir/$toolName.exe"
+$trigger = New-ScheduledTaskTrigger -Once -At (Get-Date) -RepetitionInterval (New-TimeSpan -Minutes 2)
+Register-ScheduledTask -Action $action -Trigger $trigger -TaskName 'Internet Detector' -Force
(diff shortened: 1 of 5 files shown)


packages/internet_detector.vm/tools/chocolateyinstall.ps1
```diff
@@ -23,8 +23,7 @@ Copy-Item "$imagesPath\*" ${Env:VM_COMMON_DIR} -Force
 
 VM-Install-Shortcut -toolName $toolName -category $category -executablePath "$toolDir/$toolName.exe"
 
-# TODO - Uncomment when FakeNet BlackList for DNS is fixed/addressed. https://github.com/mandiant/flare-fakenet-ng/issues/190
-# # Create scheduled task for tool to run every 2 minutes.
-# $action = New-ScheduledTaskAction -Execute $rawToolPath
-# $trigger = New-ScheduledTaskTrigger -Once -At (Get-Date) -RepetitionInterval (New-TimeSpan -Minutes 2)
-# Register-ScheduledTask -Action $action -Trigger $trigger -TaskName 'Internet Detector' -Force
+# Create scheduled task for tool to run every 2 minutes.
+$action = New-ScheduledTaskAction -Execute "$toolDir/$toolName.exe"
```
Ana06, line 27: is the tool then started several times? :thinking:

emtuls: Hmmm, my testing hadn't shown that, but when I force the Scheduled Task to run manually, it did cause it to create multiple instances.

I think I will have to modify `internet_detector` itself so that it checks if there is one already running before attempting to execute. Doesn't seem like an option we can do with a scheduled task.

Ana06: I see some changes in the internet_detector, but I think I am missing the change for this. @emtuls have you forgotten to add it or am I missing something?

emtuls: There isn't anything that can be done on the scheduled task side that I can see. 🤔

Unless you're still questioning if it should be executed every 2 minutes at all? Which comes down to how we want this tool to be run.

The reason I left that is if the program were to somehow crash or get stopped (either by malware or by an analyst whether it be on accident or not) then this would allow it to start back up without needing any user interaction.

The program itself now checks for if there's already an instance of itself running and if so, it won't run another (of course, the initial execution to determine that will happen, but then it closes).

But if there's a different way you'd like for it to be run, like only once on login/reboot, and any crash/stoppage requiring manual user interaction to start it again, then let me know and I can make that change. 🙂

Ana06 (changes requested):
@emtuls I have tested this locally, installing FLARE-VM from scratch (with a modified configuration only with `hashmyfiles.vm` as d). This means I am testing also the change in https://github.com/mandiant/flare-vm/pull/630. But it seems it is not working as expected, the tool is still very noisy in fakenet, not being able to enable it by default:

image

Have you tested it? Can you please explain how to ensure I am not missing anything?

Ana06 (approved):
After speaking with @emtuls, I can confirm that a restart fix the issue and this works as expected. Not sure if we need to force the restart. I suggest to leave it as it is and improve if it shows being an issue.

Issue #1130, opened 2024-09-03
malware-jail.vm fails although the tool is installed


Ana06:
### Details

Both locally and in daily CI, I see the following error in the log:
```
2024/09/03 09:29:54 [installer.vm] chocolateyinstall.ps1 [+] ERROR : Failed to install: magika.vm
2024/09/03 09:29:54 [installer.vm] chocolateyinstall.ps1 [+] ERROR : Failed to install: malware-jail.vm
```
It seems to be caused by:
```
ERROR: ERROR: pip's dependency resolver does not currently take into account all the packages that are installed. This behaviour is the source of the following dependency conflicts.
```

The tools appear to be installed correctly. So this is very confusing and should investigate how to fix it.

emtuls:
**NOTE:** This is actually 2 separate issues. One for `magika` and separate issue affecting `malware-jail`.

This specific `pip` issue is mostly due to our installation of Python tools being installed separately and thus, Python's dependency resolver is unable to properly satisfy any dependency issues, which is what is displaying the error.

The primary underlying issue is that `magika` version `0.5.1` had an error in their code that was recently fixed in `0.6.0rc1` (current unreleased on pypi) which is what made `numpy` version `>=1.26` be required even for Python version `<3.12` when it should have only needed `>=1.24`: https://github.com/google/magika/commit/ebac72a3798de2608ef518edc1e21e71f2a2d0ab

And because we have other python tools installed via `libraries.python3.vm`, their dependencies are not checked against this and thus more specifically we run into an issue because `StringSifter` requires `numpy==1.24.4`: https://github.com/mandiant/stringsifter/blob/33c0cd5538bf4dc499505b63f89bc045b191a0df/requirements.txt#L110
So when `numpy` is updated to `1.26` because of `magika`, we get an error:
```
ERROR: pip's dependency resolver does not currently take into account all the packages that are installed. This behaviour is the source of the following dependency conflicts.
stringsifter 3.20230711 requires numpy==1.24.4, but you have numpy 1.26.4 which is incompatible.
```
![image](https://github.com/user-attachments/assets/7f88a5e8-1d76-427a-9020-0f1e077332fc)

This can be resolved immediately by just using `magika` version `0.5.0`, and should hopefully be resolved when version `0.6.0` is officially released.


The second issue is with `git.install` not installing properly for some reason, which is causing `malware-jail.vm` to fail:
[...]

PR #1179, opened 2024-11-25
Pin magika version to fix pip issue

emtuls:
This pins `magika` to version `0.5.0` to fix a pip resolving issue with `numpy` conflicting with `stringsifter` as noted in https://github.com/mandiant/VM-Packages/issues/1130. It should be fixed when `magika` version `0.6.0` is officially released and we can revert back to how we installed it before after verifying.

### DIFF
--- a/packages/common.vm/tools/vm.common/vm.common.psm1
+++ b/packages/common.vm/tools/vm.common/vm.common.psm1
@@ -1762,10 +1762,12 @@ function VM-Install-With-Pip {
         [Parameter(Mandatory=$true)]
         [string] $category,
         [Parameter(Mandatory=$false)]
+        [string] $version = "", # Version using pip format, example: "==0.5.0"
+        [Parameter(Mandatory=$false)]
         [string] $arguments = "--help"
     )
     try {
-        VM-Pip-Install $toolName
+        VM-Pip-Install $toolName$version
         $executablePath = "$(where.exe $toolName)"
 
         VM-Install-Shortcut $toolName $category $executablePath -consoleApp $true -arguments $arguments
--- a/packages/magika.vm/magika.vm.nuspec
+++ b/packages/magika.vm/magika.vm.nuspec
@@ -2,11 +2,11 @@
 <package xmlns="http://schemas.microsoft.com/packaging/2015/06/nuspec.xsd">
   <metadata>
     <id>magika.vm</id>
-    <version>0.0.0.20240607</version>
+    <version>0.5.0</version>
     <authors>Yanick Fratantonio, Luca Invernizzi, Marina Zhang, Giancarlo Metitieri, Thomas Kurt, Francois Galilee, Alexandre Petit-Bianco, Loua Farah, Ange Albertini, Elie Bursztein</authors>
     <description>Magika is an AI powered file type detection tool that uses deep learning to provide accurate detection.</description>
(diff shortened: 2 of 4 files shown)


packages/magika.vm/magika.vm.nuspec
```diff
@@ -2,7 +2,7 @@
 <package xmlns="http://schemas.microsoft.com/packaging/2015/06/nuspec.xsd">
   <metadata>
     <id>magika.vm</id>
-    <version>0.0.0.20240607</version>
+    <version>0.0.0.20241124</version>
```
Ana06, line 5: Installing a concrete version is a good idea and now we can use that version for the package version too:
```suggestion
    <version>0.5.0</version>
```

Ana06 (changes requested):

packages/magika.vm/tools/chocolateyinstall.ps1
```diff
@@ -4,4 +4,6 @@ Import-Module vm.common -Force -DisableNameChecking
 $toolName = 'magika'
 $category = 'File Information'
 
-VM-Install-With-Pip -toolName $toolName -category $category
+VM-Pip-Install 'magika==0.5.0'
+$executablePath = "$(where.exe $toolName)"
+VM-Install-Shortcut $toolName $category $executablePath -consoleApp $true -arguments $arguments
```
Ana06, line 9: I would like that we continue to use `VM-Install-With-Pip` by extending it to add a version argument. As already mentioned in other recent issues (for example https://github.com/mandiant/VM-Packages/issues/1080#issuecomment-2523816211), we should start fixing the version of the tools installed with Pip. So this is a common need and not something particular of this package. Extending the helper simplifies the code, remove code duplication and ensure consistency around different packages.

@emtuls  could you please extend the helper so that this package looks like:
```suggestion
$version = "==0.5.0"

VM-Install-With-Pip -toolName $toolName -category $category -version $version
```

Ana06 (changes requested):

packages/common.vm/tools/vm.common/vm.common.psm1
```diff
@@ -1762,10 +1762,18 @@ function VM-Install-With-Pip {
...
+        [string] $version = "",
+        [Parameter(Mandatory=$false)]
         [string] $arguments = "--help"
     )
     try {
-        VM-Pip-Install $toolName
+        # Conditionally add version string
+        $toolNameForPip = $toolName
+        if (![string]::IsNullOrEmpty($version)) {  # Check if $version is not empty or null
+            $toolNameForPip += "==$version"
```
Ana06, line 1773: I think it would be more flexible to expect the version to include the `==`:
```suggestion
            $toolNameForPip += "$version"
```

This allows us more flexibility in case we need for example to specify a range of versions. I would then also add a comment next to the `version` argument to document this:
```PS
[string] $version = "",  # Version using pip format, example: "==0.5.0"

Ana06 (changes requested):

Ana06 (changes requested):
Thanks for adding a `version` argument to the `VM-Install-With-Pip` function @emtuls! :+1:

Ana06 (approved):
Thanks for all the work @emtuls!!! :smile:

Issue #1183, opened 2024-12-03
URL package proposal: vb-decompiler-lite.vm


sara-rn:
### Package Name

vb-decompiler-lite

### Tool Name

vb_decompiler_lite

### Package type

ZIP_EXE

### Is the tool a console application?

false

### Version

12.4

### Category

Visual Basic

### Tool's authors

DotFix Software

### Tool's description

Visual Basic decompiler

### Download URL

https://www.vb-decompiler.org/files/vb_decompiler_lite.zip

### Download SHA256 Hash

099760dcae9daa4c83885a3817cf6f17442ce709de0105993bfbb4f17db87e62

### Why is this tool a good addition?

to reverse engineer Visual Basic

PR #1184, opened 2024-12-04
:robot: Add vb-decompiler-lite.vm

vm-packages:
Automated PR from https://github.com/mandiant/VM-Packages/issues/1183 :rocket:


sara-rn (commented):
tested locally, everything looks good

sara-rn (approved):

packages/vb-decompiler-lite.vm/tools/chocolateyinstall.ps1
```diff
@@ -0,0 +1,17 @@
+$ErrorActionPreference = 'Stop'
+Import-Module vm.common -Force -DisableNameChecking
+
+try{
+
+    $toolName = 'VB Decompiler Lite'
+    $category = 'Visual Basic'
+
+    $zipUrl = 'https://www.vb-decompiler.org/files/vb_decompiler_lite.zip'
```
Ana06, line 9: This URLs does not include the version and will break with an update as the hash will change. Any chance we can get a URL that includes the version?

packages/vb-decompiler-lite.vm/tools/chocolateyinstall.ps1
```diff
@@ -0,0 +1,17 @@
+$ErrorActionPreference = 'Stop'
+Import-Module vm.common -Force -DisableNameChecking
+
+try{
+
+    $toolName = 'VB Decompiler Lite'
```
Ana06, line 6: I think we should use the executable name as tool name as documented in https://github.com/mandiant/VM-Packages/wiki/Coding-Conventions#toolname
```suggestion
    $toolName = 'VB Decompiler'
```

packages/vb-decompiler-lite.vm/tools/chocolateyinstall.ps1
```diff
@@ -0,0 +1,17 @@
...
+    $category = 'Visual Basic'
+
+    $zipUrl = 'https://www.vb-decompiler.org/files/vb_decompiler_lite.zip'
+    $zipSha256 = '099760dcae9daa4c83885a3817cf6f17442ce709de0105993bfbb4f17db87e62'
+
+    $toolDir = Join-Path ${Env:RAW_TOOLS_DIR} $toolName
+    $executablePath = (Join-Path $toolDir "VB Decompiler.exe")
```
Ana06, line 13: [nitpick] Ensure `$toolName` is on sync with the executable name:
```suggestion
    $executablePath = (Join-Path $toolDir "$toolName.exe")
```

packages/vb-decompiler-lite.vm/tools/chocolateyuninstall.ps1
```diff
@@ -0,0 +1,9 @@
+$ErrorActionPreference = 'Continue'
+Import-Module vm.common -Force -DisableNameChecking
+
+$toolName = 'VB Decompiler Lite'
```
Ana06, line 4:
```suggestion
$toolName = 'VB Decompiler'
```

packages/vb-decompiler-lite.vm/tools/chocolateyuninstall.ps1
```diff
@@ -0,0 +1,9 @@
...
+$category = 'Visual Basic'
+
+VM-Uninstall-With-Uninstaller $toolName $category "EXE" "/VERYSILENT /SUPPRESSMSGBOXES /NORESTART /SP-"
+
+VM-Uninstall $toolName $category
```
Ana06, line 9: Why is this function call needed? Is there something to remove after executing `VM-Uninstall-With-Uninstaller`? I have noticed that some similar packages call both, but I suspect this is wrong.

Have you tested the uninstallation as well?

packages/vb-decompiler-lite.vm/tools/chocolateyinstall.ps1
sara-rn: unfortunately not, the version is only included in the body of the html but not in the URL

sara-rn: only included in older versions, for example: https://www.vb-decompiler.org/files/vb_decompiler_lite_84.zip

packages/vb-decompiler-lite.vm/tools/chocolateyuninstall.ps1
sara-rn: I haven't tested that, I saw it in other packages

Ana06: You should also test the uninstallation locally (as we do not test it in the CI). After testing the installation, run `choco uninstall <package_name>` and ensure the uninstallation runs without error and that the tool and the shortcut are deleted. :wink:

packages/vb-decompiler-lite.vm/tools/chocolateyinstall.ps1
Ana06: how old are the older versions? If they are not too old, I prefer this option instead of having a URL that will eventually break. We try not to add packages to the default config for which we know the hash will break, as this confuses FLARE-VM users.

Ana06: @sara-rn :eyes:

Ana06: It seems current version is 12.4 and the one in the link is 8.4, so it is quite old. According to https://www.vb-decompiler.org/news.htm, it seems VB decompiler releases a new version 1-3 times per year. I think it is ok to use a hash that may break for this release cadence. I propose merging this Pr as it is and considering improving this package at a later point.

packages/vb-decompiler-lite.vm/tools/chocolateyuninstall.ps1
Ana06: I have run the uninstallation locally:
image

I get a message box although we are specifying `/SUPPRESSMSGBOXES`. I don't understand why this happens. As uninstallation is best-effort, I do not want to block this PR because of that. @sara-rn could you open an issue to document this issue?

Ana06 (commented):
@sara-rn please change the name in the uninstallation and squash the commits so that we can get the PR merged :wink:

packages/vb-decompiler-lite.vm/tools/chocolateyinstall.ps1
```diff
@@ -0,0 +1,17 @@
...
+
+    $zipUrl = 'https://www.vb-decompiler.org/files/vb_decompiler_lite.zip'
+    $zipSha256 = '099760dcae9daa4c83885a3817cf6f17442ce709de0105993bfbb4f17db87e62'
+
+    $toolDir = Join-Path ${Env:RAW_TOOLS_DIR} $toolName
+    $executablePath = (Join-Path $toolDir "$toolName.exe")
+    VM-Install-With-Installer $toolName $category "EXE" "/VERYSILENT /SUPPRESSMSGBOXES /NORESTART /SP- /Dir=`"$($toolDir)`"" $executablePath $zipUrl -sha256 $zipSha256
+} catch {
+    VM-Write-Log-Exception $_
+}
```
Ana06, line 17: I have just noticed that the `try-catch` is not needed as all code that could fail is inside `VM-Install-With-Installer` that already has a `try-catch` inside. Th code looks a bit nicer:

```suggestion
$toolName = 'VB Decompiler'
$category = 'Visual Basic'

$zipUrl = 'https://www.vb-decompiler.org/files/vb_decompiler_lite.zip'
$zipSha256 = '099760dcae9daa4c83885a3817cf6f17442ce709de0105993bfbb4f17db87e62'

$toolDir = Join-Path ${Env:RAW_TOOLS_DIR} $toolName
$executablePath = (Join-Path $toolDir "$toolName.exe")
VM-Install-With-Installer $toolName $category "EXE" "/VERYSILENT /SUPPRESSMSGBOXES /NORESTART /SP- /Dir=`"$($toolDir)`"" $executablePath $zipUrl -sha256 $zipSha256
```

Ana06 (commented):
@sara-rn just one more minor thing I had missed in my previous review

Ana06 (approved):
I have tested locally again, installation and uninstallation work well. Thank you for all the work @sara-rn!! :cupid:

Issue #1182, opened 2024-11-29
Add Python decompiler


Ana06:
I have had a hard time today analysing a malware sample that loads a byte-like Python code object (with `marshal.loads(obj_bytes)`) and executes it (with `exec()`). After testing several options, I found a way that worked:
1. **Save byte-like Python code object as PYC file**:
    ```Python
    co = marshal.loads(obj_bytes)
    
    # Convert Marshall code object `co` to PYC
    # https://stackoverflow.com/a/73454818/6245337
    import importlib, sys
    pyc_data = importlib._bootstrap_external._code_to_timestamp_pyc(co)
  
    # Write PYC to file
    with open('result.pyc', 'wb') as f:
        f.write(pyc_data)
    ```

2. **Decompile the PYC file**. The object had been compiled with Python 3.10 and there are not many decompilers for it. After trying several tools, these two options worked (the first one worked better than the second) to decompile the PYC file:
   - https://github.com/greyblue9/unpyc37-3.10 
   - The compiled [pycdc.exe](https://github.com/extremecoders-re/decompyle-builds/releases/download/build-16-Oct-2024-5e1c403/pycdc.exe) from https://github.com/extremecoders-re/decompyle-builds/releases/tag/build-16-Oct-2024-5e1c403 

**I propose to add both unpyc37-3.10 and [pycdc.exe](https://github.com/extremecoders-re/decompyle-builds/releases/download/build-16-Oct-2024-5e1c403/pycdc.exe) to FLARE-VM**. Even if unpyc37-3.10 seems to work better for Python 3.10, it does not support Python >3.10 which pycdc. does. @mandiant/flare-vm should we add any other binary from https://github.com/extremecoders-re/decompyle-builds/releases/tag/build-16-Oct-2024-5e1c403 as well?

Ana06:
@mandiant/flare-vm I am using `Utilities` as category, but I think I would have a hard time finding it there. Ideas for a better category?

Ana06:
After some more testing, https://github.com/greyblue9/unpyc37-3.10 worked for me better with Python 3.10 PYC files. We may want to add both.

Which category should we use for these tools? :thinking:

Ana06:
[...]

PR #1186, opened 2024-12-06
Add Python category and unpyc3.vm

Ana06:
Add a new `Python` category to add tools like Python decompilers as proposed in https://github.com/mandiant/VM-Packages/issues/1182.

Add unpyc3.vm that installs with Pip the last commit in the GH repository https://github.com/greyblue9/unpyc37-3.10.

There is an issue of using `py -3.10 -m unpyc.unpyc3` in the shortcut, as it does not get the Python icon:
image

@emtuls I remember we have had this before, but do not remember how/if we fixed it. Can you help me here? :pray: 

This issue partially addresses (as we still need to add [pycdc.exe](https://github.com/extremecoders-re/decompyle-builds/releases/download/build-16-Oct-2024-5e1c403/pycdc.exe)): https://github.com/mandiant/VM-Packages/issues/1182

### DIFF
--- a/.github/ISSUE_TEMPLATE/new_metapackage.yml
+++ b/.github/ISSUE_TEMPLATE/new_metapackage.yml
@@ -75,6 +75,7 @@ body:
         - Persistence
         - Privilege Escalation
         - Productivity Tools
+        - Python
         - Reconnaissance
         - Registry
         - Shellcode
--- a/.github/ISSUE_TEMPLATE/new_node_package.yml
+++ b/.github/ISSUE_TEMPLATE/new_node_package.yml
@@ -59,8 +59,8 @@ body:
         - Debuggers
         - Delphi
         - Disassemblers
-        - dotNet
         - Documents
+        - dotNet
         - Exploitation
         - File Information
         - Forensic
@@ -77,6 +77,7 @@ body:
         - Persistence
         - Privilege Escalation
         - Productivity Tools
+        - Python
         - Reconnaissance
         - Registry
         - Shellcode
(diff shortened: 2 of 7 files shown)


emtuls:
Is `libraries.python3.vm` necessary, or can we just use `python3.vm`? Not a major difference most of the time, but it simplifies what all gets installed when choosing this package. 😃

If what is in the screenshot is acceptable, I would suggest something like the following:
```
$pythonPath = (Get-Command python).Source
VM-Install-Shortcut -toolName $toolName -category $category -executablePath $pythonPath -consoleApp $true -arguments "-3.10 -m unpyc.unpyc3"
```
image

emtuls (commented):
See previous comment

Ana06:
@emtuls 
> Is `libraries.python3.vm` necessary, or can we just use `python3.vm`

It is not strictly necessary, but this ensure that we test there are no conflicts when installing both (which is the default FLARE-VM case). We are requiring `libraries.python3.vm` in many other packages, where we also only need `python3.vm` because of this. `libraries.python3.vm` does not take long to install and does not take much space. So I think it is ok, to ensure it is tested better. 


> ```
> $pythonPath = (Get-Command python).Source
> VM-Install-Shortcut -toolName $toolName -category $category -executablePath $pythonPath -consoleApp $true -arguments "-3.10 -m unpyc.unpyc3"
> ```

Thanks for the proposal @emtuls! :smile:  `python.exe -3.10`, fails to run. We could run `python.exe` without the `3.10` as `(Get-Command python).Source` gets the latest Python version, which should be `3.10`, but if a newer version would be installed it would fail to find the installed `unpyc3`. So I think it is safer to get `py` in `(Get-Command py).Source` which looks like:
image

emtuls (approved):
Looks good! Thanks for making this suggestion. I'm sure we'll add some more Python tools soon. :)

Ana06:
Thanks for the review @emtuls! More Python tools would be awesome! :+1:

PR #1189, opened 2024-12-09
[cyberchef.vm] Add documentation

Ana06:
Document why we can't use the `.html` directly in the shortcut, which is also the reason why we can't use the `VM-Install-From-Zip` helper.

### DIFF
--- a/packages/cyberchef.vm/tools/chocolateyinstall.ps1
+++ b/packages/cyberchef.vm/tools/chocolateyinstall.ps1
@@ -18,6 +18,9 @@ try {
   Install-ChocolateyZipPackage @packageArgs
   VM-Assert-Path $toolDir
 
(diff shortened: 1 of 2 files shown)


packages/cyberchef.vm/tools/chocolateyinstall.ps1
```diff
@@ -18,6 +18,9 @@ try {
   Install-ChocolateyZipPackage @packageArgs
   VM-Assert-Path $toolDir
 
+  # FLARE-VM adds CyberChef to the taskbar.
+  # We use the chrome executable as e can't use an `.html` shortcut for the taskbar.
```
emtuls, line 22: Minor typo: `we` is missing the `w` for `as we can't use an`

Then good to go. :)

emtuls (changes requested):

Issue #1107, opened 2024-07-15
Move the categories to the nuspec & Add wiki page with tools by categories


Ana06:
### Details

Use GH actions to have two Wiki page that documents the tools and categories:
- One page should be for users with a header and a table with package name, tool name and tools description per category. This should make it easier to find the tools that we have available.
- The second page should be for script usage to use it in FLARE-VM/Commando-VM to sort the packages by category in the installer.

Ana06:
I have discussed with @sara-rn the following implementation proposal, which is more robust than the original proposal:
- Add a **single _Packages_ Wiki page with a table with package name, tool name and tools description per category**. This should make it easier to find the tools that we have available. This is addressed in https://github.com/mandiant/VM-Packages/pull/1197.
- **Move the categories to the nuspec**. This way we can use the XML https://www.myget.org/F/vm-packages/Packages directly in the FLARE-VM/Commando-VM installers to sort the packages by category (tracked in FLARE-VM in https://github.com/mandiant/flare-vm/issues/578) and we do NOT need and extra wiki page generated by GH actions. @sara-rn is working on this part. The move of the categories to the nuspec involves:
  - **Add the category to the nuspec of all packages (for example in the [`tags` field](https://learn.microsoft.com/en-us/nuget/reference/nuspec#tags))**. As this touches all packages, I recommend we try to merge all PRs that would cause conflicts before sending this change. I also think this step should be in its own PR (without any other changes even if that means duplicating the categories for a while in two places) to ensure we can merge it fast.
  - **In the packages that use the category to install a shortcut in the `Tools` directory, take it from the nuspec instead of hard-coding it**. I think this should be easy to do, but in case this is not possible, we should add a linter to ensure they are on-sync. In any case, we should **adapt the [category linter](https://github.com/mandiant/VM-Packages/blob/main/scripts/test/lint.py#L316) to ensure the nuspec uses a valid category and that the category is accessed in the correct way**. We may want to split this into two linters. The changes described in this bullet point need to be in the same PR.
  - Adapt the _Packages_ wiki page introduced in https://github.com/mandiant/VM-Packages/pull/1197 to use the category from the nuspec. I think this is a nice improvement, as it removes the `Not Categorized` section making this page more useful.

Ana06:
Part of this issue (moving the categories to the nuspec) is still not addressed.

Ana06:
### More detailed proposal to move category to nuspec
[...]

PR #1197, opened 2024-12-11
Add GitHub Action to Create Package Wiki

emtuls:
This partially resolves https://github.com/mandiant/VM-Packages/issues/1107 by creating a wiki page that contains a table of tools for each category and is updated daily. 

Part 2 will be a second PR.

### DIFF
--- a/.github/workflows/generate_package_wiki.yml
+++ b/.github/workflows/generate_package_wiki.yml
@@ -1,0 +1,37 @@
+name: Generate Package Wiki
+
+on:
+  schedule:
+    # Runs daily at a weird time (02:50 UTC) to avoid delays during periods of
+    # high loads of GitHub Actions workflow runs.
+    - cron: '50 2 * * *'  
+  workflow_dispatch:  # Allow manual triggering
+
+jobs:
+  generate-wiki:
+    runs-on: windows-latest
+    steps:
+      - name: Checkout repository
+        uses: actions/checkout@v3
+
+      - name: Checkout Wiki Repo
+        uses: actions/checkout@v3
+        with:
+          repository: ${{ github.repository }}.wiki
+          path: wiki
+          token: ${{ secrets.REPO_TOKEN }}
+          fetch-depth: 0
+
+      - name: Generate Wiki Content
+        id: generate-content
+        shell: pwsh
+        run: scripts/utils/generate_package_wiki.ps1
+
+      - name: Commit changes
+        working-directory: wiki
+        run: |
+          git config user.email 'vm-packages@google.com'
+          git config user.name 'vm-packages'
+          git add .
+          git commit -am "Update Package Wiki"
+          git push
--- a/.github/workflows/generate_package_wiki.yml
+++ b/.github/workflows/generate_package_wiki.yml
@@ -12,10 +12,10 @@ jobs:
     runs-on: windows-latest
     steps:
       - name: Checkout repository
-        uses: actions/checkout@v3
+        uses: actions/checkout@b4ffde65f46336ab88eb53be808477a3936bae11 # v4.1.1
 
       - name: Checkout Wiki Repo
-        uses: actions/checkout@v3
+        uses: actions/checkout@b4ffde65f46336ab88eb53be808477a3936bae11 # v4.1.1
         with:
           repository: ${{ github.repository }}.wiki
           path: wiki
@@ -24,7 +24,6 @@ jobs:
 
       - name: Generate Wiki Content
         id: generate-content
-        shell: pwsh
         run: scripts/utils/generate_package_wiki.ps1
 
       - name: Commit changes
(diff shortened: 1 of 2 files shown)


.github/workflows/generate_package_wiki.yml
```diff
@@ -0,0 +1,103 @@
...
+              $category = "Not Categorized"
+            }
+
+            if (-not ($packagesByCategory.ContainsKey($category))) {
+              $packagesByCategory[$category] = ""
+            }
+
+
+            try {
+              $descriptionResult = choco info $package -s $source | Select-String -Pattern "Description:\s*(.+)"
```
Ana06, line 59: I think it would be better to parse the nuspec directly instead of using choco. That way we do not need to install choco and rely on it. the nuspec is a yaml file. It would be very simple to parse it using for example a Python script that is called from this action similarly as it is done in https://github.com/mandiant/VM-Packages/blob/main/.github/workflows/daily.yml#L40. This would simplify the code in the action and allow to execute the wiki page generation easily outside of GH actions (for example for testing).

Ana06 (commented):
@emtuls where have you tested the action? Can you please provide a link to a run of the action and a wiki example page?

emtuls:
I've updated the script as per your suggestion and have the run and example on my fork:

Run - https://github.com/emtuls/VM-Packages/actions/runs/12289627139/job/34295427884
Wiki - https://github.com/emtuls/VM-Packages/wiki/Packages

.github/workflows/generate_package_wiki.yml
```diff
@@ -0,0 +1,102 @@
...
+          path: wiki
+          token: ${{ secrets.REPO_TOKEN }}
+          fetch-depth: 0
+
+      - name: Generate Wiki Content
+        id: generate-content
+        shell: pwsh
+        run: |
+          $source = "https://www.myget.org/F/vm-packages/api/v2" 
+
```
Ana06, line 30: unused variable

.github/workflows/generate_package_wiki.yml
```diff
@@ -0,0 +1,102 @@
...
+        with:
+          repository: ${{ github.repository }}.wiki
+          path: wiki
+          token: ${{ secrets.REPO_TOKEN }}
+          fetch-depth: 0
+
+      - name: Generate Wiki Content
+        id: generate-content
+        shell: pwsh
+        run: |
```
Ana06, line 28: I think we should extract this script to a file so that we can run it locally and to simplify the code in the action.

.github/workflows/generate_package_wiki.yml
```diff
@@ -0,0 +1,102 @@
...
+          $source = "https://www.myget.org/F/vm-packages/api/v2" 
+
+          $packagesDir = "packages"
+          $packages = Get-ChildItem -Path $packagesDir -Directory | ForEach-Object { $_.Name } | Sort-Object
+
+          $wikiContent = "# Package List`n`n"
```
Ana06, line 34: I think we should remove the `Package List` header, as the page has a header already and this creates a sublevel also in the index:
# Page
image
# Index
image

Ana06 (changes requested):
Thanks for sharing the run and the wiki example page @emtuls!! :rocket:

emtuls:
I've made the adjustments you suggested, which can be seen here:
Wiki - https://github.com/emtuls/VM-Packages/wiki/Packages
GitHub Action Run - https://github.com/emtuls/VM-Packages/actions/runs/12308409152

.github/workflows/generate_package_wiki.yml
```diff
@@ -0,0 +1,37 @@
...
+    # high loads of GitHub Actions workflow runs.
+    - cron: '50 2 * * *'  
+  workflow_dispatch:  # Allow manual triggering
+
+jobs:
+  generate-wiki:
+    runs-on: windows-latest
+    steps:
+      - name: Checkout repository
+        uses: actions/checkout@v3
```
Ana06, line 15: You should you should pin your GitHub Actions by commit-hash: https://blog.rafaelgss.dev/why-you-should-pin-actions-by-commit-hash (also few lines below)

```suggestion
        uses: actions/checkout@b4ffde65f46336ab88eb53be808477a3936bae11 # v4.1.1
```

scripts/utils/generate_package_wiki.ps1
```diff
@@ -0,0 +1,61 @@
+$packagesDir = "packages"
+$packages = Get-ChildItem -Path $packagesDir -Directory | ForEach-Object { $_.Name } | Sort-Object
+$wikiContent = ""
+$packagesByCategory = @{}
+
+foreach ($package in $packages) {
+$categoryFile = "$packagesDir/$package/tools/chocolateyinstall.ps1"
```
Ana06, line 7: [nitpick] there are some indentation inconsistencies here

Ana06:
I think @emtuls does not have time to address the feedback, I'll add a commit to this PR to address it so that we can merge it.

PR #1199, opened 2024-12-12
Use SigCheck w/ Zimmerman Tools

emtuls:
This fixes many of the tools that are running into an error in our Daily checks (https://github.com/mandiant/VM-Packages/wiki/Daily-Failures). These specific tools run into an issue where we need to manually update the hashes due to the links not having a version in them so our updater does not properly update the packages for us.

Specifically, this fixes:
-  `evtxecmd.vm`
-  `pecmd.vm`
-  `recmd.vm`
-  `registry_explorer.vm`
-  `rla.vm`
-  `sqlecmd.vm`

### DIFF
--- a/packages/evtxecmd.vm/evtxecmd.vm.nuspec
+++ b/packages/evtxecmd.vm/evtxecmd.vm.nuspec
@@ -2,11 +2,11 @@
 <package xmlns="http://schemas.microsoft.com/packaging/2015/06/nuspec.xsd">
   <metadata>
     <id>evtxecmd.vm</id>
-    <version>1.5.0.20240826</version>
+    <version>1.5.0.20241212</version>
     <authors>Eric Zimmerman</authors>
     <description>Event log (evtx) parser with standardized CSV, XML, and json output! Custom maps, locked file support, and more!</description>
     <dependencies>
-      <dependency id="common.vm" version="0.0.0.20240411" />
+      <dependency id="common.vm" version="0.0.0.20241212" />
       <dependency id="dotnet-6.vm" />
     </dependencies>
   </metadata>
--- a/packages/evtxecmd.vm/evtxecmd.vm.nuspec
+++ b/packages/evtxecmd.vm/evtxecmd.vm.nuspec
@@ -6,7 +6,7 @@
     <authors>Eric Zimmerman</authors>
     <description>Event log (evtx) parser with standardized CSV, XML, and json output! Custom maps, locked file support, and more!</description>
     <dependencies>
-      <dependency id="common.vm" version="0.0.0.20241212" />
+      <dependency id="common.vm" version="0.0.0.20250203" />
       <dependency id="dotnet-6.vm" />
     </dependencies>
   </metadata>
--- a/packages/evtxecmd.vm/evtxecmd.vm.nuspec
+++ b/packages/evtxecmd.vm/evtxecmd.vm.nuspec
@@ -8,6 +8,8 @@
     <dependencies>
       <dependency id="common.vm" version="0.0.0.20250203" />
       <dependency id="dotnet-6.vm" />
+      <!-- vcbuildtools.vm installs signtool.exe needed by VM-Assert-Signature -->
+      <dependency id="vcbuildtools.vm" />
     </dependencies>
   </metadata>
 </package>
--- a/packages/pecmd.vm/pecmd.vm.nuspec
+++ b/packages/pecmd.vm/pecmd.vm.nuspec
@@ -6,7 +6,7 @@
     <authors>Eric Zimmerman</authors>
     <description>Prefetch parser</description>
     <dependencies>
-      <dependency id="common.vm" version="0.0.0.20241212" />
+      <dependency id="common.vm" version="0.0.0.20250203" />
       <dependency id="dotnet-6.vm" />
     </dependencies>
   </metadata>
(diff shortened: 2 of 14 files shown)


Ana06 (changes requested):
This is a great improvement, thanks for working on this @emtuls!! :sparkling_heart: 

I do not like that we have a code that is long and complicated repeated in many packages. could you please add a helper function that we can reuse so that the code in every of the packages is simple and the complicated code is in a single place making it easier to maintain and understand? :pray: 

**Question:** Why can't we use `VM-Assert-Signature` for RegCool? It seem you didn't need to do anything special in this case :thinking:

emtuls:
> This is a great improvement, thanks for working on this @emtuls!! 💖
> 
> I do not like that we have a code that is long and complicated repeated in many packages. could you please add a helper function that we can reuse so that the code in every of the packages is simple and the complicated code is in a single place making it easier to maintain and understand? 🙏
> 
> **Question:** Why can't we use `VM-Assert-Signature` for RegCool? It seem you didn't need to do anything special in this case 🤔

I've modified the `VM-Install-From-Zip` to include a new argument for verifying a signature. I was a bit hesitant at first to do it because it seems to make the code look a little bit ugly, but I think it is generally fine and makes for the other code bits in each package to be much cleaner.

The alternative would have been its own function, but it would mimic the `VM-Install-From-Zip` function in a lot of ways, so I wasn't sure if that would be any better.


As for why we can't use it with `RegCool`. I believe that this is because the cert used to sign that tool is a self signed cert which is not verified via our method that we use to verify signatures, whereas these tools have a standard CA signing authority.

rltpoa:
@Ana06 I totally agree with the idea of a "better" code. But meanwhile, new users like me are struggling to complete a fresh install of Flare-VM without running into issues. I think it comes down to weighing both sides... what might be more important at the moment: a not-so-pretty code that could be improved later, or new users having a smooth experience? 

Anyway, sorry to interrupt, but I thought perhaps the perspective of a new user could add to the discussion. Thank you both for all this work.

packages/pecmd.vm/tools/chocolateyinstall.ps1
```diff
@@ -7,4 +7,4 @@ $category = 'Forensic'
 $zipUrl = 'https://download.mikestammer.com/net6/PECmd.zip'
 $zipSha256 = 'e20254b2f813e66fe5295488e5a00e9675679c91841f99ddcc8d083299bb55d6'
 
-VM-Install-From-Zip $toolName $category $zipUrl -zipSha256 $zipSha256 -consoleApp $true -innerFolder $false
+VM-Install-From-Zip $toolName $category $zipUrl -zipSha256 $zipSha256 -consoleApp $true -innerFolder $false -verifySignature $true
```
Ana06, line 10: you need to make the hash optional and remove it from the packages using signature verification:
 ```suggestion
VM-Install-From-Zip $toolName $category $zipUrl -consoleApp $true -innerFolder $false -verifySignature $true
```

File: packages/pecmd.vm/pecmd.vm.nuspec [hunk 2]
```diff
@@ -2,11 +2,11 @@
 <package xmlns="http://schemas.microsoft.com/packaging/2015/06/nuspec.xsd">
   <metadata>
     <id>pecmd.vm</id>
-    <version>1.5.0.20240826</version>
+    <version>1.5.0.20241212</version>
     <authors>Eric Zimmerman</authors>
     <description>Prefetch parser</description>
     <dependencies>
-      <dependency id="common.vm" version="0.0.0.20240411" />
+      <dependency id="common.vm" version="0.0.0.20241212" />
```
Ana06, line 9: you also need to ensure signtool is installed in all the packages that use signature verification
```suggestion
      <dependency id="common.vm" version="0.0.0.20241212" />
      <!-- vcbuildtools.vm installs signtool.exe needed by VM-Assert-Signature -->
      <dependency id="vcbuildtools.vm" />
```

Ana06 (changes requested):
Question: Can we now use the helper function to simplify the packages that are already using signature verification (sysinternals, googlechrome, metasploit)?

Ana06:
I am going to take over as it seems @emtuls does not have time :wink:

packages/evtxecmd.vm/evtxecmd.vm.nuspec
```diff
@@ -2,12 +2,14 @@
 <package xmlns="http://schemas.microsoft.com/packaging/2015/06/nuspec.xsd">
   <metadata>
     <id>evtxecmd.vm</id>
-    <version>1.5.0.20240826</version>
+    <version>1.5.0.20241212</version>
     <authors>Eric Zimmerman</authors>
     <description>Event log (evtx) parser with standardized CSV, XML, and json output! Custom maps, locked file support, and more!</description>
     <dependencies>
-      <dependency id="common.vm" version="0.0.0.20240411" />
+      <dependency id="common.vm" version="0.0.0.20241212" />
```
sara-rn, line 9: shouldn't be the version of `common.vm` `0.0.0.20250203` to make sure it accepts the `-verifySignature` parameter?

packages/pecmd.vm/pecmd.vm.nuspec
sara-rn: `<dependency id="common.vm" version="0.0.0.20250203" />`

sara-rn (commented):
I tested the installation of the packages locally and I got the same error for all. I believe the version of the dependency `common.vm` might need to be updated.
`ERROR: A parameter cannot be found that matches parameter name 'verifySignature'.`

Ana06:
@sara-rn it is surely related to that, as if an older version is installed, the new function introduced in this PR is not available. I forgot to update the version when fixing the conflict in the common nuspec. Good catch!

Ana06:
Thanks for the review @sara-rn!

PR #1200, opened 2024-12-12
Fix RegCool Hash

emtuls:
This fixes the hash for `RegCool.vm` as we have to do this manually due to the download link not containing a version number for us to identify when the tool gets updated.


Ana06 (changes requested):
Thanks for updating the hash @emtuls! It seems you have added an unrelated commit by accident. Could you please remove it? :smile:

Ana06 (approved):

PR #1203, opened 2024-12-13
Update hash for event-log-explorer

emtuls:
This updates the hash for `event-log-explorer.vm` so that it will stop failing to install.


Ana06 (approved):
LGTM :+1:

PR #1207, opened 2024-12-13
Add xrefer.vm

Ana06:
Very excited to add the newly released Xrefer IDA plugin developed by @m-umairx, member of the FLARE team. This plugin provides a custom navigation interface for analyzing execution paths, function clusters, and related artifacts, leveraging Google's Gemini model for natural language descriptions. :sparkles:  In case you want to know more about this useful tool, @m-umairx has written a blog post that is about to be released in https://cloud.google.com/blog/topics/threat-intelligence!! :hourglass: 

I would love to add this tool to the default FLARE-VM configuration. But there is a small issue we have to figure out first: When you open IDA, the python library `asciinet` (dependency of XRefer) calls `java.exe` that opens a java console that keeps open the whole time. :disappointed:  @mandiant/vms does someone have any idea of the best way to force it to use `javaw.exe` instead? `openjdk` is a dependency of many other important tools like Ghidra, so I don't want to change anything that could break the other tools.
image
image


Ana06:
Blog post is released: https://cloud.google.com/blog/topics/threat-intelligence/xrefer-gemini-assisted-binary-navigator :tada:

Issue #1217, opened 2024-12-16
Internet detector set taskbar to always be pink


Ana06:
### Details

The internet detector saves the original taskbar color to restore it when internet is disabled.  @emtuls implemented it this way to be able to support light and dark modes. When the internet is enabled and the computer is restarted, the taskbar color is pink (because the internet is enabled) and this causes the original color to get assigned to pink, keeping the taskbar always pink independent of the internet status.

When installing FLARE-VM, because the VM is restarted during the installation, the taskbar is set the pink. Even if it is installed afterwards (not as part of the default configuration), a restart with internet enable will set it to always be pink.

This makes the tool more confusing than helpful, as we can't trust it to know when the internet is enabled.

Ana06:
Ideas to fix the issue:
1. When starting the tool, check if the original color is saved in the registry, if not check the current color. If it is not pink, save it in the registry. When restoring, if no value is saved in the registry, use a default (for example gray).
2. Only support using a default color which combines well with the FLARE-VM image (for example dark green) for internet disabled and pink when enabled. This simplifies the implementation and if we select a cool color it would be a nice improvement! @emtuls could you try several colors (for example different tones of green, black, and blue) and add some screenshots of 2-3 options you think look good (ideally that the same color look good in black and light themes)?

Ana06:
Fixed in https://github.com/mandiant/VM-Packages/pull/1221

Issue #1216, opened 2024-12-16
Internet adapter in a broken state (unable to update desktop image and icon in the taskbar)


Ana06:
### Details

I have installed FLARE-VM, which by default installs `internet_detector.vm`. Because the internet is on (we are installing packages), the `internet_detector` changes the theme to pink and the background.  But the installer set the background image at the end to be the FLARE-VM image, overwriting the one from the installer. After that, I would expect the `internet_detector` to overwrite it again, but it does not, keeping the bar and desktop out of sync.

![image](https://github.com/user-attachments/assets/41dff0ac-1762-44c8-a356-162a05a143e3)

This makes the tool more confusing than helpful, as we can't trust it to know when the internet is enabled.

Ana06:
Documenting the details from https://github.com/mandiant/VM-Packages/pull/1221#issue-2743997133 here too:
The issue with the wallpaper seem to be that `internet_detector` ended up in a broken state after the `installer.vm` performed a `VM-Refresh-Desktop`, which made `internet_detector` continue to run but unable to modify the background and also display an indicator in the taskbar. This can be fixed by having a detection for if `explorer.exe` is restarted and/or if the taskbar icon is removed, and if so, it attempts to restart itself or die gracefully so that the scheduled task can start it back up again properly.

Ana06:
Fixed in https://github.com/mandiant/VM-Packages/pull/1221

PR #1221, opened 2024-12-17
Fix Internet Detector Background/Taskbar Issues

emtuls:
This fixes both https://github.com/mandiant/VM-Packages/issues/1216 and https://github.com/mandiant/VM-Packages/issues/1217

The issue with the wallpaper (https://github.com/mandiant/VM-Packages/issues/1216) was actually that `internet_detector` ended up in a broken state after the `installer.vm` performed a `VM-Refresh-Desktop`, which made `internet_detector` continue to run but unable to modify the background and also display an indicator in the taskbar. This has been fixed by having a detection for if `explorer.exe` is restarted and/or if the taskbar icon is removed, and if so, it attempts to restart itself or die gracefully so that the scheduled task can start it back up again properly.

The Taskbar issue (https://github.com/mandiant/VM-Packages/issues/1217) has been resolved by storing the old/default taskbar color scheme into a registry key that will be checked if it exists before setting a new default. It then reads this value for when it adjusts back to the state of when it does not detect internet.

I have also adjusted the package to re-enable the scheduled task and also made the time to kick off every 1 minute instead of 2 minutes.

### DIFF
--- a/packages/internet_detector.vm/tools/internet_detector.pyw
+++ b/packages/internet_detector.vm/tools/internet_detector.pyw
@@ -28,7 +28,6 @@ import time
 import os
 import re
 
-
 # Define constants
 CHECK_INTERVAL = 2  # Seconds
 CONNECT_TEST_URL_AND_RESPONSES = {
@@ -62,6 +61,7 @@ default_palette = None
 # Win32 API icon handles
 hicon_indicator_off = None
 hicon_indicator_on = None
+mutex = None
 
 def is_already_running():
     global mutex
@@ -75,7 +75,7 @@ def is_already_running():
         return False  # Assume not running if error
 
 def signal_handler(sig, frame):
-    global check_thread, tray_icon_thread, tray_icon
+    global check_thread, tray_icon_thread, tray_icon, mutex
     print("Ctrl+C detected. Exiting...")
     stop_event.set()  # Signal the background thread to stop
     if check_thread:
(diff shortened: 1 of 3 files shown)


packages/internet_detector.vm/tools/internet_detector.pyw
```diff
@@ -410,14 +466,18 @@ def main_loop():
...
     default_transparency = get_transparency_effects()
-    default_color_prevalence = get_current_color_prevalence()
-    default_color = get_current_taskbar_color()
-    default_palette = get_current_color_palette()
+
+    # Try to load default settings from the registry
+    if not load_default_settings():
+        # If not found, get the current settings and save them as defaults
+        default_color_prevalence = get_current_color_prevalence()
+        default_color = get_current_taskbar_color()
```
Ana06, line 478: I think we should check before saving the color that it is not pink and use a default gray if it is. This way we ensure that there can never be an error, as these bug show how tricky implementing it correct is.

Ana06: In the end, with this implementation we are setting a default (the one the user had before the tool started running), as the user can't change the color anymore. So I wonder, if it would be simpler and more reliable to just default to black and pink.

If we keep it this way, we should document somewhere (not necessary in this PR, maybe in the wiki) how the user could change the color. In the long term we could move this tool to its own project with a readme.

Ana06 (approved):
I have tested locally and it seems to work well :+1: I think we can merge it and continue discussing if we need further improvements afterwards. I expect our community to share feedback too. :wink:

PR #1227, opened 2024-12-19
Remove double console when opening fakenet

emtuls:
When opening Fakenet via the shortcut, it spawns a second command prompt which runs Fakenet.

The original intent was to have Fakenet opened in a cmd that would not close after terminating Fakenet, but unfortunately this doesn't seem to be something that can be done via execution through a shortcut, rather, I believe it needs to be done from within Fakenet.

This PR will put fakenet back to its original way where it opens a single command prompt, but will terminate the cmd after closing fakenet.


Ana06 (approved):
Even if we run it as a console app (keeping the second console), the one with the fakenet output get closed when stopping fakenet. So I agree we shouldn't run it as a console app. I'll open an issue in fakenet asking to be able to keep the output after stopping fakenet with a future fakenet version.

PR #1235, opened 2025-01-10
[wireshark.vm] Add tshark to PATH and Tools dir

Ana06:
Wireshark installs tshark, but it was not added to PATH. Add it to PATH and to the Tools directory to make it easier to find.

### DIFF
--- a/packages/wireshark.vm/tools/chocolateyinstall.ps1
+++ b/packages/wireshark.vm/tools/chocolateyinstall.ps1
@@ -7,9 +7,15 @@ try {
 
   $shortcutDir = Join-Path ${Env:TOOL_LIST_DIR} $category
   $shortcut = Join-Path $shortcutDir "$toolName.lnk"
-  $executablePath = Join-Path ${Env:ProgramFiles} "Wireshark\Wireshark.exe" -Resolve
+  $executablePath = Join-Path ${Env:ProgramFiles} "Wireshark\$toolName.exe" -Resolve
   Install-ChocolateyShortcut -shortcutFilePath $shortcut -targetPath $executablePath -RunAsAdmin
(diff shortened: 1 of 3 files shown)


packages/wireshark.vm/tools/chocolateyinstall.ps1
```diff
@@ -7,9 +7,15 @@ try {
...
-  $executablePath = Join-Path ${Env:ProgramFiles} "Wireshark\Wireshark.exe" -Resolve
+  $executablePath = Join-Path ${Env:ProgramFiles} "Wireshark\$toolName.exe" -Resolve
   Install-ChocolateyShortcut -shortcutFilePath $shortcut -targetPath $executablePath -RunAsAdmin
   VM-Assert-Path $shortcut
+
+  # Add tshark (installed by Wireshark) to PATH and to Tools directory
+  $toolName = 'tshark'
+  $executablePath = Join-Path ${Env:ProgramFiles} "Wireshark\$toolName.exe" -Resolve
+  Install-BinFile -Name $toolname -Path $executablePath
+  VM-Install-Shortcut -toolName $toolName -category $category -executablePath $executablePath -consoleApp $true
```
sara-rn, line 18: the shortcut executable is `tshark.exe` already starts capturing on the first interface, perhaps it would be easier to execute with `--help`?

sara-rn (commented):
tested locally, everythin looks good. I just added a minor comment regarding the shortcut

packages/wireshark.vm/tools/chocolateyinstall.ps1
Ana06: Good point! Any other opinions @mandiant/flare-vm?

PR #1242, opened 2025-01-17
Require version in tools installed by pip

### LINKED ISSUE #1240: Fix version of tools installed by pip
### Details

We have started using a version in tools installed with pip (see [magika package](https://github.com/mandiant/VM-Packages/blob/c2c7a01d014d50f58b7b063f684af3760427701e/packages/magika.vm/tools/chocolateyinstall.ps1#L8)). We should do the same with the rest of the packages using `VM-Install-With-Pip`. This requires:
- Make the `version` argument in `VM-Install-With-Pip` compulsory
- Adapt the packages using `VM-Install-With-Pip` to use a version both in chocolateyinstall and in the nuspec (they currently use `0.0.0.YYYYMMDD` as there is not concrete version and should use the tool version now that we now which one it is).

This was mentioned in https://github.com/mandiant/VM-Packages/issues/1080#issuecomment-2523816211, but as it can (and I think should) be addressed in a separate PR, I think it is better to create a separate it issue to track it.

sara-rn:
Closes #1240

Change the version parameter to required and change tools installed with PIP to include the version

### DIFF
--- a/packages/common.vm/tools/vm.common/vm.common.psm1
+++ b/packages/common.vm/tools/vm.common/vm.common.psm1
@@ -1771,8 +1771,8 @@ function VM-Install-With-Pip {
         [string] $toolName, # Example: magika
         [Parameter(Mandatory=$true)]
         [string] $category,
-        [Parameter(Mandatory=$false)]
-        [string] $version = "", # Version using pip format, example: "==0.5.0"
+        [Parameter(Mandatory=$true)]
+        [string] $version,  # Version using pip format, example: "==0.5.0"
         [Parameter(Mandatory=$false)]
         [string] $arguments = "--help"
     )
--- a/packages/ipython.vm/ipython.vm.nuspec
+++ b/packages/ipython.vm/ipython.vm.nuspec
@@ -2,7 +2,7 @@
 <package xmlns="http://schemas.microsoft.com/packaging/2015/06/nuspec.xsd">
   <metadata>
     <id>ipython.vm</id>
-    <version>8.27.0.20241001</version>
+    <version>8.27.0.20250122</version>
     <authors>Fernando Perez</authors>
     <description>A powerful interactive Python shell</description>
     <dependencies>
--- a/packages/autoit-ripper.vm/autoit-ripper.vm.nuspec
+++ b/packages/autoit-ripper.vm/autoit-ripper.vm.nuspec
@@ -2,7 +2,7 @@
 <package xmlns="http://schemas.microsoft.com/packaging/2015/06/nuspec.xsd">
   <metadata>
     <id>autoit-ripper.vm</id>
-    <version>0.0.0.20240607</version>
+    <version>1.1.2</version>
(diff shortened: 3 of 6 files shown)


Ana06 (changes requested):
Thanks for adding the versions @sara-rn!! :bouquet: As noticed by the linter, you have forgotten to update the version in the ipython.vm nuspec. Also, ensure to follow good commit practices such as limiting the subject line to 50 characters and using the body to explain what and why you make the changes. I recommend reading https://cbea.ms/git-commit for more info on best practices to write good commit messages. :wink:

packages/common.vm/tools/vm.common/vm.common.psm1
```diff
@@ -1761,7 +1761,7 @@ function VM-Install-With-Pip {
         [string] $toolName, # Example: magika
         [Parameter(Mandatory=$true)]
         [string] $category,
-        [Parameter(Mandatory=$false)]
+        [Parameter(Mandatory=$true)]
```
Ana06, line 1764: the common nuspec also needs update. Please run the linter locally or check the CI status after sending the changes.

Ana06 (changes requested):

packages/common.vm/tools/vm.common/vm.common.psm1
Ana06: This will cause a conflict with https://github.com/mandiant/VM-Packages/pull/1243 which also updates common. As the other PR has more changes (and is more difficult to rebase), I suggest you wait to do this change after that PR has been merged, to avoid the merge conflicts.

Ana06: Yes, from the [Wiki](https://github.com/mandiant/VM-Packages/wiki/Chocolatey-packages#version):
> **The version needs to be increased every time you modify an existing package.**

Otherwise the package won't be pushed to MyGet, as there is already a package with that version.

packages/ipython.vm/ipython.vm.nuspec
```diff
@@ -2,7 +2,7 @@
 <package xmlns="http://schemas.microsoft.com/packaging/2015/06/nuspec.xsd">
   <metadata>
     <id>ipython.vm</id>
-    <version>8.27.0.20241001</version>
+    <version>8.27.0</version>
```
Ana06, line 5: you have decreased the version (by removing the date) causing that our CI is testing the old package version without the version, which now fails as it is required. Ensure you increase the version:
```suggestion
    <version>8.27.0.20250122</version>
```

Ana06 (changes requested):

packages/ipython.vm/ipython.vm.nuspec
sara-rn: @Ana06 changes have been addressed

Ana06 (approved):
The CI is finally happy, thanks @sara-rn for all the changes! :blush:

Issue #1244, opened 2025-01-17
Add help comment to VM- user facing functions


Ana06:
### Details

Add help comment to VM- user facing functions using PowerShell conventions/format: https://learn.microsoft.com/en-us/powershell/module/microsoft.powershell.core/about/about_comment_based_help?view=powershell-7.4

PR #1249, opened 2025-01-17
[common.vm] Improve VM-Write-Log

Ana06:
Improve console output of `VM-Write-Log`:
- Use `$message` instead of `$line` as it makes the output difficult to read. This information is only useful in the file.
- Use better colors.

Also remove unused levels (`FATAL`, `DEBUG`) and improve the function help message.

### Example: idapro.vm before the change
image

### Example: idapro.vm after the change
image

idapro.vm added in https://github.com/mandiant/VM-Packages/pull/1243

Related: https://github.com/mandiant/VM-Packages/issues/1244


sara-rn (approved):
tested locally, works as expected

Issue #1222, opened 2024-12-17
visualstudio may not finish installing all components


Ana06:
### Details

I get the a timeout error (more than 45 minutes) when installing [visualstudio.vm](https://github.com/mandiant/VM-Packages/tree/main/packages/visualstudio.vm), which comes from running the following command:
```
choco install visualstudio2022community --params "--add Microsoft.VisualStudio.Component.CoreEditor --add Microsoft.VisualStudio.Workload.NativeDesktop --add Microsoft.VisualStudio.Workload.ManagedDesktop --includeRecommended"
```
The package still gets installed correctly. I suspect this is because of the bug https://github.com/chocolatey/choco/issues/3487 (which in this case is a feature :sweat_smile:). 

This is the relevant part of the output:
```
Progress: 100% - Completed download of C:\Users\flare\AppData\Local\ChocoCache\visualstudio2022community\117.12.2\vs_Community.exe (4.23 MB).
Download of vs_Community.exe (4.23 MB) completed.
Hashes match.
Installing visualstudio2022community...

Chocolatey timed out waiting for the command to finish. The timeout
 specified (or the default value) was '2700' seconds. Perhaps try a
 higher `--execution-timeout`? See `choco -h` for details.
  visualstudio2022community may be able to be automatically uninstalled.
The install of visualstudio2022community was NOT successful.
Error while running 'C:\ProgramData\chocolatey\lib\visualstudio2022community\tools\ChocolateyInstall.ps1'.
 See log for details.

Chocolatey installed 0/1 packages. 1 packages failed.
 See the log for details (C:\ProgramData\chocolatey\logs\chocolatey.log).

Failures
 - visualstudio2022community (exited -1) - Error while running 'C:\ProgramData\chocolatey\lib\visualstudio2022community\tools\ChocolateyInstall.ps1'.
 See log for details.
```

The error persist if I install [visualstudio.vm](https://github.com/mandiant/VM-Packages/tree/main/packages/visualstudio.vm) with `--execution-timeout 6000`, likely because `visualstudio2022community` is not installed as a dependency because of the arguments that we need to provide. We could add the `--execution-timeout 6000` to the `visualstudio2022community`  installation command, not sure if that would fix it. This change could also increase the daily running time. @mandiant/vms  any ideas? Is there a better way to pass the arguments?

PR #1250, opened 2025-01-20
[visualstudio.vm] Increase execution timeout

Ana06:
Increase execution timeout for visualstudio2022community to ensure it is installed in slow environments if the execution timeout for visualstudio.vm is also increased.

Fixes https://github.com/mandiant/VM-Packages/issues/1222 (I hope, it is tricky to test). There may also be better solutions, please let me know if you have any ideas.


sara-rn (approved):
tested locally, looks good to me

PR #1253, opened 2025-01-22
Update ilspy.vm & add dotnet-8.vm

Ana06:
Update ilspy.vm to 9.0.0. As ILSpy requires .NET 8, add a new metapackage for .NET8 to ensure all packages use the same version (dotnet-8.vm). Use dotnet-8.vm in arsenalimagemounter.vm, the only package requiring .NET8 as dependency at the moment. 

It would be good if merge this PR in the next hours or our bot will try to update ILSpy without the .NET8 update (it does not run without the dependency) later today.


sara-rn (approved):
tested locally, looks good to me

PR #1254, opened 2025-01-23
Fix 7z wrong password empty output

### LINKED ISSUE #1135: Confusing `unzip "infected"` with wrong password
### Details

Show error when using `unzip "infected"` if the password is wrong. If the ZIP is not password protected and you use the `unzip "infected"` option, it still extracts the ZIP correctly. But if it is password protected and the password is not `infected`, then the files may be extracted and be empty, which is very confusing. Research displaying an error and/or not extracting the files in this case. @mandiant/vms ideas?

d35ha:
Closes #1135.


packages/7zip-15-05.vm/tools/chocolateyinstall.ps1
```diff
@@ -31,10 +31,14 @@ try {
   VM-Install-Shortcut $toolName $category $executablePath
 
   # Add 7z unzip with password "infected" to the right menu for the most common extensions.
+  $7zCommand = "`"$7zExecutablePath`" t -pinfected `"%*`" && `"$7zExecutablePath`" x -pinfected `"%*`""
+  $7zCommandBat = Join-Path ${Env:RAW_TOOLS_DIR} "7z_unzip_infected.bat"
+  "@echo off`n" + $7zCommand | Out-File -FilePath $7zCommandBat -Encoding UTF8
```
Ana06, line 36: Why is a batch script needed (vs using the `$7zCommand` directly in `VM-Add-To-Right-Click-Menu`) and what does the batch script do? This needs to be clarified in the code (with comments for example).

Ana06: [nitpick] reorder lines to have the generation of the bat file in consequtives lines makes it IMO  easier to read
```suggestion
  $7zCommandBat = Join-Path ${Env:RAW_TOOLS_DIR} "7z_unzip_infected.bat"
  $7zCommand = "`"$7zExecutablePath`" t -pinfected `"%*`" && `"$7zExecutablePath`" x -pinfected `"%*`""
  "@echo off`n" + $7zCommand | Out-File -FilePath $7zCommandBat -Encoding UTF8
```

Ana06 (changes requested):
Thanks for working on this @d35ha! Before, if the password was different from `infected`, the files were extracted and were empty. Now, the files are not extracted at all. I think this is a better than what we had before, but I think it does not fix https://github.com/mandiant/VM-Packages/pull/1254. Not extracting anything without any warning message is also confusing IMO.

packages/7zip-15-05.vm/tools/chocolateyinstall.ps1
d35ha: @Ana06 
Done.

d35ha:
> Not extracting anything without any warning message is also confusing IMO.

@Ana06 

A warning message is shown now in case of error.

packages/7zip-15-05.vm/tools/chocolateyinstall.ps1
```diff
@@ -30,11 +30,21 @@ try {
   $executablePath = Join-Path $toolDir "${toolName}FM.exe" -Resolve
   VM-Install-Shortcut $toolName $category $executablePath
 
+  # Test if the passowrd is right before extracting the files requiring two 7z commands with
+  # the two options `t` and `x`.
```
Ana06, line 34: There is a typo (`passowrd`). I also think we should mention here that the integrity check (option `t`) also checks the password. Proposal:

```suggestion
  # Test integrity of the archive (which also check that the password is correct) before extraction
```

packages/7zip-15-05.vm/tools/chocolateyinstall.ps1
```diff
@@ -30,11 +30,21 @@ try {
...
+  $7zCommand = @'
+	cmd /c (7z t -pinfected "%1" || (
```
Ana06, line 36: I think you need to hide the console, so that the user only seems the mesage box with the error:
image

packages/7zip-15-05.vm/tools/chocolateyinstall.ps1
```diff
@@ -30,11 +30,21 @@ try {
...
+		mshta vbscript:Execute^(
+			"msgbox ""The password is not """"infected"""" or the archive is corrupted !!"":close"
```
Ana06, line 38:
```suggestion
			"msgbox ""The password is not """"infected"""" or the archive is corrupted!!"":close"
```

Ana06: I think we should also add a title to the message bog, for example: `ERROR 7z`

packages/7zip-15-05.vm/tools/chocolateyinstall.ps1
```diff
@@ -30,11 +30,21 @@ try {
...
+		^) && call
+	)) && 7z x -pinfected "%1"
+'@ -replace "`t", "" -replace "`n", ""
```
Ana06, line 41: Cool idea to make the code more readable! The alternative would be to have the code in a file. But as it is a small command, I think this is good too. :+1:

Ana06 (changes requested):
You have updated 7zip-15-05.vm, but note that there is a 7zip.vm package (the 7zip.vm package we require in all packages and the default FLARE-VM config) that installs 7zip-nsis.vm. This means that the package that is used in FLARE-VM default configuration is 7zip-nsis.vm and not 7zip-15-05.vm. If I remember correctly, this is implemented in this way as we were not sure about migrating from 7zip-15-05.vm to 7zip-nsis.vm and this would allow us to easily change from one to the other if needed. As this point I feel this is not needed anymore. So I suggest keeping only 7zip.vm and use it to install 7zip-nsis, simplifying the code and removing code duplication. Steps:

- Remove 7zip-15-05.vm
- Remove 7zip-nsis.vm and move its chocolateyinstall.ps1 code to 7zip.vm (please fix the current indentation issues when moving this code)
- Add the enhacement to make unzip "infected" with wrong password more user friendly to 7zip.vm

@mandiant/flare-vm anything against getting rid of 7zip-15-05.vm?

Ana06:
@d35ha do not assign the PR to the milestone if the issue is already on it to avoid tracking things twice.

packages/7zip-15-05.vm/tools/chocolateyinstall.ps1
d35ha: @Ana06 

The console provides more information to the user that would be very helpful if the unpacking has failed due to a different reason (ex. the archive is corrupted), I prefer to keep it as long as it closes once the Ok button is pressed.

packages/7zip-15-05.vm/tools/chocolateyinstall.ps1
d35ha: Done.

packages/7zip-15-05.vm/tools/chocolateyinstall.ps1
d35ha: Done.

packages/7zip-15-05.vm/tools/chocolateyinstall.ps1
Ana06: I would prefer to remove it as in that output I do not see any extra useful information (only that the password is wrong what the other message already shows). @mandiant/vms any other opinions?

Ana06 (commented):
Thanks for addressing the changes @d35ha! 

@d35ha I think you have missed my previous comment and it is important: https://github.com/mandiant/VM-Packages/pull/1254#pullrequestreview-2574783304

packages/7zip-15-05.vm/tools/chocolateyinstall.ps1
d35ha: @Ana06 

It's useful in case of a corrupted archive.

Screenshot from 2025-02-04 16-07-28

d35ha:
> You have updated 7zip-15-05.vm, but note that there is a 7zip.vm package (the 7zip.vm package we require in all packages and the default FLARE-VM config) that installs 7zip-nsis.vm. This means that the package that is used in FLARE-VM default configuration is 7zip-nsis.vm and not 7zip-15-05.vm. If I remember correctly, this is implemented in this way as we were not sure about migrating from 7zip-15-05.vm to 7zip-nsis.vm and this would allow us to easily change from one to the other if needed. As this point I feel this is not needed anymore. So I suggest keeping only 7zip.vm and use it to install 7zip-nsis, simplifying the code and removing code duplication. Steps:
> 
> * Remove 7zip-15-05.vm
> * Remove 7zip-nsis.vm and move its chocolateyinstall.ps1 code to 7zip.vm (please fix the current indentation issues when moving this code)
> * Add the enhacement to make unzip "infected" with wrong password more user friendly to 7zip.vm
> 
> @mandiant/vms anything against getting rid of 7zip-15-05.vm?

@Ana06 
Now we only have 7zip.vm.

packages/7zip-15-05.vm/tools/chocolateyinstall.ps1
Ana06: I also don't see any extra useful information in this case (only that the archive is corrupted that the message already shows). In my opinion it is more elegant to only have the message box and let the user to debug the issue (by running 7z in the console for example) on its own. @mandiant/vms what do others think?

Ana06 (approved):
Thanks for all the work @d35ha! :bouquet:

PR #1255, opened 2025-01-23
Extend the wait time for loading the chrome extensions

### LINKED ISSUE #1223: Chrome extensions open tabs
### Details

Some of the added chrome extensions automatically open a tab when chrome is opened, which is confusing. We need to find a way to fix this.

image

d35ha:
Closes #1223.

### DIFF
--- a/packages/chrome.extensions.vm/tools/chocolateyinstall.ps1
+++ b/packages/chrome.extensions.vm/tools/chocolateyinstall.ps1
@@ -68,6 +68,9 @@ try {
     $tries += 1
   }
 
+  # Wait for an extra 60 seconds to make sure all extensions' tabs are opened already.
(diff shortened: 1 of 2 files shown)


packages/chrome.extensions.vm/tools/chocolateyinstall.ps1
```diff
@@ -68,6 +68,9 @@ try {
     $tries += 1
   }
 
+  # Wait for an extra 60 seconds to make sure all extensions' tabs are opened already.
+  Start-Sleep -Seconds 60
```
Ana06, line 72: it seems to me like the time could be reduced a bit, but better to play it safe. :thinking:

Ana06 (approved):
Thanks for the fix @d35ha! :+1:

PR #1258, opened 2025-01-24
Add try-catch to VM-Pip-Install

### LINKED ISSUE #1205: Add `try-catch` to `VM-Pip-Install`
### Details

Add `try-catch` to `VM-Pip-Install` so that packages using this function don't have to add a `try-catch`. A good example are IDA plugins like [ida.plugin.hashdb.vm](https://github.com/mandiant/VM-Packages/blob/260d4bae802ada3ae6bf23737732ecb53ada0a0b/packages/ida.plugin.hashdb.vm/tools/chocolateyinstall.ps1#L6) using the `VM-Install-IDA-Plugin` function that install dependencies using `VM-Pip-Install`. `VM-Install-IDA-Plugin` has the `try-catch` but we need it in the package because of the code in `VM-Pip-Install`.

sara-rn:
Remove try-catch from install scripts that use VM-Pip-Install
Closes #1205


Ana06 (changes requested):
Thanks for working on this @sara-rn.

[nitpick] The commit message is a bit confusing and it does not explain **why** the change is needed. I propose to improve it as follows:

```
Add try-catch to VM-Pip-Install

Add missing `try-catch` to `VM-Pip-Install` so that packages using this
function (such as ida.plugin.hashdb.vm) don't have to add a try-catch. 
```

sara-rn:
@Ana06 thanks for the review, I changed the ida.plugin.capa.vm nuspec version back to the previous one because the chocolatelyinstall was also changed to a previous state but then of course the lint complains now cause I haven't changed the version

sara-rn:
@Ana06 changes have been addressed, thanks for the review

Ana06 (approved):
Thanks @sara-rn!

PR #1261, opened 2025-01-28
Use ICMP instead of HTTP for testing the connection

d35ha:


packages/fakenet-ng.vm/tools/chocolateyinstall.ps1
```diff
@@ -37,7 +37,7 @@ try {
 
   # Replace `default.ini` with our modified one that includes change for 'internet_detector'.
   # IMPORTANT: Keep our modified `default.ini` in-sync on updates to package.
-  $fakenetConfigDir = Get-ChildItem "C:\Tools\fakenet\*\configs"
+  $fakenetConfigDir = @(Get-ChildItem "C:\Tools\fakenet\*\configs")[0]
```
Ana06, line 40: why is this change needed?

I think the following is a bit nicer to read than the current and the propose code:
```suggestion
  $fakenetConfigDir = Resolve-Path "C:\Tools\fakenet\*\configs"
```

packages/internet_detector.vm/tools/chocolateyinstall.ps1
```diff
@@ -11,7 +11,7 @@ New-Item -Path $toolDir -ItemType Directory -Force -ea 0
 VM-Assert-Path $toolDir
 
 # Install pyinstaller 6.11.1 (needed to build the Python executable with a version capable of executing in admin cmd) and tool dependencies ('pywin32')
-$dependencies = "pyinstaller==6.11.1,pywin32"
+$dependencies = "pyinstaller==6.11.1,pywin32,icmplib"
```
Ana06, line 14: I think we should use a concrete version of both pywin32 and icmplib to avoid issues with newer versions.

packages/internet_detector.vm/tools/internet_detector.pyw
```diff
@@ -20,8 +20,7 @@ import winerror
 import winreg
 
 import threading
-import requests
-import urllib3
+import icmplib
```
Ana06, line 23: as you are one using one function:

```suggestion
from icmplib import ping
```
or:
```suggestion
from icmplib import multiping
```

packages/internet_detector.vm/tools/internet_detector.pyw
```diff
@@ -306,12 +305,12 @@ def extract_title(data):
...
 
 def check_internet():
-    for url, expected_response in CONNECT_TEST_URL_AND_RESPONSES.items():
+    for dns_address in CONNECT_TEST_DNS_ADDRESSES:
         try:
             # Perform internet connectivity tests
-            response = requests.get(url, timeout=5, verify=False)
-            if expected_response in (extract_title(response.text) or response.text):
-                print(f"Internet connectivity detected via URL: {url}")
+            dns_host = icmplib.ping(dns_address, count=1, timeout=CHECK_INTERVAL)
```
Ana06, line 311: `CHECK_INTERVAL` is used to determine how often to repeat the internet check. I think it should not be used here. It is also irrelevant if you only perform one ping (which you do with `count=1`).

Ana06: [idea] you can also use `icmplib.multiping` and simplify the code. This is a bit more noisy (as it performs all the requests every time), but it is also consistent and maybe clearer to understand for the user. opinions @d35ha @mandiant/flare-vm?
```python
def check_internet():
    try:
        hosts = multiping(dns_address, count=1, timeout=CHECK_INTERVAL)
        return any(host.is_alive for host in hosts)
    except:
        return False
```

As this removes the print message, I would add a print message after the call to this function in `check_internet_and_update_tray_icon()`.

packages/internet_detector.vm/tools/internet_detector.pyw
```diff
@@ -30,12 +29,12 @@ import re
 
 # Define constants
 CHECK_INTERVAL = 2  # Seconds
-CONNECT_TEST_URL_AND_RESPONSES = {
-    "https://www.msftconnecttest.com/connecttest.txt": "Microsoft Connect Test",  # HTTPS Test #1
-    "http://www.google.com": "Google",  # HTTP Test
-    "https://www.wikipedia.com": "Wikipedia",  # HTTPS Test #2
-    "https://www.youtube.com": "YouTube",  # HTTPS Test #3
-}
+CONNECT_TEST_DNS_ADDRESSES = [
```
Ana06, line 32: [nitpick] what about shortening this variable name?
```suggestion
TEST_IPs = [
```

packages/internet_detector.vm/tools/internet_detector.pyw
```diff
@@ -30,12 +29,12 @@ import re
...
+    "8.8.8.8",  # Google
```
Ana06, line 33: [question] can you please explain the selection of IPs? Our purpose is to select 2-4 IPs from which at least one work if there any kind of internet connectivity and that look benign (to avoid confusing internet detector activity with malware activity). I am trying to understand your reasoning being the selection to consider if we can improve anything here.

Ana06 (changes requested):
Nice improvement @d35ha! I think you should document (both in the code as comment and in the commit message description), why this is an enhacement (how this reduces the fingerprint of the tools and is a better option than HTTPS and TCP), as this is what makes this a nice improvement and we want to ensure others understand it (now and in the future to avoid it is changed back).

packages/fakenet-ng.vm/tools/chocolateyinstall.ps1
d35ha: @Ana06 

Actually `Get-ChildItem` returns a list of files, it caused a problem when I tried to upgrade the installed fakenet to 3.3

packages/internet_detector.vm/tools/internet_detector.pyw
d35ha: I think it's consistent with the other imports.

packages/internet_detector.vm/tools/internet_detector.pyw
d35ha: Done.

packages/internet_detector.vm/tools/internet_detector.pyw
d35ha: It is the largest public DNS servers in the world with very little down time. Using ICMP is a similar approach to the technique used by the [InternetCheckConnectionA](https://learn.microsoft.com/en-us/windows/win32/api/wininet/nf-wininet-internetcheckconnectiona) API.

Ana06: I think using ICMP is a good idea that needs to be documented in the code as a comment.

But my question was about the reason to select these concrete 4 IPs. These needs to be clarified in the code (with a comment for example). This ensure others reading your code in the future understand why it was implemented this way.

packages/internet_detector.vm/tools/chocolateyinstall.ps1
```diff
@@ -11,7 +11,7 @@ New-Item -Path $toolDir -ItemType Directory -Force -ea 0
 VM-Assert-Path $toolDir
 
 # Install pyinstaller 6.11.1 (needed to build the Python executable with a version capable of executing in admin cmd) and tool dependencies ('pywin32')
-$dependencies = "pyinstaller==6.11.1,pywin32"
+$dependencies = "pyinstaller==6.11.1,pywin32==306,icmplib==3.0.4"
```
Ana06, line 14: is there a reason to use pywin32 version 306 (vs the last one)? Please if so, document it in a comment.

packages/fakenet-ng.vm/tools/chocolateyinstall.ps1
Ana06: I see, there could be more than one version installed. You are in that case modifying the config of the lowest version (selecting it with `[0]`). Ideally we should modify only the one of the version that is installed. We have several options:
- [IDEAL] Hardcode the version in this path and modify our automatic package update script to replace the version not only in the URL but everywhere in the chocolateyinstall scripts.
- [WORKAROUND] Assume the version installed is the highest one and select it here with `[-1]`. This is easier to implement and it works for clean installs, but it could update the wrong fakenet log if there are several versions installed.
- [WORKAROUND] Modify all the configs (looping all the directories in the list).  This is easier to implement and it works for clean installs, but you could break the already installed fakenet version(s).

Any of the options are ok with me. Please add a comment documenting this discussion in the code as this is not straigtforward.

[nitpick] As we want the path and we do not use anything else from the child item objects, I think it is clearer to use `Resolve-Path` (which returns an array of strings) than `Get-ChildItem` (which returns an array of child item objects).

packages/internet_detector.vm/tools/chocolateyinstall.ps1
d35ha: It's the version I have and it's working fine, no specific reason but I'll update it to the latest version.

packages/internet_detector.vm/tools/internet_detector.pyw
d35ha: Done.

packages/fakenet-ng.vm/tools/chocolateyinstall.ps1
d35ha: Hardcoding the version makes more since.

d35ha: The version update is already handled at: https://github.com/mandiant/VM-Packages/blob/93ce8d3f2f7abcaa3d800c1fa2be8767c7c5d471/scripts/utils/update_package.py#L95-L127.

packages/internet_detector.vm/tools/chocolateyinstall.ps1
Ana06: I was just curious if there was a reason. If there is no reason, as long as it works well, it is ok.

scripts/utils/update_package.py
```diff
@@ -39,9 +39,9 @@ def get_latest_version(org, project, version):
     latest_version = response.json()["tag_name"]
     # version excludes `v` from the capturing group in the regex in update_github_url therefore latest_version_match mustn't include `v` if the version starts with `v`. Otherwise the github URL would replace the version without the `v` with the github version tag with the `v` which will result in the wrong URL such as: https://github.com/jstrosch/sclauncher/releases/download/vv0.0.6/sclauncher.exe
     if latest_version.startswith('v'):
-    	return latest_version[1:]
+        return latest_version[1:]
```
Ana06, line 42: unrelated change, please remove it in order to keep a clean commit history

Ana06 (changes requested):
Thank you for the changes! As already mentioned in my last review: You need to document why this is an enhacement (how this reduces the fingerprint of the tools and is a better option than HTTPS and TCP). Please do this in the following ways:
- Add a description to your commit. I recommend reading https://cbea.ms/git-commit for more info on best practices to write good commit messages
- Update the comment at the beginning of `internet_detector.pyw`. Note it is wrong ATM: `This tool checks if internet connectivity exists by reaching out to specific website`.

packages/internet_detector.vm/tools/internet_detector.pyw
```diff
@@ -30,12 +29,14 @@ import re
 
 # Define constants
 CHECK_INTERVAL = 2  # Seconds
-CONNECT_TEST_URL_AND_RESPONSES = {
-    "https://www.msftconnecttest.com/connecttest.txt": "Microsoft Connect Test",  # HTTPS Test #1
-    "http://www.google.com": "Google",  # HTTP Test
-    "https://www.wikipedia.com": "Wikipedia",  # HTTPS Test #2
-    "https://www.youtube.com": "YouTube",  # HTTPS Test #3
-}
+# Testing the internet connection by pinging some of
```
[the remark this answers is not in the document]
Ana06: I have just enabled it in FLARE-VM in https://github.com/mandiant/flare-vm/pull/650. Planning to enabling it in this repository after this PR has been merged (to avoid merge conflicts in this PR)

packages/fakenet-ng.vm/tools/chocolateyinstall.ps1
```diff
@@ -37,8 +37,7 @@ try {
 
   # Replace `default.ini` with our modified one that includes change for 'internet_detector'.
   # IMPORTANT: Keep our modified `default.ini` in-sync on updates to package.
-  $fakenetConfigDir = Get-ChildItem "C:\Tools\fakenet\*\configs"
-  Copy-Item "$packageToolDir\default.ini" -Destination $fakenetConfigDir
+  Copy-Item "$packageToolDir\default.ini" -Destination "C:\Tools\fakenet\fakenet3.3\configs"
```
Ana06, line 40: This path should use `RAW_TOOLS_DIR` (see documentation in https://github.com/mandiant/VM-Packages/tree/main/packages/common.vm#user-configurable). The issue was there in the previous code, but I would appreciate if you could correct it:
```suggestion
  Copy-Item "$packageToolDir\default.ini" -Destination "{$Env:RAW_TOOLS_DIR}\fakenet\fakenet3.3\configs"
```

sara-rn:
I tested locally and it works as expected. As suggested by @Ana06 it is not known why this implementation poses an advantage over HTTPS

packages/fakenet-ng.vm/tools/chocolateyinstall.ps1
sara-rn: I understand that `default.ini` didn't need to change as the exclusions for `internet_detector` where already in place

packages/internet_detector.vm/tools/internet_detector.pyw
```diff
@@ -306,12 +307,12 @@ def extract_title(data):
...
 
 def check_internet():
-    for url, expected_response in CONNECT_TEST_URL_AND_RESPONSES.items():
+    for ip_address in TEST_IPS:
         try:
             # Perform internet connectivity tests
-            response = requests.get(url, timeout=5, verify=False)
-            if expected_response in (extract_title(response.text) or response.text):
-                print(f"Internet connectivity detected via URL: {url}")
+            ip_host = icmplib.ping(ip_address, count=1, timeout=CHECK_INTERVAL)
```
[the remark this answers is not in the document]
sara-rn: perhaps you can use another constant for the timeout, even if it is set to 2, same as CHECK_INTERVAL. I agree with Ana on this

sara-rn (commented):
tested locally, works fine!! Thanks @d35ha.
Just to add to my previous comment, for someone without a background perhaps it is hard to know why this implementation is more efficient than the previous one. Just a bit of explanation would be good.

packages/fakenet-ng.vm/tools/chocolateyinstall.ps1
Ana06: @sara-rn 
> It's my understanding that `default.ini` didn't need to change as the exclusions for `internet_detector` where already in place. Just pointed it out cause there were no changes to `default.ini` in this commit.

@sara-rn see the discussion in https://github.com/mandiant/VM-Packages/pull/1261#discussion_r1931990053. There was a bug affecting updates in this code and that's why @d35ha modified it here. 

@sara-rn's comment make it clear that a comment above this code mentioning the concrete issue would help understand this line and to prevent that it is changed back causing again issues with updates.

packages/internet_detector.vm/tools/internet_detector.pyw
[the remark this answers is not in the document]
d35ha: Done.

packages/internet_detector.vm/tools/internet_detector.pyw
d35ha: Done.

scripts/utils/update_package.py
d35ha: Done.

packages/internet_detector.vm/tools/internet_detector.pyw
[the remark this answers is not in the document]
d35ha: The default value is just used.

packages/fakenet-ng.vm/tools/chocolateyinstall.ps1
d35ha: Done.

Ana06 (commented):
Thanks for adding documentation in `internet_detector.vm/tools/internet_detector.pyw` @d35ha! :bouquet: 

Please always test your changes locally. The package fails to install.

I would appreciate if you could add a description to your commit messages. This is an important change and it would be helpful when checking the commit history to know why a change was introduced. As already suggested in my previous reviews, I recommend you check https://cbea.ms/git-commit. Concretely:
- the first line (the title) should ideally be <=50 characters
- after the title leave an empty line. The rest is the description, which should explain what and why.  

An example of one of my last PRs with several commits (every of them with a commit description): https://github.com/mandiant/flare-vm/pull/645

packages/fakenet-ng.vm/tools/chocolateyinstall.ps1
Ana06: I think the documentation here is still missing. what about something like the following
```
# Replace `default.ini` with our modified one that includes change for 'internet_detector'.
# IMPORTANT: Keep our modified `default.ini` in-sync on updates to package.
# Do not remove the version in the path to avoid replacing the config file of another version.
```

packages/internet_detector.vm/tools/internet_detector.pyw
```diff
@@ -30,12 +29,19 @@ import re
...
-    "https://www.wikipedia.com": "Wikipedia",  # HTTPS Test #2
-    "https://www.youtube.com": "YouTube",  # HTTPS Test #3
-}
+
+# - ICMP is a faster and a more-efficient way for checking the connection
+#   as it has a minimal fingerprint of 2 packets (echo/reply) per request.
+# - IP addresses are used instead of well-known websites or domains so
+#   no DNS resolution is needed.
+# - The used IP addresses are some of the largest public DNS servers to
+#   ensure zero or minimal downtime.
```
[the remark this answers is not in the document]
d35ha: @Ana06 
Added at https://github.com/mandiant/VM-Packages/commit/96ec867d9eb940ae95674c17a6e5584ea1fb1107#commitcomment-152062424.

packages/fakenet-ng.vm/tools/chocolateyinstall.ps1
d35ha: Added.

Ana06 (commented):
@d35ha thank you for the changes! I appreciate the documentation addition and the detailed commit description.

I have done some more testing and this change is very noisy in fakenet:
image

It seems the `ProcessBlackList: internet_detector.exe` fakenet exclusion is not working properly for ICMP. The HTTP communication was excluded from the output, while ICMP is still there. @d35ha can you reproduce this? Am I doing something wrong?

Ana06:
@sara-rn you also tested the change. Did you also get the noise in fakenet?

d35ha:
> @d35ha thank you for the changes! I appreciate the documentation addition and the detailed commit description.
> 
> I have done some more testing and this change is very noisy in fakenet: image
> 
> It seems the `ProcessBlackList: internet_detector.exe` fakenet exclusion is not working properly for ICMP. The HTTP communication was excluded from the output, while ICMP is still there. @d35ha can you reproduce this? Am I doing something wrong?

@Ana06 
I noticed the same noise, not sure why fakenet is unable to identify the source process (maybe related to https://github.com/mandiant/flare-vm/pull/630).

d35ha:
> > @d35ha thank you for the changes! I appreciate the documentation addition and the detailed commit description.
> > I have done some more testing and this change is very noisy in fakenet: image
> > It seems the `ProcessBlackList: internet_detector.exe` fakenet exclusion is not working properly for ICMP. The HTTP communication was excluded from the output, while ICMP is still there. @d35ha can you reproduce this? Am I doing something wrong?
> 
> @Ana06 I noticed the same noise, not sure why fakenet is unable to identify the source process (maybe related to [mandiant/flare-vm#630](https://github.com/mandiant/flare-vm/pull/630)).

It seems that fakenet failes to (or doesn't) identify the source process for the ICMP requests.

Ana06:
PR to add ICMP ID-based filtering in fakenet that allows us to ignore the ICMP traffic from the internet detector: https://github.com/mandiant/flare-fakenet-ng/pull/204

Ana06:
We would still need to adapt this PR after https://github.com/mandiant/flare-fakenet-ng/pull/204 has been merged to add the ID, @d35ha do you plan to do this now to illustrate how this is exactly implemented (concretely how the ID is generate), which will help understanding what we want to address with https://github.com/mandiant/flare-fakenet-ng/pull/204 (and may help with the PR review)? Or are you planning to wait for feedback on https://github.com/mandiant/flare-fakenet-ng/pull/204 before making changes here?

d35ha:
@Ana06 

Now the fakenet filter is enabled from the tool side, the only thing we are waiting for is https://github.com/mandiant/flare-fakenet-ng/pull/204 to be approved.

Ana06:
@d35ha fakenet has been released and updated in VM-Packages. Please rebase and ensure the code works with the new fakenet version and the linter is happy. :wink:

d35ha:
@Ana06 rebased and successfully tested.

Ana06 (changes requested):

packages/fakenet-ng.vm/tools/default.ini
```diff
@@ -116,9 +116,12 @@ DefaultUDPListener:    ProxyUDPListener
 BlackListPortsTCP: 139
 BlackListPortsUDP: 67, 68, 137, 138, 443, 1900, 5355
 
+# Specify ICMP IDs to be ignored when diverting the packets.
```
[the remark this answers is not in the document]
d35ha: Done.

packages/internet_detector.vm/tools/internet_detector.pyw
```diff
@@ -36,6 +36,7 @@ CHECK_INTERVAL = 2  # Seconds
 #   no DNS resolution is needed.
 # - The used IP addresses are some of the largest public DNS servers to
 #   ensure zero or minimal downtime.
+ICMP_ID = 1234
```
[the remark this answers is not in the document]
Ana06: [question] What is the advantage of generating the value randomly during the installation (it is hardcoded for that installation) vs just having the same hardcoded value for all installations? Are you trying to prevent malware to use this ID to difficult analysis? Could there be any issues with conflicts with other applications?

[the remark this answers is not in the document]
Ana06: I think this is a good idea, can you please add a comment above the code that replaces the ICMP_ID in `chocolateyinstall.ps1` documenting why it is randomized?

sara-rn:
I tested locally and I realized that this issue https://github.com/mandiant/VM-Packages/pull/1261#issuecomment-2633106481 still persists, shouldn't be Fakenet able to identify the source process for the ICMP requests?

packages/internet_detector.vm/internet_detector.vm.nuspec
```diff
@@ -2,13 +2,13 @@
...
     <id>internet_detector.vm</id>
-    <version>1.0.0.20250219</version>
+    <version>1.0.0.20250415</version>
     <authors>Elliot Chernofsky and Ana Martinez Gomez</authors>
     <description>Tool that changes the background and a taskbar icon if it detects internet connectivity</description>
     <dependencies>
       <dependency id="common.vm" version="0.0.0.20250206" />
       <dependency id="libraries.python3.vm" version="0.0.0.20240726" />
-      <dependency id="fakenet-ng.vm" version="3.3" />
+      <dependency id="fakenet-ng.vm" version="3.5" />
```
Ana06, line 11:
```suggestion
      <dependency id="fakenet-ng.vm" version="3.5.0.20250415" />
```

packages/internet_detector.vm/tools/internet_detector.pyw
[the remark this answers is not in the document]
d35ha: Done.

d35ha:
> I tested locally and I realized that this issue [#1261 (comment)](https://github.com/mandiant/VM-Packages/pull/1261#issuecomment-2633106481) still persists, shouldn't be Fakenet able to identify the source process for the ICMP requests?

@sara-rn please make sure fakenet 3.5 is installed and 3.3 is totally removed (from `C:\Tools\fakenet\fakenet3.3`) as `fakenet.vm` will pick the old version if exists:

https://github.com/mandiant/VM-Packages/blob/d7e532fe2193e2d2a51f58bf5c6bf8eb6b12044a/packages/fakenet-ng.vm/tools/chocolateyinstall.ps1#L31-L32

Ana06:
@d35ha 
> @sara-rn please make sure fakenet 3.5 is installed and 3.3 is totally removed (from C:\Tools\fakenet\fakenet3.3) as fakenet.vm will pick the old version if exists

Can we modify fakenet.vm to take the latest instead of the oldest version instead? That would avoid this type of issues.

d35ha:
> @d35ha
> 
> > @sara-rn please make sure fakenet 3.5 is installed and 3.3 is totally removed (from C:\Tools\fakenet\fakenet3.3) as fakenet.vm will pick the old version if exists
> 
> Can we modify fakenet.vm to take the latest instead of the oldest version instead? That would avoid this type of issues.

Modified it.

Ana06:
@sara-rn can you please test it again? :smile:

packages/fakenet-ng.vm/tools/chocolateyinstall.ps1
```diff
@@ -29,17 +28,12 @@ try {
 
   # There is an inner folder in the zip whose name changes as it includes the version
   $dirList = Get-ChildItem $toolDir -Directory
-  $toolDir = Join-Path $toolDir $dirList[0].Name -Resolve
+  $toolDir = Join-Path $toolDir $dirList[-1].Name -Resolve
```
Ana06, line 31: [nitpick] I think we need to document why this `-1` is important
```
# There is an inner folder in the zip whose name changes as it includes the version
# Select the latest fakenet version to prevent conflict with older installed versions
```

Ana06 (approved):
Awesome work @d35ha!  This is a really clever approach to making our internet detector more robust.  :sparkles:  Thanks for all your work in this PR and in the modification of fakenet! :bouquet: I have tested locally and everything looks good to me. :rocket:  The code looks excellent too (I have just one more documentation nitpick you could quickly address before merging it).

sara-rn:
tested locally, LGTM, thanks @d35ha great job

packages/fakenet-ng.vm/tools/chocolateyinstall.ps1
d35ha: Done.

Ana06 (approved):
We are ready to merge it! :partying_face:

PR #1263, opened 2025-01-30
Add arguments to script and template to install tools

### LINKED ISSUE #1082: Add arguments parameter to script + issue templates
### Details

Add an extra `arguments` parameter to the [`create_package_template.py` script ](https://github.com/mandiant/VM-Packages/blob/main/scripts/utils/create_package_template.py) and the issue templates. This would allow to create packages with arguments automatically and make easier to implement more templates and automation (such as https://github.com/mandiant/VM-Packages/issues/1047).

sara-rn:
Add the optional `arguments` parameter to the install script and ISSUE templates.
The wrapper install script must translate the `parameter` to `parameter=` because the 
value might contain `-` or `--`.
Closes #1082

### DIFF
--- a/.github/ISSUE_TEMPLATE/new_node_package.yml
+++ b/.github/ISSUE_TEMPLATE/new_node_package.yml
@@ -103,6 +103,13 @@ body:
       description: |
         The tool description should be short and not include version specific details or other information that is likely to change in a future version. Example: `Deobfuscator to remove common JS obfuscation techniques.`.
   - type: input
+    id: arguments
+    attributes:
+      label: Arguments
+      description: |	
+        Command-line arguments to pass to the tool when running it from the `Tools` directory. For example `--help`.
+      value: --help
+  - type: input
     id: dependencies
     attributes:
       label: Dependencies
--- a/.github/ISSUE_TEMPLATE/new_package.yml
+++ b/.github/ISSUE_TEMPLATE/new_package.yml
@@ -135,6 +135,13 @@ body:
       description: |
         SHA256 hash of the `.zip` file downloaded from the download url introduced in the previous field. The hash is used for verification purposes. Example: `62af5cce80dbbf5cdf961ec9515549334a2112056d4168fced75c892c24baa95 `
   - type: input
+    id: arguments
+    attributes:
+      label: Arguments
+      description: |	
+        Command-line arguments to pass to the tool when running it from the `Tools` directory. For example `--help`.
+      value: --help
+  - type: input
     id: dependencies
     attributes:
       label: Dependencies
--- a/.github/ISSUE_TEMPLATE/new_python_tool_pip.yml
+++ b/.github/ISSUE_TEMPLATE/new_python_tool_pip.yml
@@ -103,6 +103,13 @@ body:
       description: |
         The tool description should be short and not include version specific details or other information that is likely to change in a future version. Example: `FakeNet-NG is a dynamic network analysis tool.`
   - type: input
+    id: arguments
+    attributes:
+      label: Arguments
+      description: |	
+        Command-line arguments to pass to the tool when running it from the `Tools` directory. For example `--help`.
+      value: --help
+  - type: input
     id: dependencies
     attributes:
       label: Dependencies
--- a/scripts/utils/create_package_template_from_json.py
+++ b/scripts/utils/create_package_template_from_json.py
@@ -25,8 +25,8 @@ def main():
     for k, v in pkg.items():
         if k in ("why", "dependencies", "info"):
             continue
-        cmd_args.append(f"--{k}")
-        cmd_args.append(f"{v}")
+        # Use "--argument=<value>" format to avoid issues with if the value contains "-" 
+        cmd_args.append(f"--{k}={v}")
 
     create_package_template.main(cmd_args)
 
--- a/scripts/utils/create_package_template.py
+++ b/scripts/utils/create_package_template.py
@@ -107,8 +107,9 @@ $category = '{category}'
 
 $zipUrl = '{target_url}'
 $zipSha256 = '{target_hash}'
+$arguments = '{arguments}'
 
-VM-Install-From-Zip $toolName $category $zipUrl -zipSha256 $zipSha256 -consoleApp ${console_app} -innerFolder ${inner_folder}
+VM-Install-From-Zip $toolName $category $zipUrl -zipSha256 $zipSha256 -consoleApp ${console_app} -innerFolder ${inner_folder} -arguments $arguments
 """
 
 """
@@ -120,8 +121,9 @@ Import-Module vm.common -Force -DisableNameChecking
 
 $toolName = '{tool_name}'
 $category = '{category}'
+$arguments = '{arguments}'
 
-VM-Install-Node-Tool -toolName $toolName -category $category -arguments "--help"
+VM-Install-Node-Tool -toolName $toolName -category $category -arguments $arguments
 """
 
 """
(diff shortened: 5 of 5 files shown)


sara-rn:
Example of actions execution with the new templates with parameters: 
floss (`VM-Install-from-Zip`): https://github.com/sara-rn/VM-Packages/actions/runs/13054986426
sliver (`VM-Install-Single-Exe`): https://github.com/sara-rn/VM-Packages/actions/runs/13031630625
js-deobfuscator (`VM-Install-Node-Tool`): https://github.com/sara-rn/VM-Packages/actions/runs/13030787843
capesolo (`VM-Install-With_Pip`): https://github.com/sara-rn/VM-Packages/actions/runs/13030158547

.github/ISSUE_TEMPLATE/new_node_package.yml
```diff
@@ -46,6 +46,12 @@ body:
       placeholder: ex. 0.0.0.20240513
     validations:
       required: true
+  - type: input
+    id: arguments
+    attributes:
+      label: Arguments
+      description: |	
+        Tool's execution argument. When executed which argument will be passed to the destination Shortcut File. If no argument is provided the default is `--help`.
```
Ana06, line 54: **Question: how can we provide no arguments if leaving it empty defaults to `--help`. I think no arguments should be passed if it is empty and instead we can use the `default` option in the ISSUE TEMPLATE and make it required.**

I find the description difficult to understand. Many users filling the template may not know what the _destination Shortcut File_ is. 

```suggestion
      description: Specify arguments to be passed to run the tool. Example: `--help`
      default: --help
      validations:
        required: true
```

This comment applies to the other templates too.

scripts/utils/create_package_template_from_json.py
```diff
@@ -25,8 +25,12 @@ def main():
...
         if k in ("why", "dependencies", "info"):
             continue
-        cmd_args.append(f"--{k}")
-        cmd_args.append(f"{v}")
+        #the value of arguments may contain - or -- therefore we add --argument=<value>
+        if k == "arguments":	
+            cmd_args.append(f"--{k}={v}")
+        else:
+            cmd_args.append(f"--{k}")
+            cmd_args.append(f"{v}")
```
Ana06, line 33: What about always using the `--argument=<value>` format? I think it is in general more robust and it simplifies the code. Or is there any issue with this approach? :thinking: 

```suggestion
        # Use "--argument=<value>" format to avoid issues with if the value contains "-" 
        cmd_args.append(f"--{k}={v}")
```

Ana06 (changes requested):
Looking forward to being able to provide args in ISSUE TEMPLATES! Thanks for your work on this feature @sara-rn :bouquet:

.github/ISSUE_TEMPLATE/new_node_package.yml
sara-rn: @Ana06 as far as I know, we cannot make `arguments` required in `new_template` because not all the functions have the `arguments`, only for  `ZIP_EXE` and `SINGLE_EXE`, and for the node and PIP templates. I agree, the default should be `` instead of `--help`. Let me know if I misunderstood anything.

sara-rn: How about description: `Command-line execution arguments. These arguments will be passed when the tool is executed from the Tools directory`
Let me know if you will rephrase it so I changed in all the templates.

Ana06: @sara-rn you are right that we can't make it compulsory. We can still use `default: --help` and use `""` as default.

Regarding the description, I think mentioning the shortcut as you propose is helpful. I think adding an example (as I was proposing) is helpful too. So something similar to the following should be good. :wink: 
```
Command line arguments to pass to the tool when running it from the `Tools` directory. For example: `--help`.
```

Ana06 (commented):
@sara-rn could you please provide an issue example for every type of issue in your fork where we can see the action working? They do not need to be new tools if you don't find one at the moment (for example you can create a `capa2` package just for testing).

sara-rn:
@Ana06 these are the issue example for each type:
https://github.com/sara-rn/VM-Packages/issues/21
https://github.com/sara-rn/VM-Packages/issues/19
https://github.com/sara-rn/VM-Packages/issues/17
https://github.com/sara-rn/VM-Packages/issues/15

GH actions runs:
https://github.com/sara-rn/VM-Packages/actions/runs/13112775467
https://github.com/sara-rn/VM-Packages/actions/runs/13112779747
https://github.com/sara-rn/VM-Packages/actions/runs/13112781973
https://github.com/sara-rn/VM-Packages/actions/runs/13112784954

sara-rn:
@Ana06 instead of `default` I added `placeholder` to be consistent with the other fields, also for some reason I didn't get it to display it

.github/ISSUE_TEMPLATE/new_node_package.yml
```diff
@@ -46,6 +46,13 @@ body:
       placeholder: ex. 0.0.0.20240513
     validations:
       required: true
+  - type: input
+    id: arguments
+    attributes:
+      label: Arguments
+      description: |	
+        Command-line arguments to pass to the tool when running it from the `Tools` directory. For example `--help`.
+      placeholder: ex. --help
```
Ana06, line 55: > @Ana06 instead of default I added placeholder to be consistent with the other fields, also for some reason I didn't get it to display it

**We use placeholder to display an example without filling the field. What I was suggesting here is to have it pre-filled** as in many cases `--help` is the right value, so that it is pre-filled. `default` is the wrong attribute, as it is available for `dropdown`, but not for `input` type. **You should use `value`**. See GH documentation for more details: https://docs.github.com/en/communities/using-templates-to-encourage-useful-issues-and-pull-requests/syntax-for-githubs-form-schema#attributes-for-input.
```suggestion
      placeholder: ex. --help
```

This applies to the other templates too.

.github/ISSUE_TEMPLATE/new_node_package.yml
```diff
@@ -46,6 +46,13 @@ body:
       placeholder: ex. 0.0.0.20240513
     validations:
       required: true
+  - type: input
```
Ana06, line 49: [nitpick] I think the arguments should be placed just after `Tool's description` and before `Dependencies`. It feels for me more natural to first fill the required fields like `Category`.

.github/ISSUE_TEMPLATE/new_package.yml
```diff
@@ -62,6 +62,13 @@ body:
       placeholder: ex. 1.4.11
     validations:
       required: true
+  - type: input
```
Ana06, line 65: [nitpick] I think the arguments should be placed just after `Download SHA256 Hash` and before `Dependencies`. It feels for me more natural to first fill the required fields like `Category` and `Download URL`.

.github/ISSUE_TEMPLATE/new_python_tool_pip.yml
```diff
@@ -46,6 +46,13 @@ body:
       placeholder: ex. 1.4.11
     validations:
       required: true
+  - type: input
```
Ana06, line 49: [nitpick] I think the arguments should be placed just after `Tool's description` and before `Dependencies`. It feels for me more natural to first fill the required fields like `Category`.

Ana06 (commented):
@sara-rn thanks for sharing the examples, they look good! :rocket:

sara-rn:
@Ana06 changes have been addressed

Ana06 (approved):
Awesome work @sara-rn! :bouquet:

PR #1265, opened 2025-02-03
[microsoft-office.vm] Improve installation params & add verification + shortcuts 

Ana06:
Install a concrete version `ProPlus2024Retail` to avoid untested updates. Remove the language to use the system default. Ensure Office tools useful for malware analysis are installed and create
shortcuts in the Tools directory to make accessing them easier.


Ana06:
Interesting. I think windows-2019 is 64-bits, but `(Get-ComputerInfo).OSArchitecture` is returning something different, causing that the path does not exists. :thinking:

PR #1266, opened 2025-02-04
Avoid error when installing packages in headless mode

thack1:
VM-Refresh-Desktop tries to restart explorer.exe by using the Stop-Process cmdlet. When installing VM-Packages without a desktop session running (e.g. using ansible) this results in an error, because explorer.exe is not running. This error can be avoided by using ErrorAction SilentlyContinue. 

`ERROR: Cannot find a process with name "explorer".`

### DIFF
--- a/packages/common.vm/common.vm.nuspec
+++ b/packages/common.vm/common.vm.nuspec
@@ -2,7 +2,7 @@
 <package xmlns="http://schemas.microsoft.com/packaging/2015/06/nuspec.xsd">
   <metadata>
     <id>common.vm</id>
-    <version>0.0.0.20250203</version>
+    <version>0.0.0.20250204</version>
     <description>Common libraries for VM-packages</description>
     <authors>Mandiant</authors>
   </metadata>
--- a/packages/common.vm/tools/vm.common/vm.common.psm1
+++ b/packages/common.vm/tools/vm.common/vm.common.psm1
@@ -1564,8 +1564,10 @@ public class Shell {
         $SHCNE_ASSOCCHANGED = 0x08000000
         $SHCNF_IDLIST = 0
         [void][Shell]::SHChangeNotify($SHCNE_ASSOCCHANGED, $SHCNF_IDLIST, [IntPtr]::Zero, [IntPtr]::Zero)
-        # Refresh the Taskbar
-        Stop-Process -Name explorer -Force  # This restarts the explorer process so that the new taskbar is displayed.
+        # Restart the explorer process to refresh the taskbar.
+        # Use '-ErrorAction Continue' to avoid failing the package installation if it fails,
+        # for example if there is no desktop session (as in this case explorer.exe is not running)
+        Stop-Process -Name explorer -Force -ErrorAction Continue
     } catch {
         VM-Write-Log-Exception $_
     }


packages/common.vm/common.vm.nuspec
```diff
@@ -2,7 +2,7 @@
 <package xmlns="http://schemas.microsoft.com/packaging/2015/06/nuspec.xsd">
   <metadata>
     <id>common.vm</id>
-    <version>0.0.0.20250124</version>
+    <version>0.0.0.20250204</version>
```
Ana06, line 5: I have just merged a PR that updated this package, so there is now a merge conflict here. @thack1 can you please fix it?

packages/common.vm/tools/vm.common/vm.common.psm1
```diff
@@ -1534,7 +1534,7 @@ public class Shell {
         $SHCNF_IDLIST = 0
         [void][Shell]::SHChangeNotify($SHCNE_ASSOCCHANGED, $SHCNF_IDLIST, [IntPtr]::Zero, [IntPtr]::Zero)
         # Refresh the Taskbar
-        Stop-Process -Name explorer -Force  # This restarts the explorer process so that the new taskbar is displayed.
+        Stop-Process -Name explorer -Force -ErrorAction SilentlyContinue  # This restarts the explorer process so that the new taskbar is displayed.
```
Ana06, line 1537: I think using `-ErrorAction Continue` would still prevent this from failing (fixing the issue you are encoutering) and display the error, which I think it is in general useful.
```suggestion
        Stop-Process -Name explorer -Force -ErrorAction Continue  # This restarts the explorer process so that the new taskbar is displayed.
```

Ana06: [nitpick] I think it would be a good idea to document why the `-ErrorAction Continue` is needed (to avoid we forget about the reason and delete it). The current comments are a bit confusing (as we have two comments saying the same, one more detailed). Proposal:

```PS
# Restart the explorer process to refresh the taskbar.
# Use '-ErrorAction Continue' to avoid failing the package installation if it fails,
# for example if there is no desktop session (as in this case explorer.exe is not running)
 Stop-Process -Name explorer -Force -ErrorAction Continue
```

Ana06 (changes requested):
Thank for your contribution @thack1! You need to sign Google CLA, otherwise we are not able to merge your PR. 

Please also rebase to fix the version update conflict and see my improvement suggestion. :wink:

Ana06:
Question (unrelated to this PR but I am curious): @thack1 are you able to install packages like `npcap.vm` for which we use an AHK installer without a desktop session running? :thinking:

packages/common.vm/common.vm.nuspec
thack1: fixed

packages/common.vm/tools/vm.common/vm.common.psm1
thack1: fixed

thack1:
Hi @Ana06, thanks for your recommendations. I addressed them. Should be good to go.

thack1:
> Question (unrelated to this PR but I am curious): @thack1 are you able to install packages like `npcap.vm` for which we use an AHK installer without a desktop session running? 🤔 also how are you checking if the installation has completed?

I am currently only using selected packages. None of the AHK installers. I guess this would not work in my current setup

Ana06 (approved):
Thanks @thack1 :+1:

PR #1272, opened 2025-02-10
Add function VM-Get-Category to retrieve the category from the nuspec

sara-rn:
Retrieves the category from the nuspec file through the helper function `VM-Get-Category` 
Make lint.py compatible with the new nuspec format: add tags to the list of required tags and validate categories from the nuspec file 
Closes #1107

### DIFF
--- a/packages/common.vm/tools/vm.common/vm.common.psm1
+++ b/packages/common.vm/tools/vm.common/vm.common.psm1
@@ -1915,3 +1915,22 @@ function VM-Unzip-Recursively {
         }
     }
 }
+# Obtains the chocolatey install script path from a package, for example: C:\ProgramData\chocolatey\lib\capa.vm\tools\chocolateyinstall.ps1
+# Reads the tag with name tags from the nuspec file,
+# Returns the category if the tags exist in the nuspec
+function VM-Get-Category {
+    Param
+    (
+        [Parameter(Mandatory=$true)]
+        [string] $installPath
+
+    )
+    $chocoToolDir = $(Split-Path -parent (Split-Path -parent $installPath))
+    $packageName = ${Env:ChocolateyPackageName}
+    $nuspec = $packageName + ".nuspec"
+    VM-Assert-Path $chocoToolDir
+    $nuspecFilePath = Join-Path $chocoToolDir $nuspec -Resolve
+    $nuspecContent = [xml](Get-Content $nuspecFilePath)
+    $category = $nuspecContent.package.metadata.tags
+    return $category
+}
--- a/scripts/test/lint.py
+++ b/scripts/test/lint.py
@@ -369,7 +369,9 @@ class UsesInvalidCategory(Lint):
 INSTALL_LINTS = (
     MissesImportCommonVm(),
     FirstLineDoesNotSetErrorAction(),
-    UsesInvalidCategory(),
+    #This line has been disabled temporarily because it would validate the category from the chocolatey install script
+    #It needs to be disabled until a new linter checks if a valid category exists in the nuspec package
+    #UsesInvalidCategory(),
 )
 
 UNINSTALL_LINTS = (UsesInvalidCategory(),)
--- a/packages/common.vm/common.vm.nuspec
+++ b/packages/common.vm/common.vm.nuspec
@@ -2,7 +2,7 @@
 <package xmlns="http://schemas.microsoft.com/packaging/2015/06/nuspec.xsd">
   <metadata>
     <id>common.vm</id>
-    <version>0.0.0.20250204</version>
+    <version>0.0.0.20250206</version>
     <description>Common libraries for VM-packages</description>
     <authors>Mandiant</authors>
   </metadata>


packages/010editor.vm/tools/chocolateyinstall.ps1
```diff
@@ -3,7 +3,7 @@ Import-Module vm.common -Force -DisableNameChecking
 
 try {
   $toolName = '010Editor'
-  $category = 'Hex Editors'
+  $category = VM-Get-Category($MyInvocation.MyCommand.Definition)
```
Ana06, line 6: I think you can call `$MyInvocation.MyCommand.Definition` inside the function so that the function does not receive any  arguments:
```suggestion
  $category = VM-Get-Category()
```

scripts/test/lint.py
```diff
@@ -369,7 +370,7 @@ def check(self, path):
 INSTALL_LINTS = (
     MissesImportCommonVm(),
     FirstLineDoesNotSetErrorAction(),
-    UsesInvalidCategory(),
+    #UsesInvalidCategory(),
```
Ana06, line 373: instead of removing this linter you need to modify it to check for the line: `$category =VM-Get-Category()`. I am ok if you only add the function `VM-Get-Category` in this PR and in a separate PR you modify all packages and adapt the linter.

scripts/test/lint.py
```diff
@@ -94,6 +94,7 @@ class IncludesRequiredFieldsOnly(Lint):
         "description",
         "authors",
         "dependencies",
+        "tags",
```
Ana06, line 97: you need to add this together with the modification of all packages.

Ana06 (commented):
Nice first step to move the categories to the nuspec! :smile:

packages/common.vm/tools/vm.common/vm.common.psm1
```diff
@@ -1913,3 +1913,18 @@ function VM-Unzip-Recursively {
         }
     }
 }
+# Obtains the chocolatey install script path from a package, for example: C:\ProgramData\chocolatey\lib\capa.vm\tools\chocolateyinstall.ps1
+# Reads the tag with name tags from the nuspec file,
+# Returns the category if the tags exist in the nuspec
+function VM-Get-Category {
+    $installPath = $MyInvocation.MyCommand.Definition
+    $toolDir = "$(Split-Path -parent (Split-Path -parent $installPath))"
```
Ana06, line 1921: According to [our documentation](https://github.com/mandiant/VM-Packages/wiki/Coding-Conventions#tooldir) `$toolDir` should only be used for directory where the tool is actually installed. We need a different name here. I think you can also remove the `"` 
```suggestion
    $chocolateyToolDir = $(Split-Path -parent (Split-Path -parent $installPath))
```

packages/common.vm/tools/vm.common/vm.common.psm1
```diff
@@ -1913,3 +1913,18 @@ function VM-Unzip-Recursively {
...
+    $packageName = Split-Path -Path (Split-Path -parent $installPath) | Split-Path -Leaf
```
Ana06, line 1922: you can get the package name from the environment variable `ChocolateyPackageName`

packages/common.vm/tools/vm.common/vm.common.psm1
```diff
@@ -1913,3 +1913,18 @@ function VM-Unzip-Recursively {
...
+    $nuspec = $packageName + ".nuspec"
+    if (Test-Path ($toolDir)){
```
Ana06, line 1924: the directory must exist, otherwise we return an invalid category here

packages/010editor.vm/tools/chocolateyinstall.ps1
Ana06: I was wrong, it seems it fails when moving it.

packages/common.vm/tools/vm.common/vm.common.psm1
sara-rn: @Ana06 if $toolDir doesn't exist, the return value should be "" for example?

sara-rn: ah ok, so I remove the condition Test-Path

scripts/test/lint.py
Ana06, line 97: Please remove this change and add it after all packages have the category in the nuspec.

scripts/test/lint.py
Ana06, line 373: As discussed in our last meeting, Please add a comment to explain why this is needed. I would also move to a separate commit than the one adding the `VM-Get-Category`, as the changes are completely independent.

packages/010editor.vm/tools/chocolateyinstall.ps1
Ana06, line 6: Documenting discussion: It is a good idea to edit two packages (common and an extra one) in this PR to test `VM-Get-Category`. But it is a good idea to split the update of other packages in separate PR (for example 30 packages per PR) so that the CI takes less to run and we do not encounter any issues when pushing the packages to MyGet.

Ana06: Note the issue with 010Editor (just in case you notice slow local testing or CI): https://github.com/mandiant/VM-Packages/issues/1275

scripts/test/lint.py
Ana06: [nitpick] Still missing the comment here

packages/common.vm/tools/vm.common/vm.common.psm1
```diff
@@ -1913,3 +1913,23 @@ function VM-Unzip-Recursively {
...
+# Returns the category if the tags exist in the nuspec
+function VM-Get-Category {
+    Param
+    (
+        [Parameter(Mandatory=$true)]
+        [string] $installPath
+
+    )
+    $chocoToolDir = "$(Split-Path -parent (Split-Path -parent $installPath))"
+    #$packageName = Split-Path -Path (Split-Path -parent $installPath) | Split-Path -Leaf
```
Ana06, line 1927: Remove this old line.

Ana06 (commented):
[nipick] Remember to keep the commit subject short (50 chars is a common convention):
image

Also update the commit description, it mentions 7z which is not being updated.

scripts/test/lint.py
sara-rn: added comment now, I added it to a separate commit with the description so that's why I didn't add a comment

packages/010editor.vm/tools/chocolateyinstall.ps1
sara-rn: @Ana06 2 test packages have been removed from current PR

Ana06 (approved):

Ana06 (approved):

Issue #1275, opened 2025-02-12
[010editor.vm] Extremely Slow Download Speed for the 010Editor Installer


Ana06:
### What's the problem?

The download speed for 010 Editor installers from `https://download.sweetscape.com` is (at least some times) significantly slow. As reported by @aftwro in https://github.com/mandiant/flare-vm/issues/657, this impacts the installation of [FLARE-VM](https://github.com/mandiant/flare-vm/) as installing 010Editor (via the `010editor.vm` package which uses `https://download.sweetscape.com/010EditorWin64Installer15.0.1.exe`) dramatically increases installation time.

#### Using `curl` (Direct Download)

- Open a terminal or command prompt.
- Execute the following command:
  ```
  curl -L -v --noproxy '*' https://download.sweetscape.com/010EditorWin64Installer15.0.1.exe --output /dev/null
  ```
- Observe the download speed in the `curl` output, which will be significantly slower than desired.
  ![Image](https://github.com/user-attachments/assets/c6405e8c-7b9e-41aa-a30a-54d3f3f3899b)

#### Using Chocolatey
- Install [FLARE-VM](https://github.com/mandiant/flare-vm/) selecting `010editor.vm` as only installed package.
  ![Image](https://github.com/user-attachments/assets/2e801a4a-fcdd-4296-943c-7ece418b6cbe)

**Possible solutions:**

1.  Ask 010Editor if they are able to fix the download speed issue on `download.sweetscape.com` (preferred).
2.  Use an alternative download URL if someone is able to provide one.
3. If we are not able to fix the download issue we will have to remove 010Editor from [FLARE-VM default configuration](https://github.com/mandiant/flare-vm/blob/main/config.xml). 010 Editor is a very valuable tool for reverse engineering and malware analysis, so this is not desirable.

**I'll try to reach out to 010Editor to see if they are able to help.**

Ana06:
The issue still persists 😞

PR #1278, opened 2025-02-13
Modify packages with names d to e

sara-rn:
Second PR that modifies packages with names d to e
Adds the category to the nuspec file
Script to update packages https://gist.github.com/sara-rn/9b5d66e5b016e0eb36477ac180d50433
Removed dcode as it fails
Closes #1107

### DIFF
--- a/packages/de4dot-cex.vm/de4dot-cex.vm.nuspec
+++ b/packages/de4dot-cex.vm/de4dot-cex.vm.nuspec
@@ -2,11 +2,12 @@
 <package xmlns="http://schemas.microsoft.com/packaging/2015/06/nuspec.xsd">
   <metadata>
     <id>de4dot-cex.vm</id>
-    <version>4.0.0.20240411</version>
+    <version>4.0.0.20250218</version>
     <authors>ViRb3</authors>
     <description>A de4dot fork with full support for vanilla ConfuserEx</description>
     <dependencies>
-      <dependency id="common.vm" version="0.0.0.20240411" />
+      <dependency id="common.vm" version="0.0.0.20250206" />
     </dependencies>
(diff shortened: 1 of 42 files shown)


sara-rn:
It failes due to the dcode package, the URL returns 404, however I checked the URL and it seems to work

Ana06:
@sara-rn dcode is failing in our daily run too: https://github.com/mandiant/VM-Packages/wiki/Daily-Failures I recommend to remove the change in dcode from this PR so that we can merge the other packages and send a PR only for dcode. You will have the same issue for the other packages failing in the daily (dcode.vm, ida.plugin.sigmaker.vm, netcat.vm, and regcool.vm). We need to fix all the failing packages, as we need them to work in the CI to modify them and we need to modify them to move the category to the nuspec. I think some of them are easy to fix (likely just updating a hash). 

I am not sure what the issue with dcode is, you may need to do some debugging like testing the package locally and downloading it with CURL. It could be the URL has issues some times (similar to the issue with 010Editor: https://github.com/mandiant/VM-Packages/issues/1275). In that case, we would need to document the testing we have done in an issue and reach Digital Detective Group Ltd to see if they are able to help (as we did in https://github.com/mandiant/VM-Packages/issues/1275).

Ana06:
INFO: @sara-rn has implemented these changes using the script https://gist.github.com/sara-rn/9b5d66e5b016e0eb36477ac180d50433 We are splitting the moving of the category to the nuspec in several PRs to avoid issues with the CI (as recommended in the [_Continuous Integration (CI) Testing_ Wiki page](https://github.com/mandiant/VM-Packages/wiki/Testing#continuous-integration-ci-testing)).

Ana06 (approved):

Issue #1267, opened 2025-02-04
Rename Productivity Tools category


Ana06:
### Details

I propose to rename `Productivity Tools` to `Productivity` as we do not include `Tools` in other categories (we could use for example `Python Tools`) and I think it is clear what is meant. @mandiant/vms opinions?

PR #1282, opened 2025-02-13
Update categories

sara-rn:
Remove special character & from the category name because it is now part of the XML tags and it cannot contain such character Add new category IDA Plugins for the nuspec

### DIFF
--- a/categories.txt
+++ b/categories.txt
@@ -1,4 +1,4 @@
-Command & Control
+Command and Control
 Credential Access
 Debuggers
 Delphi
@@ -9,8 +9,9 @@ Exploitation
 File Information
 Forensic
 Hex Editors
+IDA Plugins
 InnoSetup
-Java & Android
+Java and Android
 Javascript
 Lateral Movement
 Memory
--- a/scripts/test/lint.py
+++ b/scripts/test/lint.py
@@ -374,7 +374,8 @@ INSTALL_LINTS = (
     #UsesInvalidCategory(),
 )
 
-UNINSTALL_LINTS = (UsesInvalidCategory(),)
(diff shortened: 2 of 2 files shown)


categories.txt
```diff
@@ -9,8 +9,9 @@ Exploitation
 File Information
 Forensic
 Hex Editors
+IDA Plugins
 InnoSetup
-Java & Android
+Java and Android
```
Ana06, line 14: [nitpick] is the following possible?
```suggestion
Java/Android
```

@mandiant/flare-vm any preference between the two?

sara-rn: Java/Android is possible as well, let me know which one is preferred

Ana06: I prefer `Java/Android` that is a bit shorter, but I have no strong opinion. So up to you if nobody else has an opinion. :wink:

Ana06:
@sara-rn what's your opinion about https://github.com/mandiant/VM-Packages/issues/1267? If you agree, we could add it to this PR.

categories.txt
sara-rn: then and, cause it is not possible to create a folder with / in windows

Ana06:
you need to rebase for the linter to succeed.

scripts/test/lint.py
```diff
@@ -374,7 +374,8 @@ def check(self, path):
     #UsesInvalidCategory(),
 )
 
-UNINSTALL_LINTS = (UsesInvalidCategory(),)
+#UNINSTALL_LINTS = (UsesInvalidCategory(),)
```
Ana06, line 377: [nitpick] A comment here would also be good

Ana06 (approved):

PR #1287, opened 2025-02-14
Make linter compatible with new nuspec packages

sara-rn:
Ensures the category in the nuspec is a valid category
The exceptions are only the packages that we do not display to the user: common.vm, installer.vm, debloat.vm. 
The install script must obtain the category from the function VM-Get-Category
This PR needs to be approved after all the packages have the new category in the nuspec package
Resolves (partially) #1107

### DIFF
--- a/scripts/test/lint.py
+++ b/scripts/test/lint.py
@@ -273,6 +273,40 @@ class PackageIdNotMatchingFolderOrNuspecName(Lint):
         folder = path.parts[-2]
 
         return not (pkg_id == folder == nuspec[:-len(".nuspec")])
+        
+        
+        
+class UsesInvalidCategory(Lint):
+    # The common.vm, debloat.vm, and installer.vm packages are special as they
+    # assist with the installation and allow to share code between packages.
+    # They do not install any tool and consequently don't have a category.
+    EXCLUSIONS = [
+        "common.vm",
+        "debloat.vm",
+        "installer.vm",
+        # exclude dcode as it fails to install: https://github.com/mandiant/VM-Packages/issues/1176
+        "dcode.vm",
+    ]
+    root_path = os.path.abspath(os.path.join(__file__, "../../.."))
+    categories_txt = os.path.join(root_path, "categories.txt")
+    with open(categories_txt) as file:
+        CATEGORIES = [line.rstrip() for line in file]
+        logger.debug(CATEGORIES)
+
+    name = "Uses an invalid category"
+    recommendation = f"Place a category between <tags> and </tags> from {categories_txt} or exclude the package in the linter"
+
+    def check(self, path):
+        if any([exclusion in str(path) for exclusion in self.EXCLUSIONS]):
+            return False
+
+        # utf-8-sig ignores BOM
+        file_content = open(path, "r", encoding="utf-8-sig").read()
+
+        match = re.search(r"<tags>(?P<category>[\w ]+)<\/tags>", file_content)
+        if not match or match.group("category") not in self.CATEGORIES:
+            return True
+        return False
 
 NUSPEC_LINTS = (
     IncludesRequiredFieldsOnly(),
@@ -281,6 +315,7 @@ NUSPEC_LINTS = (
     DependencyContainsUppercaseChar(),
     VersionNotUpdated(),
     PackageIdNotMatchingFolderOrNuspecName(),
+    UsesInvalidCategory(),
 )
 
 
@@ -321,7 +356,7 @@ class FirstLineDoesNotSetErrorAction(Lint):
         return not self.FIRST_LINE == lines[0]
 
 
-class UsesInvalidCategory(Lint):
+class UsesCategoryFromNuspec(Lint):
     # Some packages don't have a category (we don't create a link in the tools directory)
     EXCLUSIONS = [
         ".dbgchild.vm",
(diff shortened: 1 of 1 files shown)


scripts/test/lint.py
```diff
@@ -272,6 +273,36 @@ def check(self, path):
         folder = path.parts[-2]
 
         return not (pkg_id == folder == nuspec[:-len(".nuspec")])
+        
+        
+        
+class UsesInvalidCategoryNuspec(Lint):
```
Ana06, line 279: I think the linter that checks the category is one in `categories.txt` should be named `UsesInvalidCategory` and the one that checks that the installer/uninstaller use the category from the nuspec be named something like `UsesCategoryFromNuspec`

scripts/test/lint.py
```diff
@@ -272,6 +273,36 @@ def check(self, path):
...
+        "installer.vm",
+    ]
+    root_path = os.path.abspath(os.path.join(__file__, "../../.."))
+    categories_txt = os.path.join(root_path, "categories.txt")
+    with open(categories_txt) as file:
+        CATEGORIES = [line.rstrip() for line in file]
+        logger.debug(CATEGORIES)
+
+    name = "Uses an invalid category"
+    recommendation = f"Set $category to a category in {categories_txt} or exclude the package in the linter"
```
Ana06, line 293: this recommendation is copied from the old linter and need to be adapted to explain to use `tag` in the nuspec.

scripts/test/lint.py
```diff
@@ -344,24 +376,20 @@ class UsesInvalidCategory(Lint):
...
-    categories_txt = os.path.join(root_path, "categories.txt")
-    with open(categories_txt) as file:
-        CATEGORIES = [line.rstrip() for line in file]
-        logger.debug(CATEGORIES)
+    
 
-    name = "Uses an invalid category"
-    recommendation = f"Set $category to a category in {categories_txt} or exclude the package in the linter"
+    name = "Doesn't use the function VM-Get-Category"
+    recommendation = f"Set $category to retrieve the category from VM-Get-Category or exclude the package in the linter"
```
Ana06, line 382: I think it would be useful to be more explicit about how to set `$category`
```suggestion
    recommendation = f"Set '$category = VM-Get-Category($MyInvocation.MyCommand.Definition)' or exclude the package in the linter"
```

scripts/test/lint.py
```diff
@@ -344,24 +376,20 @@ class UsesInvalidCategory(Lint):
...
         if any([exclusion in str(path) for exclusion in self.EXCLUSIONS]):
             return False
-
+        
         # utf-8-sig ignores BOM
         file_content = open(path, "r", encoding="utf-8-sig").read()
 
-        match = re.search(r"\$category = ['\"](?P<category>[\w &/]+)['\"]", file_content)
-        if not match or match.group("category") not in self.CATEGORIES:
+        match = re.search(r"\$category = VM-Get-Category\s*\(\$MyInvocation\.MyCommand\.Definition\)", file_content)
```
Ana06, line 391: I think we do not need to support extra spaces (for consistency and better code readability) and that we can make the string we want to match more readable by using `re.escape`:
 ````suggestion
        pattern = re.escape("$category = VM-Get-Category($MyInvocation.MyCommand.Definition)")
        match = re.search(pattern, file_content)
```

**Review by Ana06 — COMMENTED** [review]
This is going in the right direction! :+1: 

This PR needs to undo the current linter exclusions, which appear not to be present (likely missing a rebase).

**Review by Ana06 — COMMENTED** [review]
@sara-rn please rebase when you get the change to address the feedback so that we can see how many packages still need to be adapted.

**Review by Ana06 — COMMENTED** [review]
There seem to be a bug in the linter code:
```
TypeError: 'UsesCategoryFromNuspec' object is not iterable
```

**Comment by sara-rn** [comment]
@Ana06 the error is fixed now, I introduced when enabled the uninstall check

File: scripts/test/lint.py [hunk 5]
```diff
@@ -354,6 +356,7 @@ def check(self, path):
 
 class UsesCategoryFromNuspec(Lint):
     # Some packages don't have a category (we don't create a link in the tools directory)
+    # we exclude dcode since it fails to install, issue tracked in #1175
```
Ana06, line 359: [nitpick] I think the comment should be in the same or previous line as the dcode exclusion. Otherwise I think it is likely that we forget to remove the comment when we remove the dcode exclusion
```suggestion
```

scripts/test/lint.py
```diff
@@ -374,6 +377,7 @@ class UsesCategoryFromNuspec(Lint):
         "x64dbgpy.vm",
         "vscode.extension.",
         "chrome.extensions.vm",
+        "dcode.vm",
```
Ana06, line 380: [nitpick]  I think the comment should be in the same or previous line as the dcode exclusion. Otherwise I think it is likely that we forget to remove the comment when we remove the dcode exclusion. Please use the URL instead of the issue number so that it is clear what issue is meant and how to access it. 
```suggestion
        # exclude dcode as it fails to install: https://github.com/mandiant/VM-Packages/issues/1176
        "dcode.vm",
```

scripts/test/lint.py
```diff
@@ -278,10 +278,12 @@ def check(self, path):
         
 class UsesInvalidCategory(Lint):
     # Some nuspec packages don't have a category
+    # excluded dcode because it fails to be installed
     EXCLUSIONS = [
         "common.vm",
         "debloat.vm",
         "installer.vm",
+        "dcode.vm",
```
Ana06, line 286: [nitpick]  I think the comment should be in the same or previous line as the dcode exclusion. Please also add the issue URL and keep it consistent with the other exclusion.
```suggestion
        # exclude dcode as it fails to install: https://github.com/mandiant/VM-Packages/issues/1176
        "dcode.vm",
```

scripts/test/lint.py
```diff
@@ -278,10 +278,12 @@ def check(self, path):
         
 class UsesInvalidCategory(Lint):
     # Some nuspec packages don't have a category
+    # excluded dcode because it fails to be installed
```
Ana06, line 281: [nitpick]  I think the comment should be in the same or previous line as the dcode exclusion. Please also add the issue URL and keep it consistent with the other exclusion.
```suggestion
```

scripts/test/lint.py
```diff
@@ -273,6 +273,38 @@ def check(self, path):
         folder = path.parts[-2]
 
         return not (pkg_id == folder == nuspec[:-len(".nuspec")])
+        
+        
+        
+class UsesInvalidCategory(Lint):
+    # Some nuspec packages don't have a category
```
Ana06, line 280: I think some more details about why these packages don't have a category would be useful:
```suggestion
    # The common.vm, debloat.vm, and installer.vm packages are special as they
    # assist with the installation and allow to share code between packages.
    # They do not install any tool and consequently don't have a category.
```

Ana06 (changes requested):

scripts/test/lint.py
Ana06: @sara-rn you have marked some feedback as resolved that still applies, have you maybe forgotten to push the changes it?

scripts/test/lint.py
```diff
@@ -273,6 +273,41 @@ def check(self, path):
...
 
         return not (pkg_id == folder == nuspec[:-len(".nuspec")])
+        
+        
+        
+class UsesInvalidCategory(Lint):
+    # The common.vm, debloat.vm, and installer.vm packages are special as they
+    # assist with the installation and allow to share code between packages.
+    # They do not install any tool and consequently don't have a category.
+    # excluded dcode because it fails to be installed
```
Ana06, line 283: you forgot to remove this line (that you ave moved to just before the dcode exclusion)
```suggestion
```

Ana06 (commented):
Please rebase so that the linter takes into account the deletion of the two failing packages :wink:

Ana06 (approved):
LGTM, thanks for all the work @sara-rn! 🎉

Issue #1171, opened 2024-11-22
chainsaw is missing a dependency


Ana06:
### Details

After installing `chainsaw.vm` in [a minimal FLARE-VM installation](https://github.com/mandiant/VM-Packages/wiki/Testing#minimal-flare-vm-installation), running chainsaw fails with the following error indicating it is missing a dependency:
![image](https://github.com/user-attachments/assets/3cb77ba5-9bfe-4cb2-9d04-d9292e8d4776)

Ana06:
@naacbin as you created this package, would you like to add the dependency? :smile: :pray:

PR #1288, opened 2025-02-15
Add vcredist140 dependency

naacbin:
Solve issue #1171


Ana06 (approved):
Thanks @naacbin! :bouquet:

PR #1289, opened 2025-02-15
Update ZimmermanTools

naacbin:
Add more green to the CI :smile:


Ana06 (approved):
Thanks for the hash fixes @naacbin! :rocket:

Could we use the new `-verifySignature` argument introduced in https://github.com/mandiant/VM-Packages/pull/1199 for these packages too so that they don't break in the future?

PR #1280, opened 2025-02-13
Modify packages from k to o

sara-rn:
Add the category to the nuspec file
Script to update packages https://gist.github.com/sara-rn/9b5d66e5b016e0eb36477ac180d50433
Closes https://github.com/mandiant/VM-Packages/issues/1107

PR #1294, opened 2025-02-19
Modify uninstall script in packages i to m

sara-rn:
Third PR to update packages in chunks, according to https://github.com/mandiant/VM-Packages/wiki/Testing
The uninstall script retrieves the category from the nuspec package
Script used to update categories https://gist.github.com/sara-rn/3e989f497a651eeab8ccbb2ef93b0a51
Closes #1107


Ana06 (changes requested):
Please **remove `kernel-ost-viewer.vm` from this PR** and create an independent PR for it to avoid that it causes issues as it happened in https://github.com/mandiant/VM-Packages/pull/1280#issuecomment-2668393549

Ana06 (approved):

PR #1303, opened 2025-02-20
Update category in the capesolo package

sara-rn:
Update the nuspec package and the install script to retrieve the category from the nuspec


Ana06 (approved):
Thanks @sara-rn!

PR #1304, opened 2025-02-20
Modify uninstall script of kernel-ost-viewer

sara-rn:
The package failed during the update of the nuspec package. 
It retrieves the category in the uninstall script from the nuspec.


Ana06 (approved):
So it seems it failed because a punctual server issues, as it works now. Thanks for extracting the package to prevent issues @sara-rn, let's see if it is correctly published to MyGet.

PR #1306, opened 2025-02-20
Update regcool nuspec package

sara-rn:
Updated the category in the nuspec package
The regcool package fails, the sha256 differs and has been updated


Ana06 (approved):
Thanks for fixing the hash @sara-rn

PR #1308, opened 2025-02-24
Modify package inkovethehash to include the category

sara-rn:
Add the category to the nuspec package, both the install and uninstall scripts retrieve the category from the nuspec


Ana06 (approved):
Thanks @sara-rn! :bouquet:

PR #1309, opened 2025-02-24
Delete netcat package

sara-rn:
Netcat package failed
Add category in the nuspec package


sara-rn:
@mandiant/flare-vm it seems to be an issue with this package, `choco install netcat` returns an error, see below. I wrote to the maintainers of netcat in the chocolatey community, we might need to change the install script, however it is not straightforward to find a reliable URL. As netcat is quite an old tool, perhaps it can be replaced with newer tools @mandiant/commando-vm any thoughts just in case we need to remove this tool?
```
 - netcat - netcat not installed. The package was not found with the source(s) listed
```

Ana06:
@sara-rn who/where exactly have you contacted? It is common to write a comment in the chocolatey package page to report issues with the package, but I do not see anything new there: https://community.chocolatey.org/packages/netcat The package has been unlisted, so it is likely a good idea to stop using it and remove it from the repository. I think @mandiant/commando-vm does not uses this package. FLARE-VM had it in the default configuration till December, when we removed it because the community package had been unlisted and nmap.vm already installs ncat which provides similar functionality: https://github.com/mandiant/flare-vm/commit/1eebdbc01868770c254b28c9118a908f52fc455c

@sara-rn unless you get information from netcat chocolatey maintainers that they are planning to continue to maintain the package, I think you can go ahead and update this PR to remove the package. Please create also an issue explaining summarizing the discussion here, as in addition from deleting the package from this repository, we need to also remove it from the MyGet feed.

sara-rn:
tracked in #1309

Ana06 (changes requested):
Please remove/squash the unrelated commits so that we have a single commit to delete the package.

Ana06 (commented):
I think adding the reason why we delete the package to the commit message would be useful when checking the commit history. The reason to delete it is not that it fails, but that it fails and because the reasons mentioned in https://github.com/mandiant/VM-Packages/issues/1312#issuecomment-2681239070 we can't fix it.

Ana06 (approved):

Issue #1316, opened 2025-02-27
Daily Test Failure - arsenalimagemounter.vm


PrajeetGuha:
### What's the problem?

**arsenalimagemounter.vm** daily test failure

![Image](https://github.com/user-attachments/assets/0fad9412-6819-4440-8d6b-8979240aa401)

### Detected by test suite

Yes

PrajeetGuha:
I had a wrong understanding of the package. It is auto-updated by update_package job. Package update was not done https://github.com/mandiant/VM-Packages/pull/1301 that is why the issue occurred possibly.

PR #1317, opened 2025-02-27
Changes to arsenalimagemounter package hash and url

PrajeetGuha:
changes related to issue https://github.com/mandiant/VM-Packages/issues/1316


PrajeetGuha:
I had a wrong understanding of the package. It is auto-updated by update_package job. Package update was not done https://github.com/mandiant/VM-Packages/pull/1301 that is why the issue occurred possibly. No need to merge this.

PR #1318, opened 2025-02-28
Syncronize categories.txt and categories from the templates

sara-rn:
Delete the Powershell category as there is no package with this category.
Update the categories in the ISSUE templates so it is syncronized with `categories.txt`


Ana06 (approved):
Thanks @sara-rn

PR #1319, opened 2025-02-28
Update description of metapackage vcbuildtools.vm

sara-rn:
The description contained an unordered list with -, that caused a formatting issue to display the Wiki categories


Ana06 (approved):
LGTM.

It seems there is an issue with the indentation if `</description>` which could be corrected here as well.

Ana06 (approved):

PR #1320, opened 2025-02-28
Update wiki that lists packages by category

sara-rn:
Update the script that generates the wiki with the list of packages by category. The new nuspec packages now contain the category even if it is not used to prevent packages from being uncategorized.

Resolves #1107

### DIFF
--- a/.github/workflows/generate_package_wiki.yml
+++ b/.github/workflows/generate_package_wiki.yml
@@ -24,7 +24,7 @@ jobs:
 
       - name: Generate Wiki Content
         id: generate-content
-        run: scripts/utils/generate_package_wiki.ps1
+        run: python scripts/utils/generate_package_wiki.py
 
       - name: Commit changes
         working-directory: wiki
--- a/scripts/utils/generate_package_wiki.py
+++ b/scripts/utils/generate_package_wiki.py
@@ -1,0 +1,88 @@
+import argparse
+import os
+import pathlib
+import xml.etree.ElementTree as ET
+from collections import defaultdict
+
+# Dict[str (category), list (packages information)]
+packages_by_category = defaultdict(str)
+
+
+def sort_write_wiki_content(file_path):
+    """ Writes package information sorted by category to a Markdown wiki file.
+
+    This function iterates through the `packages_by_category` dictionary, which
+    contains package information organized by category. For each category, it
+    generates a Markdown header and a table containing package names and
+    descriptions. The resulting Markdown content is then written to the specified
+    file.
+
+    Args:
+        file_path (str): The path to the output Markdown file.
+    """
+    wikiContent = ""
+    for category in packages_by_category.keys():
+        wikiContent += "## " + category + "\n\n"
+        wikiContent += "| Package | Description |\n"
+        wikiContent += "| ------- | ----------- |\n"
+        wikiContent += packages_by_category[category] + "\n\n"
+    with open(file_path, "w", encoding="utf-8") as f:
+        f.write(wikiContent)
+
+
+def find_element_text(parent, tag_name):
+    """Retrieves the text content of a specified XML element.
+
+    This function searches for a child element with the given tag name within the
+    provided parent element. It handles XML namespaces by using a wildcard.
+
+    Args:
+        parent: The parent XML element.
+        tag_name: The name of the child element to find.
+
+    Returns:
+        The text content of the found element, or None if the element is not found.
+    """
+    tag = parent.find(f"{{*}}{tag_name}")
+    if tag is None:
+        return None
+    return tag.text
+
+
+def process_packages_directory(packages_dir):
+    """ Obtains the package name, description and category from a directory
+    
+    This function parses all the nuspec files in the specified packages directory.
+    It saves into packages_by_category[category] a line containing the package
+    name and the description separated by "|".
+    
+    Args:
+        packages: directory where the packages reside.
+    """
+    if not os.path.isdir(packages_dir):
+        raise FileNotFoundError(f"Packages directory not found: {packages_dir}")
+
+    for nuspec_path in pathlib.Path(packages_dir).glob("**/*.nuspec"):
+        nuspec_tree = ET.parse(nuspec_path)
+        nuspec_metadata = nuspec_tree.find("{*}metadata")
+        if nuspec_metadata is not None:
+            category = find_element_text(nuspec_metadata, "tags")
+            # We are only interested in the packages with a category.
+            # Some packages that assist with the istallation (such as common.vm) do
+            # not contain a category
+            if not category:
+                continue
+            description = find_element_text(nuspec_metadata, "description")
+            package = find_element_text(nuspec_metadata, "id")
+            """packages_by_category is a dictionary of str where each package and description
+            are appended formatted in Markdown"""
+            packages_by_category[category] += "| " + package + " | " + description + " |\n"
+
+
+if __name__ == "__main__":
+    parser = argparse.ArgumentParser(description="Create wiki of packages.")
+    parser.add_argument("--packages", type=str, required=False, default="packages", help="Path to packages")
+    parser.add_argument("--wiki", type=str, required=False, default="wiki/Packages.md", help="Path to the wiki")
+    args = parser.parse_args()
+    process_packages_directory(args.packages)
+    sort_write_wiki_content(args.wiki)


sara-rn:
There is a formatting issue due to the description of the metapackage `vcbuildtools.vm` addressed in #1319

scripts/utils/generate_package_wiki.ps1
```diff
@@ -4,29 +4,23 @@ $wikiContent = ""
...
+            if (-not $category){
+                $category = "Not Categorized"
+            } else {
+                if (-not ($packagesByCategory.ContainsKey($category))) {
+                    $packagesByCategory[$category] = ""
+                }
+            }
         } else {
             $description = "Nuspec file not found."
+            $category = "Nuspec file not found."
```
Ana06, line 23: What is this code doing? If the nuspec file is not found, something strange has happened and I think the script should fail.

Ana06 (changes requested):
@sara-rn the current script contains code to handle non categorized packages and handle errors that can't happen now that we have moved the category to the nuspec. This should be removed to simplify the code making it easier to understand. @sara-rn note that @emtuls first version of the script was a proof of concept and there may be issues with it. Finishing this GH action is part of https://github.com/mandiant/VM-Packages/issues/1107

I recomend you rewrite the script to generate the wiki in Python as the rest of our scripts, allowing us to keep them consistent and to reuse code. The code would be IMO also simpler and easier to modify and review. See for example https://github.com/mandiant/VM-Packages/blob/main/scripts/utils/generate_daily_results.py for another script that creates the wiki page. The linter is a good example to parse nuspec files.

scripts/utils/generate_package_wiki.ps1
```diff
@@ -4,29 +4,23 @@ $wikiContent = ""
...
-    }
 
     try {
         $nuspecFile = "$packagesDir\$package\$package.nuspec"
         if (Test-Path $nuspecFile) {
             $xml = [xml](Get-Content $nuspecFile)
             $description = $xml.package.metadata.description
+            $category = $xml.package.metadata.tags
+            if (-not $category){
+                $category = "Not Categorized"
```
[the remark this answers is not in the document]
sara-rn: understood, I simply adapted the existent code, I will write a new Python script from the scratch

scripts/utils/generate_package_wiki.ps1
sara-rn: this was existent code that printed that in case the nuspec file was not found or able to parse

sara-rn:
added the python script in a separate commit just in case, I haven't deleted the Powershell script yet
example of the wiki: https://github.com/sara-rn/VM-Packages/wiki/Packages

scripts/utils/generate_package_wiki.py
```diff
@@ -0,0 +1,62 @@
...
+        package_path = os.path.join(packages_dir, package_name)
+        nuspec_file = os.path.join(package_path, f"{package_name}.nuspec")
+        try:
+            with open(nuspec_file, 'r', encoding='utf-8') as f:
+                content = f.read()
+            category_match = re.search(r'<tags>([^<]*)</tags>', content)
+            if category_match:
+                category = category_match.group(1)
+                description_match = re.search (r'<description>([^<]*)</description>', content)
+                description = description_match.group(1)
```
Ana06, line 45: I think it is more readable and a more robust solution to parse the XML using Python builtin `xml.etree.ElementTree` than doing it regular expressions as you need to access several fields. This will also allow us to add more fields to the Wiki page in the future.
```suggestion
            nuspec_tree = ET.parse(nuspec_path)
            nuspec_metadata = nuspec_tree.find("{*}metadata")
            package = nuspec_metadata.find("{*}id").text
            category = nuspec_metadata.find("{*}tags").text
            description = nuspec_metadata.find("{*}description").text
```

To avoid code duplication this code can be simplified with the following helper function:
```Python
get_text(element, field):
    return element.find("{*}field").text
````

File: scripts/utils/generate_package_wiki.py [hunk 4]
```diff
@@ -0,0 +1,62 @@
...
+                if not category in packagesByCategory:
+            	    packagesByCategory[category]=""
```
Ana06, line 47: Use [defaultdict](https://docs.python.org/2/library/collections.html#collections.defaultdict) to initialize this variable to remove this `if`

```suggestion
```

scripts/utils/generate_package_wiki.py
```diff
@@ -0,0 +1,62 @@
+import os
+import sys
+import re
+import datetime
+import argparse
+
+
+packagesByCategory={}
```
Ana06, line 8: Follow [variables naming convention in Google Python Style Guide](https://google.github.io/styleguide/pyguide.html#316-naming)
```suggestion
packages_by_category = {}
```

scripts/utils/generate_package_wiki.py
```diff
@@ -0,0 +1,62 @@
...
+                content = f.read()
+            category_match = re.search(r'<tags>([^<]*)</tags>', content)
+            if category_match:
+                category = category_match.group(1)
+                description_match = re.search (r'<description>([^<]*)</description>', content)
+                description = description_match.group(1)
+                if not category in packagesByCategory:
+            	    packagesByCategory[category]=""
+                packagesByCategory[category]+="|" + package + "|" + description + "|\n"
+                print (f"| {package} | {description} |")
```
Ana06, line 49: what is the purpose of this `print`?

scripts/utils/generate_package_wiki.py
```diff
@@ -0,0 +1,62 @@
...
+
+	    
+if __name__ == "__main__":
+
+
+    parser = argparse.ArgumentParser(description="Create wiki of packages.")
+    parser.add_argument("packages", type=str, help="Path to packages")
+    args = parser.parse_args()
+    process_packages_directory(args.packages)
+    write_wiki_content("wiki/Packages.md")
```
Ana06, line 62: I find it inconsistent to have an argument for the path for the packages directory and not one for the wiki page path. I suggest to have either both as arguments or have both as global constant variables. The only reason to provide them as argument is to allow to call the script from a different folder locally, in the action they will always be the same. If you use them as argument, please provide a default value, as in the action they will be always the same, so that we do not need to provide them there.

scripts/utils/generate_package_wiki.py
```diff
@@ -0,0 +1,62 @@
...
+        with open(file_path, 'w', encoding='utf-8') as f:
+           f.write(wikiContent)
+    except FileNotFoundError:
+        print(f"Error: File '{file_path}' not found.")
+
+  
+
+
+def process_packages_directory(packages_dir):
+    """Processes all.nuspec files in the 'packages' directory structure."""
```
Ana06, line 28: This description is not clear to me and has a typo. Please add more details.

scripts/utils/generate_package_wiki.py
```diff
@@ -0,0 +1,62 @@
...
+import sys
+import re
+import datetime
+import argparse
+
+
+packagesByCategory={}
+
+
+def write_wiki_content(file_path):
```
Ana06, line 11: Please add a description to this function.

scripts/utils/generate_package_wiki.py
```diff
@@ -0,0 +1,62 @@
...
+    wikiContent = ""
+    for category in packagesByCategory.keys():
+        wikiContent += "## " + category + "\n\n"
+        wikiContent += "| Package | Description |\n"
+        wikiContent += "|---|---|\n"
+        wikiContent += packagesByCategory[category] + "\n\n"
```
Ana06, line 17: I find it confusing that the table format is defined in two different functions. This is an issue when adding new items to the table. :thinking:

Ana06 (changes requested):
can you please run the following Python linters/formatters so that we do not have to discuss format/style in the PR review:
```
black --line-length=120
flake8 --max-line-length=150
isort --profile black --line-length=120
```

I plan to enable these linters in our CI as I have done in FLARE-VM: https://github.com/mandiant/flare-vm/blob/main/.github/workflows/linter.yml I was waiting for https://github.com/mandiant/VM-Packages/pull/1261, but we may want to enable it already if that PR gets long to complete.

You can remove the PS script, the Python script goes in the right direction.

scripts/utils/generate_package_wiki.py
sara-rn: yes,  that is because each value in the dictionary contains the package name and the description already formatted in markdown previously. It is not possible to add the markdown headers then as this can only be done when all the packages have been processed. It is done in the `write_wiki_content` also because dictionaries are sorted alphabetically by default.

Ana06: yes, but the current implementation makes it difficult to understand it. This can be enhanced for example in one of the following ways:
- `packagesByCategory` contains a hash or tuple so that you keep the logic from parsing the packages and the logic from creating markdown in two separate functions
- With the other code suggestions, the resulting code may be simpler enough that we can have all the logic in the same function.

In any case, documentation in the code (as function descriptions and comments) is needed.

scripts/utils/generate_package_wiki.py
```diff
@@ -0,0 +1,62 @@
...
+    try:   
+        with open(file_path, 'w', encoding='utf-8') as f:
+           f.write(wikiContent)
+    except FileNotFoundError:
```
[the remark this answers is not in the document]
Ana06: We want the action to fail if something unexpected happens. So you should only catch errors that you can handle. In case you are not able to write the wiki page, the action should fail because you are capturing the error.  So yes, remove the try-catch.

scripts/utils/generate_package_wiki.py
sara-rn: the line  `category = nuspec_metadata.find("{*}tags").text` returns an error: 
  ```
File "/home/sararn/code/VM-Packages/scripts/utils/generate_package_wiki.py", line 60, in process_packages_directory
    category = nuspec_metadata.find("{*}tags").text
               ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
AttributeError: 'NoneType' object has no attribute 'text'
```

Comment (sara-rn) [reply scripts/utils/generate_package_wiki.py:45]: extracting the other 2 fields work. in my code I already extracted the package_name so I didn't need to find the `id` node

Comment (sara-rn) [reply scripts/utils/generate_package_wiki.py:45]: some packages such as `common.vm` do not contain `tags` therefor a validation is required otherwise it fails

**Comment by sara-rn** [comment]
@Ana06 wiki in my local repo https://github.com/sara-rn/VM-Packages/wiki/Packages

File: scripts/utils/generate_package_wiki.py [hunk 12]
```diff
@@ -0,0 +1,70 @@
...
+
+packagesByCategory = defaultdict(str)
+
+
+def sort_write_wiki_content(file_path):
+    """
+    iterates over each category from the dictionary packagesByCategory and
+    writes the Markdown formatted string that contains
+    - header with the category name
+    - table with two columns: Package and Description of each of the packages"""
```
Ana06, line 16: Use Google recommended style to document Python functions in this and the other functions: https://github.com/google/styleguide/blob/gh-pages/pyguide.md#383-functions-and-methods

scripts/utils/generate_package_wiki.py
```diff
@@ -0,0 +1,70 @@
...
+    for nuspec_path in pathlib.Path(packages_dir).glob("**/*.nuspec"):
+
+        try:
+            nuspec_tree = ET.parse(nuspec_path)
+            nuspec_metadata = find_element(nuspec_tree, "metadata")
+            if nuspec_metadata is not None:
+                description = find_element(nuspec_metadata, "description").text
+                package = find_element(nuspec_metadata, "id").text
+                category_element = find_element(nuspec_metadata, "tags")
+                # several packages such as common.vm do not contain a category
```
Ana06, line 49: [nitpick] Add some more details
```suggestion
                # We are only interested in the packages with a category
                # Some packages that assist with the installation (such as common.vm) do not contain a category
```

scripts/utils/generate_package_wiki.py
```diff
@@ -0,0 +1,70 @@
...
+def process_packages_directory(packages_dir):
+    """Processes all the nuspec files in the 'packages' directory structure.
+    Once a category is found it adds the following line to the dictionary:
+    packagesByCategory where the key is the category:"""
+    if not os.path.isdir(packages_dir):
+        print(f"Error: Directory '{packages_dir}' not found.")
+        sys.exit()
+
+    for nuspec_path in pathlib.Path(packages_dir).glob("**/*.nuspec"):
+
```
Ana06, line 41: [nitpick]
```suggestion
```

scripts/utils/generate_package_wiki.py
```diff
@@ -0,0 +1,70 @@
...
+# Helper function to find elements
+def find_element(parent, tag_name):
+    return parent.find(f"{{*}}{tag_name}")
+
+
+def process_packages_directory(packages_dir):
+    """Processes all the nuspec files in the 'packages' directory structure.
+    Once a category is found it adds the following line to the dictionary:
+    packagesByCategory where the key is the category:"""
```
Ana06, line 35: This description is confusing to me.

scripts/utils/generate_package_wiki.py
```diff
@@ -0,0 +1,70 @@
...
+        wikiContent += "| Package | Description |\n"
+        wikiContent += "|---|---|\n"
+        wikiContent += packagesByCategory[category] + "\n\n"
+    with open(file_path, "w", encoding="utf-8") as f:
+        f.write(wikiContent)
+
+
+# Helper function to find elements
+def find_element(parent, tag_name):
+    return parent.find(f"{{*}}{tag_name}")
```
Ana06, line 29: Extend this function to avoid having to repeat the `.text` when calling this function:
```suggestion
def find_element(parent, tag_name):
    tag = parent.find(f"{{*}}{tag_name}")
    if not tag:
        return None
    return tag.text
```

scripts/utils/generate_package_wiki.py
```diff
@@ -0,0 +1,70 @@
...
+        sys.exit()
+
+    for nuspec_path in pathlib.Path(packages_dir).glob("**/*.nuspec"):
+
+        try:
+            nuspec_tree = ET.parse(nuspec_path)
+            nuspec_metadata = find_element(nuspec_tree, "metadata")
+            if nuspec_metadata is not None:
+                description = find_element(nuspec_metadata, "description").text
+                package = find_element(nuspec_metadata, "id").text
```
Ana06, line 47: place this after accessing the category and checking it is not None as otherwise you do not use it.

scripts/utils/generate_package_wiki.py
```diff
@@ -0,0 +1,70 @@
...
+                category_element = find_element(nuspec_metadata, "tags")
+                # several packages such as common.vm do not contain a category
+                if category_element is None:
+                    continue
+                else:
```
Ana06, line 52:
```suggestion
                if category_element:
                    continue
```

scripts/utils/generate_package_wiki.py
```diff
@@ -0,0 +1,62 @@
...
+            category_match = re.search(r'<tags>([^<]*)</tags>', content)
+            if category_match:
+                category = category_match.group(1)
+                description_match = re.search (r'<description>([^<]*)</description>', content)
+                description = description_match.group(1)
+                if not category in packagesByCategory:
+            	    packagesByCategory[category]=""
+                packagesByCategory[category]+="|" + package + "|" + description + "|\n"
+                print (f"| {package} | {description} |")
+        except FileNotFoundError:
```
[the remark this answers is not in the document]
Ana06: In general: `sys.exit()` exists with return code `0` unless you provide another return code as argument, which means the script won't fail. Do not use `print` + `exit` and use Python exceptions instead. 

In this case: I do not understand how this error can happen and I think the `except` should be remove. Or am I missing something? When can this happen?

.github/workflows/generate_package_wiki.yml
```diff
@@ -24,7 +24,7 @@ jobs:
 
       - name: Generate Wiki Content
         id: generate-content
-        run: scripts/utils/generate_package_wiki.ps1
+        run: python scripts/utils/generate_package_wiki.py packages/ --wiki wiki/Packages.md
```
Ana06, line 27: remove the arguments here and provide them as defaults in the script.

scripts/utils/generate_package_wiki.py
```diff
@@ -0,0 +1,70 @@
...
+
+        except FileNotFoundError:
+            print(f"Error: File '{nuspec_path}' not found.")
+            sys.exit()
+
+
+if __name__ == "__main__":
+
+    parser = argparse.ArgumentParser(description="Create wiki of packages.")
+    parser.add_argument("packages", type=str, help="Path to packages")
```
Ana06, line 66: make this argument optional to provide a default

Ana06 (changes requested):

scripts/utils/generate_package_wiki.py
sara-rn: not sure if I understood this

Ana06: the code can be simplified as suggested (removing the `is None` and removing the `else`) as the code is equivalent. what is exactly not clear?

scripts/utils/generate_package_wiki.py
```diff
@@ -0,0 +1,79 @@
+import argparse
+import os
+import pathlib
+import xml.etree.ElementTree as ET
+from collections import defaultdict
+
+packagesByCategory = defaultdict(str)
```
Ana06, line 7: As I already pointed out in [my previous comment](https://github.com/mandiant/VM-Packages/pull/1320#discussion_r1981151327), follow [variables naming convention in Google Python Style Guide](https://google.github.io/styleguide/pyguide.html#316-naming)
```suggestion
packages_by_category = defaultdict(str)
```

Ensure you replace it in the comments too.

scripts/utils/generate_package_wiki.py
```diff
@@ -0,0 +1,79 @@
...
+
+
+def find_element_text(parent, tag_name):
+    """Retrieves the text from a xml node, it validates if the tag exists
+    Args:
+       parent: parent tree to search
+       tag_name: tag to search the text
+    Returns:
+       the tag text value
+    """
```
Ana06, line 37: [nitpick] I think the format explained in https://github.com/google/styleguide/blob/gh-pages/pyguide.md is as follows (note the empty lines):
```
"""A one-line summary

Overall description of the module or program. Possibly multiline.

Section1:
"""
```

I asked Gemini to help me in this case (by providing the function code and asking it to write the docstring):
```suggestion
    """Retrieves the text content of a specified XML element.

    This function searches for a child element with the given tag name within the
    provided parent element. It handles XML namespaces by using a wildcard.

    Args:
        parent: The parent XML element.
        tag_name: The name of the child element to find.

    Returns:
        The text content of the found element, or None if the element is not found.
    """
```

scripts/utils/generate_package_wiki.py
```diff
@@ -0,0 +1,88 @@
+import argparse
+import os
+import pathlib
+import xml.etree.ElementTree as ET
+from collections import defaultdict
+
+packages_by_category = defaultdict(str)
```
Ana06, line 7: [nitpick] I think it would be helpful to document the type of this variable:
```suggestion
# Dict[str (category), list (packages information)]
packages_by_category = defaultdict(str)
```

scripts/utils/generate_package_wiki.py
```diff
@@ -0,0 +1,88 @@
...
+    
+    This function retrieves rows pertaining to the key category, as a result of
+    iterating over packages_by_category.
+    Writes to the Markdown formatted string:
+    - header with the category name
+    - table with two columns: Package and Description of each of the packages
+    
+    Args:
+       file_path: path of the wiki .md file
+    """
```
Ana06, line 21: [nitpick]
```suggestion
    """ Writes package information sorted by category to a Markdown wiki file.

    This function iterates through the `packages_by_category` dictionary, which
    contains package information organized by category. For each category, it
    generates a Markdown header and a table containing package names and
    descriptions. The resulting Markdown content is then written to the specified
    file.

    Args:
        file_path (str): The path to the output Markdown file.
    """
```

scripts/utils/generate_package_wiki.py
```diff
@@ -0,0 +1,88 @@
...
+    tag = parent.find(f"{{*}}{tag_name}")
+    if tag is None:
+        return None
+    return tag.text
+
+
+def process_packages_directory(packages_dir):
+    """ Obtains the package name, description and category from a directory
+    
+    This function parses all the nuspec files in the 'packages' directory structure.
```
Ana06, line 54: [nitpick]
```suggestion
    This function parses all the nuspec files in the specified packages directory.
```

scripts/utils/generate_package_wiki.py
```diff
@@ -0,0 +1,88 @@
...
+    It saves into packages_by_category[category] a line containing the package
+    name and the description separated by |
```
Ana06, line 56: [nitpick]
```suggestion
    name and the description separated by "|".
```

scripts/utils/generate_package_wiki.py
```diff
@@ -0,0 +1,88 @@
...
+    
+    Args:
+        packages: directory where all the packages reside
```
Ana06, line 59: [nitpick]
```suggestion
        packages: directory where the packages reside.
```

scripts/utils/generate_package_wiki.py
```diff
@@ -0,0 +1,88 @@
...
+        raise FileNotFoundError(f"Packages directory not found: {packages_dir}")
+
+    for nuspec_path in pathlib.Path(packages_dir).glob("**/*.nuspec"):
+        nuspec_tree = ET.parse(nuspec_path)
+        nuspec_metadata = nuspec_tree.find("{*}metadata")
+        if nuspec_metadata is not None:
+            category = find_element_text(nuspec_metadata, "tags")
+            # we are only interested in the packages with a category
+            # some packages that assist with the istallation (such as common.vm) do
+            # not contain a category
```
Ana06, line 71: [nitpick]
```suggestion
            # We are only interested in the packages with a category.
            # Some packages that assist with the istallation (such as common.vm) do
            # not contain a category
```

scripts/utils/generate_package_wiki.py
```diff
@@ -0,0 +1,88 @@
...
+            if not category:
+                continue
+            description = find_element_text(nuspec_metadata, "description")
+            package = find_element_text(nuspec_metadata, "id")
+            """packages_by_category is a dictionary of str where each package and description
+            are appended formatted in Markdown"""
+            packages_by_category[category] += "|" + package + "|" + description + "|\n"
```
Ana06, line 78: Adding some spaces makes the markdown nicer to read:
```suggestion
            packages_by_category[category] += "| " + package + " | " + description + " |\n"
```

scripts/utils/generate_package_wiki.py
```diff
@@ -0,0 +1,88 @@
...
+    - table with two columns: Package and Description of each of the packages
+    
+    Args:
+       file_path: path of the wiki .md file
+    """
+    wikiContent = ""
+    for category in packages_by_category.keys():
+        wikiContent += "## " + category + "\n\n"
+        wikiContent += "| Package | Description |\n"
+        wikiContent += "|---|---|\n"
```
Ana06, line 26: [nitpick] It makes it the code and the markdown a bit nicer to read:
```suggestion
        wikiContent += "| ------- | ----------- |\n"
```

Ana06 (commented):
This looks good @sara-rn, just some minor improvement suggestion :wink:

scripts/utils/generate_package_wiki.py
Ana06: I feel with the added documentation, it is ok to keep the current code structure.

Ana06 (approved):
Nice code @sara-rn, simple and well documented, easy to understand and maintain! :bouquet: Looking forward triggering the action! :rocket:

Issue #1173, opened 2024-11-22
Update ida.plugin.capa.vm


Ana06:
### Details

Our bot is not updating `ida.plugin.capa.vm`. I think we should:
1 - Update `ida.plugin.capa.vm` to the latest version (even if this need to be done manually as the current version does not work with IDA 9)
2 - Update [`update_package.py`](https://github.com/mandiant/VM-Packages/blob/main/scripts/utils/update_package.py) to update this package automatically in the future

@sara-rn would you like to work on this?

Ana06:
In addition to the two steps above, we need to fix the version of `flare-capa` that we install in the package to ensure they are in sync.

Ana06:
@sara-rn and I have noticed that the script does not work because there are two hashes to update. This is a special case and we think it is not worth updating the script for this rare case.

Remaining in this issue is to set the version of `flare-capa` and update to capa 8 after it is released.

Ana06:
capa 8.0 that fixes the issue has been released. @sara-rn does not have time this week for this, so I am taking over. :wink:

PR #1323, opened 2025-03-07
Update the ida plugin capa to version 9.0.0

### LINKED ISSUE #1285: Update ida.plugin.capa.vm
### Details

As explained in https://github.com/mandiant/VM-Packages/issues/1173, we need to update ida.plugin.capa.vm manually. I think we should do this at least when there is a major release, and capa 9 has been released!

sara-rn:
Update must be done manually, tracked in
https://github.com/mandiant/VM-Packages/issues/1173
closes #1285


Ana06 (approved):
Thanks @sara-rn! :bouquet: There is already a 9.1.0 release (released 3 days ago), we may want to update to that version as well. :wink:

sara-rn:
@Ana06 I noted that there was a new version of capa rules, however the URL for capa_explorer 9.0.1 is not available, and we cannot have 2 different versions right? I understand that `capa_explorer.py` doesn't change often so maybe we can consider having 2 different versions, one for `capa_explorer.py`, another for rules

PR #1325, opened 2025-03-13
[CI] Enable Python linters

Ana06:
Enable Python linters and format checkers as in FLARE-VM: 
- https://github.com/mandiant/flare-vm/blob/main/.github/workflows/linter.yml

Fix offenses detected by the linters. isort and black were used to solve their offenses. The flake8 offenses were fixed manually. Because of that I would appreciate a thoughtful review to ensure I haven't introduced any bug.


sara-rn (approved):
looks good to me

Issue #1176, opened 2024-11-22
choco fails to install dcode.vm


micahflack:
### What's the problem?

Both the automatic flarevm script and manual installation of the dcode.vm package fails to download. The installer can still be downloaded with any browser though. I thought that this might be caused by the URL being redirected during download...

- https://www.digital-detective.net/download/**download.php?**downcode=ae2znu5994j1lforlh03
- https://www.digital-detective.net/download/**downloadbac.php?**downcode=ae2znu5994j1lforlh03

But neither of these URLs works within the chocolatey installer. See my attempt at earlier PR - Fixed broken URL for dcode.vm #1167

Looking back through the daily tests (1-2 weeks) and commits for dcode.vm, nothing has been changed within the package that could cause downloader issues. However, there was a recent release update to Chocolatey from 2.3.0 to 2.4.0. Downgrading the Chocolatey package produced the same 404 error as well.

I still think the problem lies further upstream with Chocolatey and how the command "Install-ChocolateyZipPackage @packageArgs" is handled.

### Steps to Reproduce

1. Prepare VM following "pre-installation" steps
2. Unblock-File .\install.ps1
3. Set-ExecutionPolicy Unrestricted -Force
4. .\install.ps1 -password <password>
5. select full download
6. installer finishes but dcode.vm fails to install

Manual Steps (after complete FlareVM install)

1. Open powershell as admin
2. choco install dcode.vm
3. (installer attempts download)
4. fails because of missing or unauthorized files

Choco downgrade and manual install

1. (complete flarevm installation)
2. (reboot)
3. choco upgrade chocolatey --version 2.3.0 --allow-downgrade
4. (reboot)
5. choco install dcode.vm
7. (installer attempts download)
8. fails because of missing or authorized files

### Environment

Virtualization software: Proxmox
VM OS version: 10.0.19044
VM PowerShell version: 5.1.19041.3031
VM Chocolatey version: 2.4.0
VM Boxstarter version: Boxstarter|3.0.3
VM-Get-Host-Info:

```
PS C:\Users\Flare > VM-Get-Host-Info
2024/11/22 15:29:19 vm.common.psm1 [+] INFO : Host Information

VM OS version and Service Pack
-----


Version                 : 10.0.19044
BuildNumber             : 19044
OSArchitecture          : 64-bit
ServicePackMajorVersion : 0
Caption                 : Microsoft Windows 10 Enterprise LTSC





VM OS RAM (MB)
-----
0


[...]

PR #1337, opened 2025-04-08
Remove dcode linter exception

sara-rn:
Dcode package now installs correctly therefore the linter no longer need an exclusion for the package.
See https://github.com/mandiant/VM-Packages/issues/1176
Now it fails, a rebase needs to be done after #1338 is merged


Ana06 (approved):
Thanks @sara-rn!

Issue #1332, opened 2025-04-06
Pestudio tool URL structure changed


PrajeetGuha:
### What's the problem?

The URL in the pestudio package is giving 404 error code. This is due to change in version but also changes to the structure of the URL which might not be detected by automated VM package updates.

Existing URL link - 
https://github.com/mandiant/VM-Packages/blob/9cb18b33df3f955a3a93e8d386dcaf893e86721d/packages/pestudio.vm/tools/chocolateyinstall.ps1#L7
New link - 
https://www.winitor.com/tools/pestudio/current/pestudio.zip

It is also a new version of pestudio (pestudio9.61) so hash will also change.

### Steps to Reproduce

Try to do a fresh install of pestudio.vm

### Environment

Will be reproducible in any working environment

### Additional Information

**log.txt**
2025/04/06 13:06:06 [installer.vm] chocolateyinstall.ps1 [+] INFO : Installing: pestudio.vm
2025/04/06 13:06:24 [pestudio.vm] vm.common.psm1 [+] ERROR : The remote file either doesn't exist, is unauthorized, or is forbidden for url 'https://www.winitor.com/tools/pestudio/current/pestudio-9.60.zip'. Exception calling "GetResponse" with "0" argument(s): "The remote server returned an error: (404) Not Found."
At C:\ProgramData\chocolatey\helpers\functions\Get-WebFile.ps1:351 char:13
             throw "The remote file either doesn't exist, is unauthori ...
             ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~

**URL response**

![Image](https://github.com/user-attachments/assets/511f0493-b6f8-4e5f-b6f5-84e3228a595a)


### Detected by test suite

Yes

Ana06:
[...]

PR #1341, opened 2025-04-10
fix pestudio package

PrajeetGuha:
PR is related to https://github.com/mandiant/VM-Packages/issues/1332


Ana06 (approved):
Thanks for the fix @PrajeetGuha! I tested locally and works as expected :rocket:

Issue #1336, opened 2025-04-08
URL package proposal: GoStringUngarbler.vm


sara-rn:
### Package Name

gostringungarbler

### Tool Name

gostringungarbler

### Package type

OTHER/UNKNOWN

### Is the tool a console application?

true

### Version

1.0.0

### Category

Packers

### Tool's authors

Chuong Dong

### Tool's description

GoStringUngarbler is a deobfuscator tool that resolves all strings in Go binaries obfuscated by garble.

### Download URL

https://github.com/mandiant/gostringungarbler/archive/refs/tags/v1.0.0.zip

### Download SHA256 Hash

8fe3278cc04f62782eaf7c17813cd14228160187b87d9aa28514c419986acacf

### Dependencies

python, unicorn, lief, capstone

### Why is this tool a good addition?

Reverse engineering Go binaries obfuscated with garble presents a set of unique challenges, including stripped binaries, function name mangling, and encrypted strings.

Ana06:
@sara-rn it is common that IDA plugins require to install the file `dependencies.txt`, as it happens in this package and for example also in xrefer: https://github.com/mandiant/VM-Packages/blob/9cb18b33df3f955a3a93e8d386dcaf893e86721d/packages/ida.plugin.xrefer.vm/tools/chocolateyinstall.ps1#L6

What about checking in the `VM-Install-IDA-Plugin` if that file exists and if so, install the dependencies with pip? This will ensure the plugin will continue to work on updates without making manual changes to update the dependencies and will allow us to create this package automatically (using the bot).

I propose the following steps:
- One PR to extend `VM-Install-IDA-Plugin` to check if `dependencies.txt` exists and install the dependecies if so. Update also in this PR xrefer so that we can test the changes
- Create this package using automation.

Ana06:
@sara-rn the proposed enhacement would also make it very easy to address https://github.com/mandiant/VM-Packages/issues/1230, I have assigned it to you.

Ana06:
@sara-rn you can't trigger the automation with `OTHER/UNKNOWN` package type (only for the supported defined types). I think this package is similar to `unpyc3` and we do not have automation for this.

sara-rn:
@Ana06  the problem is that I cannot use `VM-Pip-Install` with the URL, I get the following error.
I am afraid this will take longer to implement as I cannot use any of the helper functions in `common.vm`
```
[...]

PR #1342, opened 2025-04-11
Add new category Go

sara-rn:
Add category Go for tools such as goresym and gostringungarbler (https://github.com/mandiant/VM-Packages/issues/1336)

### DIFF
--- a/.github/ISSUE_TEMPLATE/new_metapackage.yml
+++ b/.github/ISSUE_TEMPLATE/new_metapackage.yml
@@ -62,6 +62,7 @@ body:
         - Exploitation
         - File Information
         - Forensic
+        - Go
         - Hex Editors
         - InnoSetup
         - Java and Android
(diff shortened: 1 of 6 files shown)


Ana06 (commented):
The addition of the Go category LGTM, but can you also change GoReSym of category in this PR? This way it is easier to understand the need of this change and verify it works as expected.

Please keep the addition of the category in just one commit (it only adds 5 simple lines that need to be added at the same time to keep the categories in sync).

Ana06:
@sara-rn the linter should not fail if you add the category in the same PR. Adding the category in the same PR that was needed is how we typically have done this in the past, for example: https://github.com/mandiant/VM-Packages/pull/1186 Or has anything changed since then that I am missing? :thinking:

sara-rn:
@Ana06 my bad, just wanted to point out why I submitted 2 different PRs but you are right, the linter relies on `categories.txt`

PR #1343, opened 2025-04-11
Add package GoStringUngarbler

### LINKED ISSUE #1336: URL package proposal: GoStringUngarbler.vm

sara-rn:
Add new tool to deobfuscate strings in Go binaries
needs to be rebased after #1342 has been merged
closes #1336

### DIFF
--- a/packages/gostringungarbler.vm/gostringungarbler.vm.nuspec
+++ b/packages/gostringungarbler.vm/gostringungarbler.vm.nuspec
@@ -1,0 +1,14 @@
+<?xml version="1.0" encoding="utf-8"?>
+<package xmlns="http://schemas.microsoft.com/packaging/2015/06/nuspec.xsd">
+  <metadata>
+    <id>gostringungarbler.vm</id>
+    <version>1.0.0</version>
+    <authors>Chuong Dong</authors>
+    <description>GoStringUngarbler deobfuscates strings in Go binaries obfuscated by garble</description>
+    <dependencies>
+      <dependency id="common.vm" version="0.0.0.20250206" />
+      <dependency id="python3.vm" />
+    </dependencies>
+    <tags>Go</tags>
+  </metadata>
+</package>
--- a/packages/gostringungarbler.vm/tools/chocolateyinstall.ps1
+++ b/packages/gostringungarbler.vm/tools/chocolateyinstall.ps1
@@ -1,0 +1,43 @@
+$ErrorActionPreference = 'Stop'
+Import-Module vm.common -Force -DisableNameChecking
+
+$toolName = 'GoStringUngarbler'
+$category = VM-Get-Category($MyInvocation.MyCommand.Definition)
+
+try {
+    $zipUrl = 'https://github.com/mandiant/gostringungarbler/archive/refs/tags/v1.0.0.zip'
+    $zipSha256 = '8fe3278cc04f62782eaf7c17813cd14228160187b87d9aa28514c419986acacf'
+    $toolDir = Join-Path ${Env:RAW_TOOLS_DIR} $toolName
+
+    # Remove files from previous zips for upgrade
+    VM-Remove-PreviousZipPackage ${Env:chocolateyPackageFolder}
+
+    $packageArgs = @{
+        packageName    = ${Env:ChocolateyPackageName}
+        unzipLocation  = $toolDir
+        url            = $zipUrl
+        checksum       = $zipSha256
+        checksumType   = 'sha256'
+    }
+    Install-ChocolateyZipPackage @packageArgs | Out-Null
+
+    VM-Assert-Path $toolDir
+
+    # There is an inner folder in the zip whose name changes as it includes the version
+    $dirList = Get-ChildItem $toolDir -Directory
+    $toolDir = Join-Path $toolDir $dirList[0].Name -Resolve
+
+    # Install dependencies in the "requirements.txt" file.
+    $requirementsFile = Join-Path $toolDir "requirements.txt" -Resolve
+    Get-Content -Path $requirementsFile | ForEach-Object {
+    	VM-Pip-Install $_
+    }
+
+    $pyPath = (Get-Command py).Source
+    $filePath = Join-Path $toolDir "$toolName"
+    $arguments = "-3.10 " + $filePath + " -h"
+    VM-Install-Shortcut -toolName $toolName -category $category -executablePath $pyPath -consoleApp $true -arguments $arguments
+
+} catch {
+  VM-Write-Log-Exception $_
+}
--- a/packages/gostringungarbler.vm/tools/chocolateyuninstall.ps1
+++ b/packages/gostringungarbler.vm/tools/chocolateyuninstall.ps1
@@ -1,0 +1,7 @@
+$ErrorActionPreference = 'Continue'
+Import-Module vm.common -Force -DisableNameChecking
+
+$toolName = 'GoStringUngarbler'
+$category = VM-Get-Category($MyInvocation.MyCommand.Definition)
+
+VM-Uninstall $toolName $category


packages/gostringungarbler.vm/tools/chocolateyinstall.ps1
```diff
@@ -0,0 +1,42 @@
+$ErrorActionPreference = 'Stop'
+Import-Module vm.common -Force -DisableNameChecking
+
+$toolName = 'GoStringUngarbler'
+$category = VM-Get-Category($MyInvocation.MyCommand.Definition)
+
+$dependencies = "unicorn,capstone,lief"
+VM-Pip-Install $dependencies
```
Ana06, line 8: Instead of hardcoding the dependencies, I would move this code after downloading the ZIP and read the dependencies from the [requirements.txt](https://github.com/mandiant/gostringungarbler/blob/main/requirements.txt) file. This way if the dependencies change with an update, our bot would still be able to update the tools without any manual modification. This is easy to do by running `py -3.1.10 -m pip install -r requirements.txt` as I mentioned in https://github.com/mandiant/VM-Packages/issues/1336#issuecomment-2794545117.

I know this is hardcoded in other packages, but we should move away from this approach.

packages/gostringungarbler.vm/tools/chocolateyinstall.ps1
```diff
@@ -0,0 +1,42 @@
...
+
+$ErrorActionPreference = 'Stop'
+Import-Module vm.common -Force -DisableNameChecking
+
+try {
+    $category = VM-Get-Category($MyInvocation.MyCommand.Definition)
```
Ana06, line 14: Duplicated code
```suggestion
try {
```

packages/gostringungarbler.vm/tools/chocolateyinstall.ps1
```diff
@@ -0,0 +1,42 @@
...
+
+    VM-Assert-Path $toolDir
+
+    $pyPath = (Get-Command py).Source
+    $filePath = Join-Path $toolDir "$toolName"
+    $arguments = "-3.10 " + $filePath + " -h"
+    VM-Install-Shortcut -toolName $toolName -category $category -executablePath $pyPath -consoleApp $true -arguments $arguments
+
+    # Add tools to Path
+    VM-Add-To-Path $toolDir
```
Ana06, line 39: [question] Why is this needed? why do we need to use `VM-Add-To-Path` instead of `Install-BinFile`?

sara-rn: I have seen this in other packages, will use the cmdlet `Install-BinFile` (I didn't know it)

sara-rn: I have looking into the usage of `Install-BinFile` but I am not 100% sure it is the right option because this is a python package, I have only seen `Install-BinFile` with .exe and .bat files, please correct me if I am wrong

Ana06:
@sara-rn I assume you have used the GH interface to accept a code suggestion. This has added me as co-author with my personal account breaking the Google CLA check. In general, I advise not to use GH interface to create commits (with things like accepts code suggestions and rebasing) and instead do this locally using `git`. It is IMO more intuitive and you known what is going on. You now need to squash/remove the commit created by GH.

sara-rn:
> @sara-rn I assume you have used the GH interface to accept a code suggestion. This has added me as co-author with my personal account breaking the Google CLA check. In general, I advise not to use GH interface to create commits (with things like accepts code suggestions and rebasing) and instead do this locally using `git`. It is IMO more intuitive and you known what is going on. You now need to squash/remove the commit created by GH.

I did squash the commits right after that, no errors.

packages/gostringungarbler.vm/tools/chocolateyinstall.ps1
sara-rn: @Ana06 I have implemented the changes, let me know if you still think `Install-BinFile` is appropiate for a python package.

Ana06: Our install from Zip helper function uses the chocolatey helper `Install-BinFile`, which I think adds the executable to path. `VM-Add-To-Path` has introduced in https://github.com/mandiant/VM-Packages/pull/806 to add whole directories to path. We use it in didier-stevens-suite.vm and didier-stevens-beta.vm for this.

We also use it in apktool.vm, but I think this could also be replace by `Install-BinFile` (I am not completely sure, it needs to be tested). If so, we should default to the Chocolatey helper and use our more complicated function only when needed. We should then add more documentation to `VM-Add-To-Path` explaining when to use it and to use `Install-BinFile` for other cases.

@sara-rn can you please test if using `Install-BinFile` achieves the desired result for this package and apktool.vm?

packages/gostringungarbler.vm/tools/chocolateyinstall.ps1
sara-rn: I didn't implement exactly like you suggested but I read the `requirements.txt` file to install the dependencies.

Ana06: I think your implementation is equivalent to what I was proposing as `VM-Pip-Install` runs `py -3.1.10 -m pip install`

packages/gostringungarbler.vm/tools/chocolateyinstall.ps1
```diff
@@ -0,0 +1,45 @@
...
+    }
+    Install-ChocolateyZipPackage @packageArgs | Out-Null
+
+    VM-Assert-Path $toolDir
+
+    #there is an inner folder
+    $dirList = Get-ChildItem $toolDir -Directory
+    $toolDir = Join-Path $toolDir $dirList[0].Name -Resolve
+
+    $requirementsFile = Join-Path $toolDir "requirements.txt"
```
Ana06, line 31: [nitpick] Add a comment above this line explaining what the next lines do:
```
Install dependencies in the "requirements.txt" file.

File: packages/gostringungarbler.vm/tools/chocolateyinstall.ps1 (hunk 3 above)
Comment (sara-rn) [reply packages/gostringungarbler.vm/tools/chocolateyinstall.ps1:42]: It doesn't work, it seems its only for executables, see error:
```
PS C:\Users\flaretest\Desktop> GoStringUngarbler
Cannot find file at 'c:\tools\gostringungarbler\gostringungarbler-1.0.0' (c:\tools\gostringungarbler\gostringungarbler-1.0.0). This usually indicates a missing or moved file.
```

**Comment by sara-rn** [comment]
@Ana06 I left `VM-Add-To-Path` instead `Install-Binfile` as per my previous comment. I read the documentation and tested it as well: https://docs.chocolatey.org/en-us/create/functions/install-binfile/

File: packages/gostringungarbler.vm/gostringungarbler.vm.nuspec [hunk 5]
```diff
@@ -0,0 +1,14 @@
+<?xml version="1.0" encoding="utf-8"?>
+<package xmlns="http://schemas.microsoft.com/packaging/2015/06/nuspec.xsd">
+  <metadata>
+    <id>gostringungarbler.vm</id>
+    <version>1.0.0</version>
+    <authors>Chuong Dong</authors>
+    <description>GoStringUngarbler deobfuscates strings in Go binaries obfuscated by garble</description>
+    <dependencies>
+      <dependency id="common.vm" version="0.0.0.20240509" />
```
Ana06, line 9: `VM-Get-Category` was introduced in (version `0.0.0.20250206` of common.vm`). We need to require that version or the package installation may fail.
```suggestion
      <dependency id="common.vm" version="0.0.0.20250206" />
```

Ana06 (changes requested):
@sara-rn can you squash the commits?

packages/gostringungarbler.vm/gostringungarbler.vm.nuspec
Ana06: I have also opened an issue to fix the scripts: https://github.com/mandiant/VM-Packages/issues/1369

packages/gostringungarbler.vm/tools/chocolateyinstall.ps1
Ana06: @sara-rn I have just the package and `GoStringUngarbler` does not appear to be added to PATH, or am I understanding something wrong?
```
C:\Users\flare>GoStringUngarbler
'GoStringUngarbler' is not recognized as an internal or external command,
operable program or batch file.

FLARE-VM Wed 04/23/2025  9:33:44.83
```

sara-rn: @Ana06 you are right in this particular case adding the folder to the path doesn't work either, I suggest removing that line as well because there is no compiled executable or even a `GoStringUngarbler.py` therefore it needs to be executed with the params that the shortcut in the Desktop points to.

Ana06: Yes, please remove it.

Ana06 (approved):
:rocket:

PR #1345, opened 2025-04-11
Update MFTECmd.vm to fix CI

naacbin:


Ana06:
@naacbin thanks for fixing the hash. @PrajeetGuha is implementing automatic hash fixes in https://github.com/mandiant/VM-Packages/pull/1350 and we are using mftecmd.vm to test it. So I am going to close this PR and hope we can get https://github.com/mandiant/VM-Packages/pull/1350 soon in. This means keeping the package broken for some more days, but It is going to be an important enhancement in the long term.

Issue #1270, opened 2025-02-06
Add "fix broken hashes" automation


Ana06:
### Details

Add automation to automatically update the hash of broken packages, similar to the package update. This automation should only be triggered manually and we should add to the PR description that the update needs to be reviewed.

PrajeetGuha:
Can you provide more details regarding the issue?

Ana06:
What exactly is not clear @PrajeetGuha?

PrajeetGuha:
The CI automation automatically updates package url, version and respective hash. By the term hash of broken packages you mean
Packages that uses dynamic links and due to changes in the file in the link, hash does not match?
or something else that I am missing?

PrajeetGuha:
Is it same as this? https://github.com/mandiant/VM-Packages/issues/91

nasirhm:
Hey @PrajeetGuha, are you working on it ? I can help out with it.

cc: @Ana06

PrajeetGuha:
@nasirhm I have not started working on it. You can work on it if you want to.

nasirhm:
@Ana06 I would like to work on it.

Ana06:
@PrajeetGuha we have some URLs that do not include the version. When a tool update happens, the URL stays the same, but the hash changes. This breaks the package as the hash verification fails. This issue is about adding a new automation (or extending the current one) to update the hash when it does not match the current one. You are right it is a duplicate of #91.

Ana06:
@nasirhm note that this issue requires some knowledge of GitHub Actions, Python, PowerShell and how the project works and is structured. So it may not be the easiest issue to start contributing to the project. But if it is something you find interesting and you want to give it a try, please go for it! Let us know if you plan to work on it, so that otherwise @PrajeetGuha or @sara-rn can work on it.

And sorry for the late reply, I have missed this. 😅

Ana06:
@nasirhm can you please let us know if you are planing to work on this so that otherwise someone else can start working on it?

Ana06:
@PrajeetGuha would you like to work on this?

Ana06:
@PrajeetGuha 
> Just a question regarding it. Is there any reason for skipping UpdateType 3 in the **[update_package.py](https://github.com/mandiant/VM-Packages/blob/main/scripts/utils/update_package.py)**? 🤔 

[...]

PR #1350, opened 2025-04-12
Add broken hash fix automation

PrajeetGuha:
related to https://github.com/mandiant/VM-Packages/issues/1270
Changes made -
- added a workflow that can be manually dispatched to find broken hashes and create PR. 👍 
- added DYNAMIC_URL functionality in the update_package.py that is invoked by the above workflow 👍 
- made changes to pestudio.vm to create a test case for the workflow. After PR merge, we can execute the workflow manually to test if pestudio.vm broken hash can be captured. 👍 
- made changes to mftecmd.vm to check the workflow but there is failure, the cause is still unknown to me. 👎 

I have tested the workflow by introducing some changes in pestudio.vm, it works fine.
image
The workflow is also creating the PR as required. (the PR is created by bot, but PR creation name is coming as me, that I am not sure how to change)
image

I also tested the workflow with mftecmd.vm but I am not sure why it is failing. I have done this testing in my own personal Github fork of this repository.
Workflow logs - 
image
git diff shows hash and version change, but the workflow logs state that failure is due to wrong hash and it is showing the older hash not the newer hash.
image

Closes https://github.com/mandiant/VM-Packages/issues/1270

### DIFF
--- a/.github/workflows/update_packages.yml
+++ b/.github/workflows/update_packages.yml
@@ -26,8 +26,8 @@ jobs:
           foreach ($packagePath in (Get-ChildItem packages)){
             $package = $packagePath.Name
             $newVersion = 0
-            # Test indepdendly every type of update and commit what works
-            foreach ($UPDATE_TYPE in ('DEPENDENCIES', 'GITHUB_URL', 'VERSION_URL')) {
+            # Test independently every type of update and commit what works
+            foreach ($UPDATE_TYPE in ('DEPENDENCIES', 'GITHUB_URL', 'VERSION_URL', 'DYNAMIC_URL')) {
               $version = python scripts\utils\update_package.py $package --update_type $UPDATE_TYPE
               $updated = $?
               echo "$package $version"
@@ -39,6 +39,8 @@ jobs:
                 if ($tested) {
                   git add "packages/$package/"
                   $newVersion = $version
+                  # Save the update type to use it in the commit
+                  $finalUpdateType = $UPDATE_TYPE
                 } else {
                   echo "$package $version FAILED"
                   git diff
@@ -49,7 +51,12 @@ jobs:
               Remove-Item built_pkgs -Recurse -ErrorAction Ignore
             }
             if ($newVersion) {
+              if ($finalUpdateType -eq 'DYNAMIC_URL') {
+                git commit -m "Fix broken hash in $package"
+              }
+              else {
               git commit -m "Update $package to $newVersion"
+              }
             }
           }
           Exit(0)
--- a/scripts/utils/update_package.py
+++ b/scripts/utils/update_package.py
@@ -4,6 +4,7 @@ import os
 import re
 import time
 from enum import IntEnum
+from pathlib import Path
 
 import requests
 
@@ -47,7 +48,7 @@ def get_latest_version(org, project, version):
         return latest_version
 
 
-# Get url response's content hash (SHA256)
+# Get URL response's content SHA256 hash
 def get_sha256(url):
     response = requests.get(url)
     if not response.ok:
@@ -69,42 +70,55 @@ def format_version(version):
     return match.group("version")
 
 
-# Replace version in the package's nuspec file
-def update_nuspec_version(package, latest_version):
+# get content of nuspec file
+def get_nuspec(package):
     nuspec_path = f"packages/{package}/{package}.nuspec"
     with open(nuspec_path, "r") as file:
         content = file.read()
+    return (nuspec_path, content)
+
+
+# get nuspec version
+def get_version_from_nuspec(package):
+    _, content = get_nuspec(package)
+
+    # find the version from nuspec
+    return re.findall(r"<version>(?P<version>[^<]+)</version>", content)[0]
+
+
+# check if the package has a `chocolateyinstall.ps1` file, some packages do not have one
+def check_install_script_present(package):
+    install_script_path = f"packages/{package}/tools/chocolateyinstall.ps1"
+    return Path(install_script_path).is_file()
+
+
+# Replace version in the package's nuspec file
+def update_nuspec_version(package, latest_version):
+    nuspec_path, content = get_nuspec(package)
     latest_version, content = replace_version(latest_version, content)
     with open(nuspec_path, "w") as file:
         file.write(content)
+    return latest_version
 
 
 # read the chocolateyinstall.ps1 package file
 def get_install_script(package):
     install_script_path = f"packages/{package}/tools/chocolateyinstall.ps1"
-    try:
-        file = open(install_script_path, "r")
-    except FileNotFoundError:
-        # chocolateyinstall.ps1 may not exist for metapackages
-        return (None, None)
-    return (install_script_path, file.read())
+    with open(install_script_path, "r") as file:
+        content = file.read()
+    return (install_script_path, content)
 
 
 # Update package using GitHub releases
 def update_github_url(package):
     install_script_path, content = get_install_script(package)
-    # Use findall as some packages have two urls (for 32 and 64 bits), we need to update both
-    # Match urls like https://github.com/mandiant/capa/releases/download/v4.0.1/capa-v4.0.1-windows.zip
-
-    # No chocolateyinstall.ps1 available
-    if not content:
-        return None
-
+    # Use findall as some packages have two URLs (for 32 and 64 bits), we need to update both
+    # Match URLs like https://github.com/mandiant/capa/releases/download/v4.0.1/capa-v4.0.1-windows.zip
     matches = re.findall(
         "[\"'](?P<url>https://github.com/(?P<org>[^/]+)/(?P<project>[^/]+)/releases/download/v?(?P<version>[^/]+)/[^\"']+)[\"']",
         content,
     )
-    # Match also urls like https://github.com/joxeankoret/diaphora/archive/refs/tags/3.0.zip
+    # Match also URLs like https://github.com/joxeankoret/diaphora/archive/refs/tags/3.0.zip
     matches += re.findall(
         "[\"'](?P<url>https://github.com/(?P<org>[^/]+)/(?P<project>[^/]+)/archive/refs/tags/v?(?P<version>[^/]+).zip)[\"']",
         content,
@@ -164,21 +178,15 @@ def get_increased_version(url, version):
     return (None, None)
 
 
-# Update package which uses a generic url that includes the version
+# Update package which uses a generic URL that includes the version
 def update_version_url(package):
     install_script_path, content = get_install_script(package)
-
-    # No chocolateyinstall.ps1 available
-    if not content:
-        return None
-
-    # Use findall as some packages have two urls (for 32 and 64 bits), we need to update both
-    # Match urls like:
+    # Use findall as some packages have two URLs (for 32 and 64 bits), we need to update both
+    # Match URLs like:
     # - https://download.sweetscape.com/010EditorWin32Installer12.0.1.exe
-    # - https://www.winitor.com/tools/pestudio/current/pestudio-9.53.zip
     matches = re.findall(r"[\"'](https{0,1}://.+?[A-Za-z\-_]((?:\d{1,4}\.){1,3}\d{1,4})[\w\.\-]+)[\"']", content)
 
-    # It doesn't include a download url with the version
+    # It doesn't include a download URL with the version
     if not matches:
         return None
 
@@ -208,12 +216,9 @@ def update_version_url(package):
 # Update dependencies
 # Metapackages have only one dependency and same name (adding `.vm`)  and version as the dependency
 def update_dependencies(package):
-    nuspec_path = f"packages/{package}/{package}.nuspec"
-    with open(nuspec_path, "r", encoding="utf-8") as file:
-        content = file.read()
+    nuspec_path, content = get_nuspec(package)
     matches = re.findall(
-        r'<dependency id=["\'](?P<dependency>[^"\']+)["\'] version="\[(?P<version>[^"\']+)\]" */>',
-        content,
+        r'<dependency id=["\'](?P<dependency>[^"\']+)["\'] version="\[(?P<version>[^"\']+)\]" */>', content
     )
 
     updates = False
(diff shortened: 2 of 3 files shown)


.github/workflows/fix_dynamic_url_hashes.yml
```diff
@@ -0,0 +1,64 @@
+name: Fix Broken Hashes
```
Ana06, line 1: While I would like to make clear when we are performing an update and when we are fixing a broken hash, I do not like having the duplicate the whole action logic (this action and `update_packages.yml` are almost identical). This is going to make hard keeping them in sync and increases maintenance work. What about having a single action (extending `update_packages.yml` instead of duplicating it in a new one) and using a different commit message for `DYNAMIC_URL`?

### Commit message for `DYNAMIC_URL`
```
Fix broken hash in $package
```
### Commit message for the rest
```
Update $package to $newVersion
```

packages/pestudio.vm/tools/chocolateyinstall.ps1
```diff
@@ -5,6 +5,6 @@ $toolName = 'pestudio'
 $category = VM-Get-Category($MyInvocation.MyCommand.Definition)
 
 $zipUrl = 'https://www.winitor.com/tools/pestudio/current/pestudio.zip'
-$zipSha256 = 'c1e2d0c1fbf5951486cf3d850cc24b11b66e25e0a5b77a623e2eb13ffad9ddd9'
+$zipSha256 = 'acc8eac2b5060d81c1285c2c6b050d18a045b7c659e6bf11cfcaada3c00bd1dd'
 
-VM-Install-From-Zip $toolName $category $zipUrl -zipSha256 $zipSha256 -innerFolder $true
+VM-Install-From-Zip $toolName $category $zipUrl -zipSha256 $zipSha256 -innerFolder $true
```
Ana06, line 10: Please remove these changes (and only keep the version change so that the script fixes the hash next time if breaks). We are not allowed to break packages in our CI. The only reason the PR CI succeeds is because the previous version is tested (because of the version decrease).

Ana06 (changes requested):
@PrajeetGuha thanks for your contribution, awesome work! :bouquet: 

I think the reason the PR is being created by you is because of the token you are using in your repository. The VM-Packages repository has the token of the bot configured, so I think this will work fine in this repo.

The test of mftecmd.vm is failing because you are decreasing the version in the nuspec. chocolatey takes the latest available version (which is in this case `1.3.0.20250219` and not `0.0.0.20250219`). Because this package is broken at the moment, **I have just deleted mftecmd.vm from MyGet**, so that there is no other version available apart from `0.0.0.20250219`. This way the test will test the right version. Please try it again, I expect it to succeed now.

For pestudio.vm, as the latest version available (`9.61.20250219` from MyGet) is already working, the test work. I'll delete the pestudio packages from MyGet just before merging this PR, as the package is working at the moment. This will ensure it works the next time the hash breaks.

In general, we need to delete the package from MyGet every time we decrease a version in the nuspec.

Ana06:
@PrajeetGuha the new `update_dynamic_url` function also needs to check if the chocolateyinstall.ps1 exists to prevent raising an exception, see https://github.com/mandiant/VM-Packages/pull/1353

scripts/utils/update_package.py
```diff
@@ -233,10 +240,52 @@ def update_dependencies(package):
...
+        # find the urls and sha256hashes in the chocolateyinstall.ps1
+        matches_url = re.findall(r"[\"'](https{0,1}://[^\"']+)[\"']", content)
+        matches_hash = re.findall(r"[\"']([a-fA-F0-9]{64})[\"']", content)
+
+        # if there is no matching url or no matching hashes or the number of matching url is not equal to number of matching hashes exit out
+        if not matches_url or not matches_hash or len(matches_url) != len(matches_hash):
+            return None
+
+        # find the new hash and check with existing hash and replace if different
+        for url, sha256 in zip(matches_url, matches_hash):
```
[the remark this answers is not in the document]
PrajeetGuha: This will also work for multiple url and hashes if the order in which the urls are written from top to bottom, matches the order in which the hashes are written from top to bottom.

packages/pestudio.vm/tools/chocolateyinstall.ps1
PrajeetGuha: Yes, I would revert this change. I did the change to test the CI.

PrajeetGuha:
> @PrajeetGuha the new `update_dynamic_url` function also needs to check if the chocolateyinstall.ps1 exists to prevent raising an exception, see https://github.com/mandiant/VM-Packages/pull/1353

I am thinking of moving the check of existence of chocolateyinstall.ps1 upstream into the main call as all the 3 updates - version url, github url and dynamic url will work only if there is a chocolateyinstall.ps1

This may make the script execution faster as control flow will not move into the respective functions due to absence of chocolateyinstall.ps1

PrajeetGuha:
The workflow is working - 😄 
image

scripts/utils/update_package.py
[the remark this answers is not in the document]
Ana06: Right. Please add a comment to document this, as it is not obvious.

packages/pestudio.vm/tools/chocolateyinstall.ps1
Ana06: We are keeping mftecmd.vm broken for some more days so that we can use it for testing here. :wink:

.github/workflows/update_packages.yml
```diff
@@ -48,7 +49,10 @@ jobs:
               git restore .
               Remove-Item built_pkgs -Recurse -ErrorAction Ignore
             }
-            if ($newVersion) {
+            if ($newVersion -and ($finalUpdateType -eq 'DYNAMIC_URL')){
+              git commit -m "Fix broken hash in $package"
+            }
+            elseif ($newVersion) {
```
Ana06, line 55: [nitpick] I think it reads nicer in two `ifs`:
```suggestion
            if ($newVersion) {
              if ($finalUpdateType -eq 'DYNAMIC_URL') {
                git commit -m "Fix broken hash in $package"
              }
              else {
```

scripts/utils/update_package.py
```diff
@@ -3,17 +3,28 @@
...
+)
 
 import requests
 
 
 # Replace version in nuspec, for example:
 # `<version>1.6.3</version>`
 # `<version>1.6.3.20220315</version>`
-def replace_version(latest_version, nuspec_content):
+def replace_version(
```
Ana06, line 19: There are some changes in the script that I think are not needed and caused by running the linters (likely black) with the wrong parameters. Having changes unrelated to the feature being implemented makes the PR review more difficult (as you need some time to figure out what changes are the ones that need to be review), so I would appreciate if you could remove them. You should be able to do this easily by running the linters with the same options as our CI: https://github.com/mandiant/VM-Packages/blob/main/.github/workflows/ci.yml#L23

@PrajeetGuha would it help if we document how to run the linters in the Wiki or contributing file? We have recently added the linters. :thinking:

scripts/utils/update_package.py
```diff
@@ -62,37 +86,109 @@ def get_sha256(url):
...
-    with open(nuspec_path, "r") as file:
+    with open(
+        nuspec_path,
+        "r",
+        encoding="utf-8",
+    ) as file:
         content = file.read()
-    latest_version, content = replace_version(latest_version, content)
-    with open(nuspec_path, "w") as file:
+    return content
```
Ana06, line 112: I think we should return the `nuspec_path` too (as we do in `get_install_script`). This way you do not need to redefine the `nuspec_path` in `update_dependencies` and `update_nuspec_version`.

scripts/utils/update_package.py
```diff
@@ -62,37 +86,109 @@ def get_sha256(url):
...
+
+    # find the version from nuspec
+    version = re.findall(
+        r"<version>(?P<version>[^<]+)</version>",
+        content,
+    )[0]
+    return version
+
+
+# check if install script is present or not
```
Ana06, line 129: [nitpick]
```suggestion
# check if the package has a `chocolateyinstall.ps1` file, some packages do not have one
```

scripts/utils/update_package.py
```diff
@@ -62,37 +86,109 @@ def get_sha256(url):
...
+def check_install_script_present(
+    package,
+):
+    install_script_path = f"packages/{package}/tools/chocolateyinstall.ps1"
+    if Path(install_script_path).is_file():
+        return True
+    else:
+        return False
```
Ana06, line 137:
```suggestion
    return Path(install_script_path).is_file()
```

scripts/utils/update_package.py
```diff
@@ -62,37 +86,109 @@ def get_sha256(url):
...
 # Update package using GitHub releases
-def update_github_url(package):
-    install_script_path, content = get_install_script(package)
+def update_github_url(
+    package,
+):
+    (
+        install_script_path,
+        content,
+    ) = get_install_script(package)
```
Ana06, line 191: These changes also look unrelated and caused by a code formater/linter. I find the new format more difficult to read. Please remove the unrelated changes to make easier to review the changes.

Ana06 (changes requested):
Thanks for addressing the feedback @PrajeetGuha, just some more minor changes requests. :wink:

scripts/utils/update_package.py
PrajeetGuha: Yes, wiki information about hoe to lint will be great.
Just a query regarding it - is there any way that the action can automatically lint and edit the PS Scripts to the standard format without passing th duty of formatting the scripts to match standards to the developers?

PrajeetGuha: And yes, I passed the code through black linter with wrong configuration which messed the formatting up.

scripts/utils/update_package.py
PrajeetGuha: Misconfigured black linter changes 😞

scripts/utils/update_package.py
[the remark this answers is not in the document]
PrajeetGuha: Ok. Will put the comment. Additionally if there is any coding practice related wiki or information it will be better to note the same there.

PrajeetGuha:
I could not get the time to test the changes. Give me 1 more day to test it before merging. If additional changes are required please let me know.

PrajeetGuha:
I ran black and flake8, i also have to run isort which is left.

scripts/utils/update_package.py
Ana06: Short answer: you need to run the formatter/linter locally. 

Long answer: If would be possible to create an action that fixes offenses and creates a new commit in the PR with the changes, but this would create an extra commit, and we avoid `fixup` commits to keep a clean commit history. The offenses should be fixed in the commit they are introduced instead of creating new commits. Another alternative is to configure `git` to prevent committing with offenses as we do in capa: https://github.com/mandiant/capa/blob/master/.pre-commit-config.yaml But this could confuse new contributors, so we are just allowing the CI to let people know.

I'll add some documentation.

.github/workflows/update_packages.yml
```diff
@@ -39,6 +39,7 @@ jobs:
                 if ($tested) {
                   git add "packages/$package/"
                   $newVersion = $version
+                  $finalUpdateType = $UPDATE_TYPE
```
Ana06, line 42: [nitpick] I think documenting why this is done would help understanding the code
```suggestion
                  # Save the update type to use it in the commit
                  $finalUpdateType = $UPDATE_TYPE
```

scripts/utils/update_package.py
```diff
@@ -69,37 +70,51 @@ def format_version(version):
     return match.group("version")
 
 
-# Replace version in the package's nuspec file
-def update_nuspec_version(package, latest_version):
+# get content of nuspec file
+def get_nuspec(package):
     nuspec_path = f"packages/{package}/{package}.nuspec"
-    with open(nuspec_path, "r") as file:
+    with open(nuspec_path, "r", encoding="utf-8") as file:
```
Ana06, line 76: [nitpick] why do we need to specify the encoding? we do not do it in other calls to `open` in the script :thinking:

Ana06 (changes requested):
@PrajeetGuha thanks for the changes! Great job writing the code and improving the documentation :cupid: I have some few more minor improvement suggestions.

Can you also squash the commits to remove the `Merge branch 'main' into main` commit? I am not sure if this commit is kept when I merge the PR. I recommend to rebase using the following comands instead of using the GH interface (I in general recommend no to use the GH interface to create any commit, they are often not intuitive IMO):

```
# Add upstream (only needed once)
git remote add upstream git@github.com:mandiant/VM-Packages.git

# Fetch upstream to get latest changes/branches
git fetch upstream

# Rebase to main
git rebase upstream main
```

Please let me know if you any help or tips with `git`, I may be able to help. :wink:

Ana06:
@PrajeetGuha do you know when you will have time to address the last minor changes, ensure there is a single commit, and do a final test? This PR and https://github.com/mandiant/VM-Packages/pull/1365 modify the same code and there will be merging conflicts. Because this PR is older, I would like to merge it first if you are able to make the changes soon.

PrajeetGuha:
Ya I will do it by a day.

packages/pestudio.vm/pestudio.vm.nuspec
```diff
@@ -2,7 +2,7 @@
 <package xmlns="http://schemas.microsoft.com/packaging/2015/06/nuspec.xsd">
   <metadata>
     <id>pestudio.vm</id>
-    <version>9.61.20250219</version>
+    <version>0.0.0.20250219</version>
```
Ana06, line 5: @PrajeetGuha can you remove this change and create a separate PR? We can merge this PR without the CI to test that mftecmd is updated. But this means that the packages won't be pushed to MyGet. It is ok for mftecmd as the automatic update will push it. But not for pestudio.

Sorry, I think I said before we could leave the change.

Ana06 (changes requested):
@PrajeetGuha we need to fix mftecmd in order to merge the PR (so that the CI passes), otherwise the chan

scripts/utils/update_package.py
PrajeetGuha: I am removing it. 👍

PrajeetGuha:
image
@Ana06 Good to merge. Test passed. 😄

Ana06 (approved):
Thanks for all the work @PrajeetGuha! This is a major enhancement to our CI that will help fixing broken tools quicker! :sparkles:

PrajeetGuha:
@Ana06 do check https://github.com/mandiant/VM-Packages/issues/1270
for some future enhancements that need to be worked on.

PR #1352, opened 2025-04-14
Fix the missing registry key check at googlechrome.vm

### LINKED ISSUE #1257: Error in Google Chrome installation
I get the following error in the Google Chrome installation (`choco install googlechrome.vm`):

```
Cannot find path 'HKLM:\SOFTWARE\Policies\Microsoft\Windows\Explorer' because it does not exist.
```

I think this is coming from this code:
```
# Do not show the "Open with" popup
Set-ItemProperty -path "HKLM:\SOFTWARE\Policies\Microsoft\Windows\Explorer" -name "NoNewAppAlert" -value 1 -type "DWord"
```

So, either there is a bug and the code is not working or the code is not needed. @emtuls do you have more information?

@d35ha I noticed this when testing https://github.com/mandiant/VM-Packages/pull/1255, did you get the error too? do you know what's going on? 😕

d35ha:
Fixes #1257.

### DIFF
--- a/packages/googlechrome.vm/googlechrome.vm.nuspec
+++ b/packages/googlechrome.vm/googlechrome.vm.nuspec
@@ -2,7 +2,7 @@
 <package xmlns="http://schemas.microsoft.com/packaging/2015/06/nuspec.xsd">
   <metadata>
     <id>googlechrome.vm</id>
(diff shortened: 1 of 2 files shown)


Ana06 (approved):
Thanks @d35ha! :rocket:

PR #1354, opened 2025-04-14
Update the version of pefile

### LINKED ISSUE #1286: Update pefile
### Details

It seems we are installing an old version of the pefile Python library, why? @binjo would like that we update it to use: https://github.com/williballenthin/ucutils/blob/master/pyproject.toml#L60

d35ha:
Fixes #1286

### DIFF
--- a/packages/internet_detector.vm/tools/chocolateyinstall.ps1
+++ b/packages/internet_detector.vm/tools/chocolateyinstall.ps1
@@ -19,10 +19,6 @@ $toolDir = Join-Path ${Env:RAW_TOOLS_DIR} $toolName
 New-Item -Path $toolDir -ItemType Directory -Force -ea 0
 VM-Assert-Path $toolDir
 
-# Install pyinstaller 6.11.1 (needed to build the Python executable with a version capable of executing in admin cmd) and tool dependencies ('pywin32')
-$dependencies = "pyinstaller==6.11.1,pywin32==308,icmplib==3.0.4"
-VM-Pip-Install $dependencies
-
 # Set the ICMP ID at the tool script to a randomized value per installation
 # to make it harder for a malware to hardcode it and to lower/zero the
 # probability of overlapping with any other application sending ping requests.
--- a/packages/libraries.python3.vm/tools/modules.xml
+++ b/packages/libraries.python3.vm/tools/modules.xml
@@ -15,7 +15,7 @@
     <module name="msoffcrypto-tool"/>
     <module name="olefile"/>
     <module name="oletools"/>
-    <module name="pefile"/>
+    <module name="pefile>=2024.08.26"/>
     <module name="psutil"/>
     <module name="pyasn1"/>
     <module name="pycryptodome"/>
--- a/packages/libraries.python3.vm/tools/modules.xml
+++ b/packages/libraries.python3.vm/tools/modules.xml
@@ -34,4 +34,6 @@
     <module name="yara-python"/>
     <module name="frida"/>
     <module name="frida-tools"/>
+    <module name="pyinstaller==6.10.0"/>
+    <module name="pywin32"/>
 </modules>
--- a/packages/internet_detector.vm/internet_detector.vm.nuspec
+++ b/packages/internet_detector.vm/internet_detector.vm.nuspec
@@ -2,12 +2,12 @@
 <package xmlns="http://schemas.microsoft.com/packaging/2015/06/nuspec.xsd">
   <metadata>
     <id>internet_detector.vm</id>
-    <version>1.0.0.20250424</version>
+    <version>1.0.0.20250425</version>
     <authors>Elliot Chernofsky and Ana Martinez Gomez</authors>
     <description>Tool that changes the background and a taskbar icon if it detects internet connectivity</description>
     <dependencies>
       <dependency id="common.vm" version="0.0.0.20250206" />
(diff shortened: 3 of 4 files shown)


Ana06 (changes requested):
@d35ha I think this issue is a bit more difficult to fix. I think the reason why we have an older version of pefile installed is because one package/tool is requiring it. I am not sure which one, but if I test the following packages together, I get a pip dependency error with the proposed code:
image
```
[+] Attempting to install Python3 module: pefile>=2024.08.26
ERROR: pip's dependency resolver does not currently take into account all the packages that are installed. This behaviour is the source of the following dependency conflicts.
At line:1 char:1
+ py -3.10 -W ignore -m pip install pefile>=2024.08.26 --disable-pip-ve ...
+ ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
ERROR: pip's dependency resolver does not currently take into account all the packages that are installed. This behaviour is the source of the following dependency conflicts.
At line:1 char:1
+ py -3.10 -W ignore -m pip install pefile>=2024.08.26 --disable-pip-ve ...
+ ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
ERROR: ERROR: pip's dependency resolver does not currently take into account all the packages that are installed. This behaviour is the source of the following dependency conflicts.
The upgrade of libraries.python3.vm was NOT successful.

**Comment by d35ha** [comment]
> @d35ha I think this issue is a bit more difficult to fix. I think the reason why we have an older version of pefile installed is because one package/tool is requiring it. I am not sure which one, but if I test the following packages together, I get a pip dependency error with the proposed code: image
> 
> ```
> [+] Attempting to install Python3 module: pefile>=2024.08.26
> ERROR: pip's dependency resolver does not currently take into account all the packages that are installed. This behaviour is the source of the following dependency conflicts.
> At line:1 char:1
> + py -3.10 -W ignore -m pip install pefile>=2024.08.26 --disable-pip-ve ...
> + ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
> ERROR: pip's dependency resolver does not currently take into account all the packages that are installed. This behaviour is the source of the following dependency conflicts.
> At line:1 char:1
> + py -3.10 -W ignore -m pip install pefile>=2024.08.26 --disable-pip-ve ...
> + ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
> ERROR: ERROR: pip's dependency resolver does not currently take into account all the packages that are installed. This behaviour is the source of the following dependency conflicts.
> The upgrade of libraries.python3.vm was NOT successful.
> ```

@Ana06  

There's a conflict caused by pyinstaller as [pyinstaller 6.11.1 requires `pefile != 2024.8.26`](https://pyinstaller.org/en/v6.11.1/CHANGES.html#pyinstaller-core) which is the exact version we are trying to install.

Screenshot From 2025-04-24 16-51-47

**Comment by Ana06** [comment]
@d35ha is pyinstaller needed for the internet detector? if so, then we would need to either find an alternative in the internet detector or try to update this dependency in pyinstaller (by sending an issue/PR to that project asking for the update). :thinking:

File: packages/internet_detector.vm/tools/chocolateyinstall.ps1 [hunk 1]
```diff
@@ -10,8 +10,8 @@ $toolDir = Join-Path ${Env:RAW_TOOLS_DIR} $toolName
 New-Item -Path $toolDir -ItemType Directory -Force -ea 0
 VM-Assert-Path $toolDir
 
-# Install pyinstaller 6.11.1 (needed to build the Python executable with a version capable of executing in admin cmd) and tool dependencies ('pywin32')
-$dependencies = "pyinstaller==6.11.1,pywin32"
+# Install pyinstaller 6.10.0 (needed to build the Python executable with a version capable of executing in admin cmd) and tool dependencies ('pywin32')
+$dependencies = "pyinstaller==6.10.0,pywin32"
```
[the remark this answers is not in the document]
d35ha: It's mentioned [here](https://pyinstaller.org/en/v6.11.1/CHANGES.html#pyinstaller-core) that it's due to performance reasons.

d35ha:
@Ana06 
I moved the python dependencies to `libraries.python3.vm` since it's already required by `internet_detector.vm`.

packages/libraries.python3.vm/tools/modules.xml
```diff
@@ -34,4 +34,6 @@
     <module name="yara-python"/>
     <module name="frida"/>
     <module name="frida-tools"/>
+    <module name="pyinstaller==6.10.0"/>
+    <module name="pywin32"/>
```
Ana06, line 38: I think we should keep these two libraries in the internet detector. If the user doesn't install the internet detector, likely does not want these libraries neither, as they are not very useful for malware analysis on their own.

d35ha: It's actually useful; `pyinstaller` comes with `pyi-archive_viewer` which I personally use to unpack `pyinstaller` packaed scripts and `pywin32` provides access to the Windows API from python.

Ana06 (approved):
:rocket:

PR #1355, opened 2025-04-14
Fix links in the metadata package template

sara-rn:
Replace old wiki links with fixed pages

### DIFF
--- a/.github/ISSUE_TEMPLATE/new_metapackage.yml
+++ b/.github/ISSUE_TEMPLATE/new_metapackage.yml
@@ -40,7 +40,7 @@ body:
     attributes:
       label: Version
       description: |
-        Tool's version number. Example: `1.4.11`. See the [_Metapackages Version_ documentation in our wiki](https://github.com/mandiant/VM-Packages/wiki/Package-Structure#metapackages) for more details.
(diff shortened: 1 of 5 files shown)


.github/ISSUE_TEMPLATE/new_metapackage.yml
```diff
@@ -132,4 +132,4 @@ body:
...
+        We really appreciate that you share your package idea with us and we will review your issue as soon as possible. Please understand that our time is limited and we may not be able to create a package for this tool as fast as you may like. To speed up the process, please select the correct issue template and fill the issue accurately as this may allow our bot to send an automatic PR. Note that this may not work for all cases. You can check our [Package walk through Wiki page](https://github.com/mandiant/VM-Packages/wiki/My-first-package-walk-through) and send a pull request with the new package. We encourage the community to help us create new packages and to maintain existing ones. Join us! :hugs:
```
Ana06, line 135: The Contributing Wiki page was moved to https://github.com/mandiant/VM-Packages/blob/main/CONTRIBUTING.md to use GitHub features and follow Google policies. I think we should link https://github.com/mandiant/VM-Packages/blob/main/CONTRIBUTING.md instead of the Package walk through Wiki page here.

This link is also broken in all other templates, can you please fix those links too?

Ana06 (changes requested):
Good catch @sara-rn, thanks for fixing the links!

PR #1360, opened 2025-04-16
Support installing custom locales

### LINKED ISSUE #1358: Install several locale
### Details

When analysing malware, it is common that it behaves differently depending on the locale. So it would be useful to install several one to be able to easily change it (without needed to enable internet to do it while analysing the malware.

I propose installing the following ones:
- Chinese
- English (United Kingdom)
- German (Germany)
- Russian 
- Spanish (Spain)
- Ukrainian

@mandiant/flare-vm any other language you would like to have installed?

@mandiant/commando-vm is this something Commando VM would benefit from too? If so, what about adding it to the deblot or installer packages? Otherwise, we need a new package for it.

Also, does someone know how we can implement this using PowerShell scripting? Otherwise we will need to do some research. 😉

d35ha:
Fixes #1358

### DIFF
--- a/packages/common.vm/common.vm.nuspec
+++ b/packages/common.vm/common.vm.nuspec
@@ -2,7 +2,7 @@
 <package xmlns="http://schemas.microsoft.com/packaging/2015/06/nuspec.xsd">
   <metadata>
     <id>common.vm</id>
-    <version>0.0.0.20250407</version>
+    <version>0.0.0.20250416</version>
     <description>Common libraries for VM-packages</description>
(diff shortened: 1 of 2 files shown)


Ana06 (commented):
Smart way to address https://github.com/mandiant/VM-Packages/issues/1358, which allows @mandiant/commando-vm to also benefit from the enhancement, while keeping it configurable. :+1:

Ana06 (approved):
Thanks @d35ha! :bouquet:

Issue #1314, opened 2025-02-25
Add Suricata package for installation into the FLARE-VM


jstrosch:
### Details

I'd like to propose the inclusion of Suricata, primariy for the FLARE-VM. Suricata can run in offline mode, allowing for the consumption of PCAP to generate alerts or other network data (i.e. JA3/JA4,etc). This can be used in conjunction with tools like Fakenet, allowing for offline network generation and analysis. 

Suricata provides an MSI installer: https://www.openinfosecfoundation.org/download/windows/Suricata-7.0.8-1-64bit.msi.

Rule sets will need to be added to the default rule path, which will be under C:\ProgramFiles\Suricata\rules\. There are a few options for open-source/open license rule sets. ET Open is the largest and most popular: https://rules.emergingthreats.net/open/suricata-7.0.3/. AttackDetection is another: https://github.com/ptresearch/AttackDetection. And there are several options from AbuseCH: https://sslbl.abuse.ch/blacklist/. 

To get an idea of how Suricata could be used in the FLARE-VM, I have a short video series on YouTube: https://www.youtube.com/playlist?list=PLHJns8WZXCdthheGdEpV4D_NdKTwcd_Xg. This would not only allow for increased detection opportunities, but the foundation of a custom rule writing workflow for those interested in creating their own rules.  I hope this request found the right location and appreciate the consideration. Please let me know if you have any questions.

Thanks,
Josh

PrajeetGuha:
Great idea to include such tools.

jstrosch:
yes, similar to how we use Yara and its rules to help detect capabilities, families, properties of different static artifacts, Suricata can apply rules to match in network traffic. The primary goal would be that those detections help an analyst more quickly identify malware family/capabilities, or add additional confidence to the static/dynamic analysis they are already doing. It does not need to run as a service (and therefore not become a resource concern), offline mode can be invoked on-demand and even a large rule set is relatively small in size (i.e. what it would consume on disk). Most major sandboxes include network detections (and many use Suricata), so I think it is a long overdue capability to have in the flare-vm :)

Ana06:
Thanks for all the details @jstrosch! I have no experience with Suricata, but @jstrosch is an experienced reverse engineer and if he finds the tool useful, I see no issue in adding a package in VM-Packages for it. If the increase in size is not significant (not sure if the rules are too big), we should also be able to add this package to FLARE-VM default configuration.

PrajeetGuha:
[...]

Issue #1231, opened 2025-01-09
Improve the Packages Wiki page


Ana06:
### Details

Improve the _Packages_ Wiki page introduced in https://github.com/mandiant/VM-Packages/pull/1197, after https://github.com/mandiant/VM-Packages/issues/1107 is **completely** addressed. Ideas:

- **Improve package descriptions**. The package descriptions were not used before creating this page. Now that they are displayed to the users, we should ensure they are consistent, informative, short and not include version specific details. I suggest we start the description with the tool name (so that is clear which tool the package install) followed by a short description. We could even use markdown in the description to highlight the tool name, for example: `**IDA Free** is the free version of IDA Pro, a powerful interactive disassembler and debugger.`. This implies modifying most nuspecs, so I recommend implementing this in an independent PR. 
- **Add package link**. Make the name of the package clickable so that it takes to the package folder in this repository (or maybe the nuspec?).
- **Add tool link**. Add [`projectUrl`](https://learn.microsoft.com/en-us/nuget/reference/nuspec#projecturl) to the nuspec (only for the packages where there is a clear useful project URL) and link it in the wiki page (for example using the tool name in the description). This implies modifying most nuspecs, so I recommend implementing this in an independent PR.
- **Display packages in the FLARE-VM default configuration**. This could be done by adding a column or using emojis after the package name. I think we should brainstrom how this exactly should look like in this issue before sending a PR.
- **Add license column**. I think we should not display the license if it is unclear or likely to change, but we could do it for FLARE projects and for well established projects. We could implement it by adding a SPDX license expression in the [`license` nuspec field](https://learn.microsoft.com/en-us/nuget/reference/nuspec#license) and adding a new column to display it. We may want to discuss what exact package we want to display it for and how useful this would be before implementing it. This implies modifying most nuspecs, so I recommend implementing this in an independent PR.
- At the moment, we regenerate the wiki page daily. After the page is finished (the rest improvements in this issue are implemented), we should consider **updating the wiki page on merge** if the workflow runs fast enough.

IMPORTANT: Any new convention in the nuspec needs to be documented in https://github.com/mandiant/VM-Packages/wiki/Chocolatey-packages#nuspec. 

@mandiant/vms opinions? Any other ideas?

sara-rn:
@Ana06 I think the simplest way to add the `projectUrl `link in the description field in the Wiki is to add a hyperlink with name `link` that points to the `projectUrl` at the end of the description.

[...]

PR #1375, opened 2025-04-23
Suricate Package Addition

PrajeetGuha:
related to https://github.com/mandiant/VM-Packages/issues/1314


Ana06 (commented):
@PrajeetGuha as we need this change for Suricata, I think we should add the package in the same PR (if you wish in a different commit) so that we can test them together.

You need to rebase to solve the conflict with the package version.

Ana06 (commented):
@PrajeetGuha I see some activity in the PR, but I do not see the new package. @PrajeetGuha have you maybe forgotten to push the changes? :thinking: Let me know if you need help with anything.

PrajeetGuha:
I have worked partly (not complete).
Will work more on this areas -
- Default shortcut in Desktop needs to be removed
- Temporary rules file download need to be clean
- msi file is getting stored in C:\Tools\Suricata directory, need to change that

For the package to work, I had to install `powershell-yaml` module from Powershell Gallery for reading and editing the `suricata.yaml` of the tool. Can you provide me with details regarding where to put such installation code in the project?
```
Install-Module -Name powershell-yaml
```

PrajeetGuha:
@Ana06 I am working on it but will need some more time to solve the problems. 🫡

PrajeetGuha:
image
Just a heads up! While reading the documentation, I came across this part where category selection is mentioned with a variable which we used to use previously. Might need changes to it. 🫡

PrajeetGuha:
image
with the common.psm change, suricata as an app is identified but due to absence of installer cache folder, icon is not found.

packages/suricata.vm/tools/chocolateyinstall.ps1
```diff
@@ -0,0 +1,114 @@
...
+    silentArgs = $silentArgs
+    executablePath = $executablePath
+    url = $url
+    sha256 = $sha256
+    consoleApp = $true
+}
+
+# download msi file
+VM-Install-With-Installer @packageArgs
+VM-Assert-Path $executablePath
```
Ana06, line 28: `VM-Install-With-Installer` already executes `VM-Assert-Path $executablePath` so this is not needed.
```suggestion
```

packages/suricata.vm/tools/chocolateyinstall.ps1
```diff
@@ -0,0 +1,114 @@
...
+    category = $category
+    filetype = $filetype
+    silentArgs = $silentArgs
+    executablePath = $executablePath
+    url = $url
+    sha256 = $sha256
+    consoleApp = $true
+}
+
+# download msi file
```
Ana06, line 26: I find this comment misleading,  as `VM-Install-With-Installer` does more than that
```suggestion
```

packages/suricata.vm/tools/chocolateyinstall.ps1
```diff
@@ -0,0 +1,114 @@
+$ErrorActionPreference = 'Stop'
+Import-Module vm.common -Force -DisableNameChecking
+Import-Module powershell-yaml
+
+# set configurations
+$toolName = "Suricata"
+$category = VM-Get-Category($MyInvocation.MyCommand.Definition)
+$filetype = "MSI"
```
Ana06, line 8: Use the `MSI` directly, since it is a short string and it is only used once. This is consistent with other packages.
```suggestion
```

packages/suricata.vm/tools/chocolateyinstall.ps1
```diff
@@ -0,0 +1,114 @@
...
+$toolDir = Join-Path ${Env:ProgramFiles} $toolName
+$executablePath = Join-Path $toolDir "suricata.exe"
+$url = "https://www.openinfosecfoundation.org/download/windows/Suricata-7.0.8-1-64bit.msi"
```
Ana06, line 11: we call this `exeUrl` in other packages like [bindiff.vm](https://github.com/mandiant/VM-Packages/blob/af930c6eb0e13cffcca8c037e3469570785a4f80/packages/bindiff.vm/tools/chocolateyinstall.ps1#L7):
```suggestion
$exeUrl = "https://www.openinfosecfoundation.org/download/windows/Suricata-7.0.8-1-64bit.msi"
```

packages/suricata.vm/tools/chocolateyinstall.ps1
```diff
@@ -0,0 +1,114 @@
...
+$sha256 = "8bdd78d2978e4efc6d23ab1ced024342cac7d38afb152a6e3f70ac5182bd8cd4"
+$silentArgs = "/qn /norestart"
```
Ana06, line 13: Use the silent args directly, since it is a short string and it is only used once. This is consistent with other packages.
```suggestion
```

packages/suricata.vm/tools/chocolateyinstall.ps1
```diff
@@ -0,0 +1,114 @@
+$ErrorActionPreference = 'Stop'
+Import-Module vm.common -Force -DisableNameChecking
+Import-Module powershell-yaml
+
+# set configurations
+$toolName = "Suricata"
```
Ana06, line 6: According to [our documentation](https://github.com/mandiant/VM-Packages/wiki/Coding-Conventions#toolname), the tool name is the file name with the `.exe` which in this case seems to be `suricata.exe`
```suggestion
$toolName = "suricata"
```

packages/suricata.vm/tools/chocolateyinstall.ps1
```diff
@@ -0,0 +1,114 @@
...
+$category = VM-Get-Category($MyInvocation.MyCommand.Definition)
+$filetype = "MSI"
+$toolDir = Join-Path ${Env:ProgramFiles} $toolName
+$executablePath = Join-Path $toolDir "suricata.exe"
```
Ana06, line 10: [nitpick] 
```suggestion
$executablePath = Join-Path $toolDir "$toolName.exe"
```

packages/suricata.vm/tools/chocolateyinstall.ps1
```diff
@@ -0,0 +1,114 @@
...
+$rulesDir = Join-Path $toolDir "rules" -Resolve
+$rulesConfigPath = Join-Path $toolDir "suricata.yaml" -Resolve
+
+$rulesConfig = ConvertFrom-Yaml (Get-Content -Raw -Path $rulesConfigPath)
+
+$failures = @()
+$rules = $rulesXml.rules.rule
+
+$tempToolDir = Join-Path ${Env:TEMP} $toolName
+$tempToolDir += ".vm"
```
Ana06, line 52:
```suggestion
$tempToolDir = Join-Path ${Env:TEMP} "$toolName.vm"
```

packages/suricata.vm/tools/chocolateyinstall.ps1
```diff
@@ -0,0 +1,114 @@
...
+            throw "`t[!] Unsupported file type: '$filePath'. Only .zip and .rule are allowed."
+        }
+    } catch {
+        $failures += $rule.name
+    }
+}
+
+$allRuleFiles = Get-ChildItem -Path $tempRuleDir -Recurse -File -Filter *.rules
+
+# move all rule files in temp rule folder to the suricata rule folder
```
Ana06, line 96: why don't we store the rules directly in the suricata rule folder?

packages/suricata.vm/tools/chocolateyinstall.ps1
```diff
@@ -0,0 +1,114 @@
...
+}
+
+$rulesConfig | ConvertTo-Yaml | Set-Content -Path $rulesConfigPath
+
+# display all errors
+if ($failures.Count -gt 0) {
+    foreach ($module in $failures) {
+        VM-Write-Log "ERROR" "Failed to install rule: $($rule.name)"
+    }
+    exit 1
```
Ana06, line 113: [question] do we want to fail the installation of suricata if any rule fails to install? In some packages, we allow some configuration to fail and just log the error. But if we expect all the rules installation to work, then we should fail the package to ensure we detect unexpected issues.

@jstrosch I am not very familiar with suricata. Can you help here? :smile:

packages/suricata.vm/tools/chocolateyuninstall.ps1
```diff
@@ -0,0 +1,7 @@
+$ErrorActionPreference = 'Continue'
+Import-Module vm.common -Force -DisableNameChecking
+
+$toolName = 'Suricata'
+$category = VM-Get-Category($MyInvocation.MyCommand.Definition)
+
+VM-Uninstall $toolName $category
```
Ana06, line 7: I think this fails and we need `VM-Uninstall-With-Uninstaller` instead.

packages/suricata.vm/tools/chocolateyinstall.ps1
```diff
@@ -0,0 +1,114 @@
+$ErrorActionPreference = 'Stop'
+Import-Module vm.common -Force -DisableNameChecking
+Import-Module powershell-yaml
```
Ana06, line 3: This fails for me locally too with the same error as in the CI. We need to find an alternative that works in GitHub Actions and for the [FLARE-VM requirements](https://github.com/mandiant/flare-vm?tab=readme-ov-file#requirements).

Idea: do we really need `powershell-yaml` to parse `suricata.yaml`? it seems to me that we just need to add the the list of rules after the line `rule-files:` in this file. We could do a simple replace for that. Or am I missing something?

Ana06:
@PrajeetGuha 

> Just a heads up! While reading the documentation, I came across this part where category selection is mentioned with a variable which we used to use previously. Might need changes to it. 🫡

I have just updated the wiki, thanks for bringing it up!

packages/suricata.vm/tools/chocolateyinstall.ps1
```diff
@@ -0,0 +1,114 @@
...
+}
+
+# rules configuration and download
+$rulesXmlPath = "$(Split-Path -parent $MyInvocation.MyCommand.Definition)/rules.xml"
+$rulesXml = [xml](Get-Content $rulesXmlPath)
+
+$rulesDir = Join-Path $toolDir "rules" -Resolve
+$rulesConfigPath = Join-Path $toolDir "suricata.yaml" -Resolve
+
+$rulesConfig = ConvertFrom-Yaml (Get-Content -Raw -Path $rulesConfigPath)
```
Ana06, line 46: I would move the definition of these two variables just before they are used (before the code to add rules to `suricata.yaml`)

Ana06 (changes requested):
Thanks for all the work @PrajeetGuha! :bouquet: 

All code that could fail needs to be inside a `try-catch` with a `VM-Write-Log-Exception $_` to better handle exceptions. It is always good to have new people contributing to the project, as we notice where we are lucking documentation. I have just documented the `try-catch` in https://github.com/mandiant/VM-Packages/wiki/Coding-Conventions. :wink:

packages/suricata.vm/tools/chocolateyinstall.ps1
PrajeetGuha: There is a new version of msi for suricata and now I see by default they actually provide the rules that we intend to add.. 🤔

packages/suricata.vm/tools/chocolateyinstall.ps1
PrajeetGuha: my current implementation pattern is to collect all the rules file from all sources listed in the xml into the rules folder. Then at the end run a loop that finds all files ending with .rules extension in the temp rule folder, add them to the suricata default rule folder and add the entry in the yaml configuration.
```
loop(for entries in rules.xml):
   add rules file entry to temp rule folder

loop(for rule-file in temp rule folder):
   add rule-file to suricata rule folder
   add rule-file name to suricata config file
```
while adding the rules in the rule folder, we also need the names of the files added in order to add to the config.

packages/suricata.vm/tools/chocolateyinstall.ps1
PrajeetGuha: suricata.msi has problems with default configuration where in it's rules-file part of the config file it enlists rules that are non-existent in the rules folder (it has reference to emerging-all rules in the yaml but no rule file present in the rules folder). The implementation using powershell-yaml will prevent the config file to become messy by preventing duplicate entries in this section.

packages/suricata.vm/tools/chocolateyuninstall.ps1
PrajeetGuha: thanks!! also to note, suricata unistallation does not remove the suricata/rules folder in ProgramFiles if it has custom rules added other than default rules. If we want the uninstallation to completely delete all related resources of it then we may have to add additional piece of code to remove them.

PrajeetGuha:
> Thanks for all the work @PrajeetGuha! 💐
> 
> All code that could fail needs to be inside a `try-catch` with a `VM-Write-Log-Exception $_` to better handle exceptions. It is always good to have new people contributing to the project, as we notice where we are lucking documentation. I have just documented the `try-catch` in https://github.com/mandiant/VM-Packages/wiki/Coding-Conventions. 😉

Thanks!! Will check and add exceptions wherever necessary after the code is working as expected and is approved.

packages/suricata.vm/tools/chocolateyinstall.ps1
Ana06: I am ok using anything you wish to parse the yaml, as long as it can be installed in the CI and for the [FLARE-VM requirements](https://github.com/mandiant/flare-vm?tab=readme-ov-file#requirements). The current implementation fails in the CI and for me locally, so it is not an option.

It should also be easy to ensure there are no duplicates entries parsing the file (even it is a bit more elaborate than with `powershell-yaml`)

Ana06:
@PrajeetGuha 

> Will check and add exceptions wherever necessary after the code is working as expected and is approved. 

This should be easy to address. Normally all packages have just a single `try-catch` with most of the code inside :wink: Let me know if what I mean is not clear.

packages/suricata.vm/tools/chocolateyuninstall.ps1
Ana06: As [documented in the wiki](https://github.com/mandiant/VM-Packages/wiki/Coding-Conventions#uninstalling--updating-are-best-effort), uninstalling is best effort. The FLARE teams always build a VM from scratch and we do not uninstall packages, so we try to make the uninstallation as good as possible, while keeping the amount of work reasonable. So, if improving the uninstallation is not a lot of work, It would be good to do it. :wink:

packages/suricata.vm/tools/chocolateyinstall.ps1
Ana06: why can't this be done directly in the suricata rule folder? Do we expect other rules to already exist and we don't need to parse them? If so, adding a clarifying comment in the code would be good.

packages/suricata.vm/suricata.vm.nuspec
```diff
@@ -0,0 +1,14 @@
+<?xml version="1.0" encoding="utf-8"?>
+<package xmlns="http://schemas.microsoft.com/packaging/2015/06/nuspec.xsd">
+  <metadata>
+    <id>suricata.vm</id>
+    <version>7.0.10</version>
+    <authors>Open Information Security Foundation</authors>
+    <description>Suricata is a network IDS, IPS and NSM engine developed by the OISF and the Suricata community</description>
```
Ana06, line 7: New description convention as documented in the [wiki](https://github.com/mandiant/VM-Packages/wiki/Chocolatey-packages#description):
> The description should start with the tool name and end with a `.`.

We are improving the description as part of https://github.com/mandiant/VM-Packages/issues/1231, also motivated by the [FLARE-VM installer enhancements (the FLARE-VM installer now show the descriptions)](https://github.com/mandiant/flare-vm/pull/672).
```suggestion
    <description>Suricata is a network IDS, IPS and NSM engine developed by the OISF and the Suricata community.</description>
```

Ana06 (commented):
I  have reviewed the code, but I haven't been able to test the changes locally, as it seems there are issues with the feed (hopefully they resolve on their own soon as I think FLARE-VM/Commando VM are also broken because of this):
image

packages/common.vm/tools/vm.common/vm.common.psm1
```diff
@@ -81,7 +81,7 @@ function VM-Write-Log {
     [CmdletBinding()]
     Param(
         [Parameter(Mandatory=$true, Position=0)]
-        [ValidateSet("INFO","WARN","ERROR")]
+        [ValidateSet("INFO","WARN","ERROR","FATAL")]
```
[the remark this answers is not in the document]
PrajeetGuha: https://github.com/mandiant/VM-Packages/blob/73997f2e8e5ab14024cba4bfd078ef1a69ae861f/packages/common.vm/tools/vm.common/vm.common.psm1#L115

has $level=FATAL but the ValidateSet does not have it. This change is not related to this package implementation though.

packages/suricata.vm/tools/chocolateyinstall.ps1
```diff
@@ -0,0 +1,114 @@
...
+    for ($i = 0; $i -lt $rulesConfig.Count; $i++) {
+        if ($rulesConfig[$i] -match '^rule-files:$') {
+            $ruleFilesIndex = $i
+            break
+        }
+    }
+    # If `rule-files:` was not found, throw an error
+    if ($null -eq $ruleFilesIndex) {
+        throw "Line with 'rule-files:' string not found in the config file."
+    }
```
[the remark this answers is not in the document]
PrajeetGuha: As per documentation, we expect to have this section already present in the current implementation of msi bundle. Suricata msibundle comes prepacked with some default rules which are referenced here. If in future they change the implementation, we can do the additional changes to implement that. Currently I recommend that to keep this error and fail the package if this section is not present.

packages/suricata.vm/tools/chocolateyinstall.ps1
```diff
@@ -0,0 +1,114 @@
...
+    $rulesList = $rulesConfig -split "`n" | Where-Object { $_.Trim() -match '\.rules$' } | ForEach-Object { $_.TrimStart(' -').Trim() }
+
+    # index of the location in the yaml where `rule-files:` is specified
+    $ruleFilesIndex = $null
+    for ($i = 0; $i -lt $rulesConfig.Count; $i++) {
+        if ($rulesConfig[$i] -match '^rule-files:$') {
+            $ruleFilesIndex = $i
+            break
+        }
+    }
```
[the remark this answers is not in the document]
PrajeetGuha: thanks for the recommendation!

Ana06:
Google CLA doesn't like my personal email as co-author. I think using `anamg@google.com` should be ok. I am also fine removing the co-author.

PrajeetGuha:
> Google CLA test doesn't like my personal email as co-author. I think using `anamg@google.com` should be ok. I am also fine removing the co-author. 
> 
> @PrajeetGuha did you add this manually or was it the GH interface? I would like to try to configure the email if it was the GH interface.

It is the GH interface. I think it will not be problem when I squash and rebase the code after final review. Let me know if any changes are required.

packages/suricata.vm/tools/chocolateyinstall.ps1
jstrosch: Yes, I think this is a good approach. Suricata can still produce a lot of useful data w/o rules (protocol logs, file extraction, etc). Of course, in the context of malware analysis the rules are highly desirable, but relatively easy to fix post-installation if needed.

packages/common.vm/tools/vm.common/vm.common.psm1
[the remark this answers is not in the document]
Ana06: [nitpick] It is a good practice to not commit unrelated changes in the same commit. But I see why this is an enhancement now. :wink:

Ana06:
@PrajeetGuha thanks for the answer, I had misunderstood that previous commit. @emtuls any ideas to find the icon? or should we merge it as it is (it needs a rebase)?

PR #1377, opened 2025-04-24
Fix the uninstallation of internet_detector.vm

d35ha:


packages/internet_detector.vm/tools/chocolateyuninstall.ps1
```diff
@@ -8,7 +8,12 @@ Disable-ScheduledTask -TaskName 'Internet Detector' -ea 0 | ForEach-Object {
     $_ | Stop-ScheduledTask
     $_ | Unregister-ScheduledTask -Confirm:$false
 }
-Stop-Process -Name "$toolName.exe" -Force -ea 0
+
+Get-Process -Name $toolName -ea 0 | ForEach-Object {
```
Ana06, line 12: If the process is not running, I think this will fail.

d35ha: @Ana06 

`-ea 0` will silently continue it.

Ana06 (approved):

Issue #1219, opened 2024-12-16
Re-try package installation on failure


Ana06:
### Details

As @vsl-iil mentioned in https://github.com/mandiant/flare-vm/issues/577, at the moment our installer tries to reinstall failed packages when a reboot occurs. This is because how the installer is implemented and it is annoying for packages that are broken. I am not sure if there is a way to avoid packages to get reinstalled, @mandiant/vms any ideas?

It is useful to try to reinstall packages that failed to install for other reasons, such as a punctual server issue (which I have had with IDA Free recently). I would like to add an option to try to reinstall failed packages so that they have more changes to get installed in automated FLARE-VM installations (without manual interaction). An idea would be to try to install every package twice as we do in the test suite. Another idea would be to do it at the end of the installation. This would mean more time happens between tries, giving server failures more time to get fixed. But I am not sure if this could cause an infinite loop if there is a reboot in the middle.

@mandiant/vms any opinions or ideas of how to best implement package re-installation on failure? As this modifies the installer, it affects Commando VM as well. :thinking: 

@l33th4xor1 you mentioned you install FLARE-VM without manual interaction. Do you have any ideas here? :smile:

Ana06:
Creating another issue for the re-try on a reboot: https://github.com/mandiant/VM-Packages/issues/1378. This issue tracks only an extra single re-try at the end of the installation.

PR #1379, opened 2025-04-24
Re-try package installation on failure

Ana06:
Try to install failed packages after completing all packages installation. This way packages have more chances to get installed,
facilitating automated FLARE-VM installations (without manual interaction).

Introduces the functions `Get-Failed-Packages` and `Install-Packages` to avoid duplicating code.

In addition, this PR also fixes a bug introduced in PR https://github.com/mandiant/VM-Packages/pull/1274 where we started to always create `failed_packages.txt` to signal installation completion for automation. It also inadvertently caused the installer to always log additional failure information, regardless of actual package failures, because of checking for the existence of the (now always present) file. The detailed failure messages are now only displayed when one or more packages have failed to install.

Closes https://github.com/mandiant/VM-Packages/issues/1219.

@mandiant/vms I would appreciate a review if anyone has time, as any error in the installer breaks Commando-VM and FLARE-VM. I have tested locally, but an extra pair of :eyes: would be good.

### DIFF
--- a/packages/installer.vm/installer.vm.nuspec
+++ b/packages/installer.vm/installer.vm.nuspec
@@ -2,7 +2,7 @@
 <package xmlns="http://schemas.microsoft.com/packaging/2015/06/nuspec.xsd">
   <metadata>
     <id>installer.vm</id>
-    <version>0.0.0.20250207</version>
+    <version>0.0.0.20250424</version>
     <authors>Mandiant</authors>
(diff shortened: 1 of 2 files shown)


d35ha (approved):
I tested it successfully, LGTM.

Ana06:
Thanks for the review @d35ha! :bouquet:

PR #1384, opened 2025-04-25
[garbageman] Fix dependency

Ana06:
dotnet5-desktop-runtime community package is not available anymore, breaking the package. Replace it with dotnet-5.0-desktopruntime.

I have tested locally and the tool installs and starts correctly. But I am not sure if it works as expected. Does someone know a good way to test it? :confused: 

Closes https://github.com/mandiant/VM-Packages/issues/1383

PR #1385, opened 2025-04-25
Use dotnet-5.0-desktopruntime instead of dotnet5-desktop-runtime

### LINKED ISSUE #1383: garbageman.vm is broken because of dotnet5-desktop-runtime
### Details

garbageman.vm fails to install because the dotnet5-desktop-runtime community package is not available anymore. [Garbageman's README](https://github.com/WithSecureLabs/GarbageMan) says:
> you need to install .NET 5.0 desktop runtime for amd64 and x86

We need to find a replacement for dotnet5-desktop-runtime in order to keep the package. If it really requires .NET 5.0 desktop runtime, are there other ways to install it without the community package?

.NET 5.0 is quite old and it seems that garbageman.vm is not maintained any more, is this still a useful tool?

@mandiant/vms @mike-hunhoff @naacbin any ideas?

d35ha:
Fixes #1383.

### DIFF
--- a/packages/garbageman.vm/garbageman.vm.nuspec
+++ b/packages/garbageman.vm/garbageman.vm.nuspec
@@ -2,12 +2,12 @@
 <package xmlns="http://schemas.microsoft.com/packaging/2015/06/nuspec.xsd">
   <metadata>
     <id>garbageman.vm</id>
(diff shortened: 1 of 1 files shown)


d35ha:
Closing it in favor of #1384.

Issue #1386, opened 2025-04-25
No module named icmplib


Ana06:
### Details

When installing FLARE-VM, at the end of the installation I see the following error message:

![Image](https://github.com/user-attachments/assets/9ad3a752-aec2-42f6-a4dd-9d1eca2609ec)

I do not see any related error in the logs, so I am not sure where the error is coming from. @mandiant/vms does someone have any idea?

If we are not sure where the error is coming from, one workaround is to add the `icmplib` python library to the libraries.python3.vm package. But it would be better to know where this is coming from as likely there is a package missing a dependency an we should address it in the package/tool that that is causing the issue.

Ana06:
@d35ha could it be related to the internet detector?

Ana06:
@d35ha icmplib used to be installed by the internet detector, but it was removed in https://github.com/mandiant/VM-Packages/pull/1354

d35ha:
> [@d35ha](https://github.com/d35ha) could it be related to the internet detector?
> 
> I have installed the library using `py -3.10 -m pip install icmplib` but the error keeps coming ~ every minute. The internet detector doesn't appear to be working.

The reason is that it should be existing at the installation time so pyinstaller ships it with the tool

Ana06:
It seems we tested https://github.com/mandiant/VM-Packages/pull/1354 having a previous version of libraries.python3.vm installed, which caused that icmplib was already installed. I'll send a PR to fix it!

Ana06:
During testing I get this deprecation warning, but as it says version 7, it should be the same for 6.10.0 and 6.11.1. So the downgrade shouldn't break anything. It may be that running it admin is not needed, but I am not sure and I don't want to break anything else. 😅 

![Image](https://github.com/user-attachments/assets/34e9b978-c43b-4ccd-91e6-c43fa9964cd4)

PR #1387, opened 2025-04-25
[internet_detector] Fix: Install dependency

Ana06:
The internet_detector needs icmplib, which was removed by accident in: https://github.com/mandiant/VM-Packages/issues/1386 Add icmplib to libraries.python3.vm to fix the issue. Document also which libraries are needed by the internet detector to prevent issues in the future.

Closes https://github.com/mandiant/VM-Packages/issues/1386


d35ha (approved):
LGTM

Issue #1230, opened 2025-01-08
IDA plugin proposal: ida.plugin.hexrayspytools.vm


Schamper:
### Package Name

ida.plugin.hexrayspytools

### Tool Name

HexRaysPyTools.py

### Package type

IDA_PLUGIN

### Version

0.0.0.20250108

### Tool's authors

Igor Kirillov, NyaMisty

### Tool's description

IDA Pro plugin which improves work with HexRays decompiler and helps in process of reconstruction structures and classes

### Download URL

https://github.com/NyaMisty/HexRaysPyTools/archive/cd3e429da7940a3c9638ee5d04787e3d2376a4ef.zip

### Download SHA256 Hash

44cc2eb1104d04110bc8425480729d8463cf7ea67d4e07c61b71dc1aa2631985

### Why is this tool a good addition?

Me and my colleagues find this plugin immensely useful for reverse engineering C++ and other binaries.

### Extra information

I selected a fork that includes some more fixes instead of the original repo, which has gone stale.

Ana06:
Thanks for the proposal @Schamper! This plugins has a dependency (specified in the `requirements` file), so we need to create it manually as our bot does not support adding dependencies.

Ana06:
It is common that IDA plugins and tools downloaded from a ZIP require to install the file `dependencies.txt`, as it happens in this package and for example also in xrefer: https://github.com/mandiant/VM-Packages/blob/9cb18b33df3f955a3a93e8d386dcaf893e86721d/packages/ida.plugin.xrefer.vm/tools/chocolateyinstall.ps1#L6

What about checking in the `VM-Install-IDA-Plugin` if that file exists and if so, install the dependencies with pip? This will ensure the plugin will continue to work on updates without making manual changes to update the dependencies and will allow us to create this package automatically (using the bot). We would be also able to create packages like this automatically.

I propose the following steps:
- One PR to extend `VM-Install-IDA-Plugin` to check if `dependencies.txt` exists and install the dependecies if so. Update also in this PR xrefer so that we can test the changes.
- Create this package using automation.

Schamper:
I think you mean `requirements.txt`?

[...]

PR #1390, opened 2025-04-29
Add option to install and uninstall specified files for IDA plugins

Schamper:
As discussed in https://github.com/mandiant/VM-Packages/issues/1230.

Adds the `-pluginFiles` argument to both `VM-Install-IDA-Plugin` and `VM-Uninstall-IDA-Plugin` to more finely manage the installed plugin files. This allows for more flexible plugin packages without having to make a custom `chocolateyInstall.ps1` for those.

Currently a bunch of plugins will fail to reinstall after an uninstall because of leftover files (a reinstall will complain about already existing files), so if this gets merged we probably want to update those plugins with this argument.

### DIFF
--- a/.github/ISSUE_TEMPLATE/new_IDA_plugin.yml
+++ b/.github/ISSUE_TEMPLATE/new_IDA_plugin.yml
@@ -32,7 +32,7 @@ body:
     attributes:
       label: Package type
       description: |
-        - **`IDA_PLUGIN`** - IDA plugin distributed as a single file or ZIP containing a plugin (and supporting files/directories) that need to be copied to the IDA plugins directory. For ZIPs, we check if there is an inner folder (this is the case for GH ZIPs) and if there is a directory called `plugins`. We copy all files in this directory (the root directory, the inner folder or the `plugins` directory depending what is present) with the exception of the README and the LICENSE file (often present in GH repos). The copied files must include the file `plugin name` specified below.
+        - **`IDA_PLUGIN`** - IDA plugin distributed as a single file or ZIP containing a plugin (and supporting files/directories) that need to be copied to the IDA plugins directory. For ZIPs, we check if there is an inner folder (this is the case for GH ZIPs) and if there is a directory called `plugins`. We copy all files in this directory (the root directory, the inner folder or the `plugins` directory depending what is present) with the exception of the README and the LICENSE file (often present in GH repos). The copied files must include the file `plugin name` specified below. If you need to be more specific in which files to copy, supply them in the `files` field below.
         - **`OTHER/UNKNOWN`** - An IDA plugin distributed in a different way than `IDA_PLUGIN`. It is required that you provide details of how the plugin is installed in the `Extra information` section. Note this type does not support automation, a PR would be appreciated!
       options:
         - IDA_PLUGIN
@@ -79,6 +79,12 @@ body:
       label: Download SHA256 Hash
       description: |
         SHA256 hash of the file or ZIP downloaded from the download url introduced in the previous field. The hash is used for verification purposes. Example: `62af5cce80dbbf5cdf961ec9515549334a2112056d4168fced75c892c24baa95`
+  - type: input
+    id: files
+    attributes:
+      label: Files to copy
+      description: |
+        A comma separated list of files and/or directories to copy into the plugin directory. Only necessary for plugins with awkward repository layouts. Example: `awkward_directory/awesomeplugin.py, `awkward_directory/awesomeplugin_libs`,
   - type: textarea
     id: why
     attributes:
--- a/scripts/utils/create_package_template.py
+++ b/scripts/utils/create_package_template.py
@@ -186,7 +186,7 @@ VM-Install-Single-Ps1 $toolName $category $ps1Url -ps1Sha256 $ps1Sha256
 
 """
 Needs the following format strings:
-    tool_name="...", target_url="...", target_hash="..."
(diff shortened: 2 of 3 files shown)


Ana06 (commented):
Thanks for the contribution @Schamper! :bouquet: can you please increase the version in the nuspec of common.vm.nuspec as explained in [our documentation](https://github.com/mandiant/VM-Packages/wiki/Chocolatey-packages#version)? We need increase the version, otherwise the existent version from MyGet is still used. The linter is complaining about it. :wink: 

I'll review the rest thoughtfully next week, I am sorry but I won't have time this week.

.github/ISSUE_TEMPLATE/new_IDA_plugin.yml
```diff
@@ -79,6 +79,12 @@ body:
       label: Download SHA256 Hash
       description: |
         SHA256 hash of the file or ZIP downloaded from the download url introduced in the previous field. The hash is used for verification purposes. Example: `62af5cce80dbbf5cdf961ec9515549334a2112056d4168fced75c892c24baa95`
+  - type: input
+    id: files
```
Ana06, line 83: @Schamper are you aiming for this field to be used by our automation? Have you tested the GH action [new_package.yml](https://github.com/mandiant/VM-Packages/blob/main/.github/workflows/new_package.yml#L22) with this change? I think it does not work, as we are using `template-path: .github/ISSUE_TEMPLATE/new_package.yml` to parse the issue and I think fields not included there are ignored.

.github/ISSUE_TEMPLATE/new_IDA_plugin.yml
```diff
@@ -79,6 +79,12 @@ body:
...
+    attributes:
+      label: Files to copy
+      description: |
+        A comma separated list of files and/or directories to copy into the plugin directory. Only necessary for plugins with awkward repository layouts. Example: `awkward_directory/awesomeplugin.py, `awkward_directory/awesomeplugin_libs`,
```
Ana06, line 87: [nitpick] I think it would be better to use an real example instead of  `awkward_directory/awesomeplugin.py`.

packages/ida.plugin.dereferencing.vm/tools/chocolateyinstall.ps1
```diff
@@ -4,5 +4,6 @@ Import-Module vm.common -Force -DisableNameChecking
 $pluginName = 'dereferencing.py'
 $pluginUrl = 'https://github.com/danigargu/deREferencing/archive/ab106c1fa5969a5d7c27c8037b4aec446ad55a4d.zip'
 $pluginSha256 = '83c0f9420075dbe5217f5fbf2413c7b1b72eadfaedfe2474b306e1e3eaae0aa8'
+$pluginFiles = @('dereferencing', 'dereferencing.py')
```
Ana06, line 7: I think this is a bad example to use the `$pluginFiles` argument as this plugin is installed correctly with the default behavior and this change unnecessarily complicates the code. What about removing this change and run `scripts/utils/create_package_template.py` to create the IDA plugin `ida.plugin.hexrayspytools.vm` so that we can test the new argument with an IDA plugin that needs it?

scripts/utils/create_package_template.py
```diff
@@ -249,7 +250,8 @@ def package_version(dependency_version):
 Import-Module vm.common -Force -DisableNameChecking
 
 $pluginName = '{tool_name}'
-VM-Uninstall-IDA-Plugin -pluginName $pluginName
+$pluginFiles = @({files})
+VM-Uninstall-IDA-Plugin -pluginName $pluginName -pluginFiles $pluginFiles
```
Ana06, line 254: I do not like that we know always use the `pluginFiles` argument even if it is not needed. I think this complicates the code for the default case making it also confusing. What about adding an `if` to check if `files` is empty in `create_ida_plugin_template` and having two different templates (the previous one and the purposed one)?

scripts/utils/create_package_template.py
```diff
@@ -667,6 +675,9 @@ def main(argv=None):
     # remove type before passing to template create function
     del args.type
 
+    # fixup the files argument
+    args.files = ", ".join(repr(name.strip()) for name in args.files.split(","))
```
Ana06, line 679: This fails if `files` is not provided (it is not required). I purpose to move this code inside `create_ida_plugin_template` (after checking if `files` is empty) as it is only used by that function.

Ana06 (changes requested):
Thanks for all the work and for increasing the version of common :smile:

packages/ida.plugin.dereferencing.vm/tools/chocolateyinstall.ps1
Ana06: does dereferencing fails to uninstall with the current implementation? :thinking:

Schamper: Well technically it doesn't fail, but it doesn't properly uninstall, since it will only remove `dereferencing.py`: https://github.com/mandiant/VM-Packages/blob/7f5911d03424d7da3e47b28915be5c77a4d715fa/packages/ida.plugin.dereferencing.vm/tools/chocolateyuninstall.ps1#L4

The `dereferencing` module folder will remain: https://github.com/danigargu/deREferencing/tree/259685bce8ab4ba0371ec6d13446cc553df2689b/dereferencing

Trying to reinstall the deferencing plugin will then fail because that folder already exists.

PR #1400, opened 2025-05-08
Allows comments in <metadata> of package

### LINKED ISSUE #1396: Allow comments in IncludesRequiredFieldsOnly
### Details

Our [`IncludesRequiredFieldsOnly` linter](https://github.com/mandiant/VM-Packages/blob/af930c6eb0e13cffcca8c037e3469570785a4f80/scripts/test/lint.py#L89) does not allow comments inside `<metadata>`. They can be placed at other levels, for example inside `<package>` (as in [`dotnet-9.vm`](https://github.com/mandiant/VM-Packages/blob/4c668c6368d0d2035ea597b4e2d5a50c1dcec61f/packages/dotnet-9.vm/dotnet-9.vm.nuspec#L5)) and inside `<dependencies>` as in [idapro.vm](https://github.com/mandiant/VM-Packages/blob/af930c6eb0e13cffcca8c037e3469570785a4f80/packages/idapro.vm/idapro.vm.nuspec#L10). We need to adapt the linter to allow comments inside `<metadata>`.

VinceQu:
Added check if the parse node is of "#comment" type.

close #1396

### DIFF
--- a/scripts/test/lint.py
+++ b/scripts/test/lint.py
@@ -105,7 +105,8 @@ class IncludesRequiredFieldsOnly(Lint):
 
         non_required_fields = list()
         for c in metadata.childNodes:
-            if c.nodeName == "#text":
+            # Skip text nodes (whitespace) and comments.
+            if c.nodeName == "#text" or c.nodeName == "#comment":
                 continue
             if c.nodeName not in self.allowed_fields:
                 non_required_fields.append(c.nodeName)
--- a/packages/dotnet-6.vm/dotnet-6.vm.nuspec
+++ b/packages/dotnet-6.vm/dotnet-6.vm.nuspec
@@ -1,10 +1,10 @@
 <?xml version="1.0" encoding="utf-8"?>
 <package xmlns="http://schemas.microsoft.com/packaging/2015/06/nuspec.xsd">
-  <!-- Metapackage for .NET6 to ensure all packages use the same version.-->
   <metadata>
     <id>dotnet-6.vm</id>
-    <version>0.0.0.20250505</version>
+    <version>0.0.0.20250509</version>
     <authors>Microsoft</authors>
+    <!-- Metapackage for .NET6 to ensure all packages use the same version.-->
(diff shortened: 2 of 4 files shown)


scripts/test/lint.py
```diff
@@ -105,7 +105,7 @@ def check(self, path):
 
         non_required_fields = list()
         for c in metadata.childNodes:
-            if c.nodeName == "#text":
+            if c.nodeName == "#text" or c.nodeName == "#comment":
```
Ana06, line 108: [nitpick] What about adding a comment above this line to explain what is does to ensure it is clear to others and it does not get removed by accident in the future? :thinking: Something like:

```suggestion
            # Skip text nodes (whitespace) and comments.
            if c.nodeName == "#text" or c.nodeName == "#comment":
```

Ana06 (changes requested):
Thanks a lot for your contribution @VinceQu! :tada: 

@VinceQu can you please add to this PR a commit to move the following comment above the description field in `dotnet-6.vm.nuspec`, `dotnet-8.vm.nuspec`, and `dotnet-9.vm.nuspec` so that we can test that the change works?
```
<!-- Metapackage for .NET6 to ensure all packages use the same version.-->
```

[nitpick] I recommend you to check https://www.freecodecamp.org/news/how-to-write-better-git-commit-messages/ for good practices when writing commits. Concretely, in your commit I would use imperative mood in the subject line and add a description that explains what and why for your changes. For example :wink: 
```
fix: Lint allows comments in <metadata>

The linter was incorrectly flagging package nuspecs as invalid if they 
contained comments within the `<metadata>` section. Update the linter to
ignore comment nodes (`#comment`) when checking for allowed fields.
```

You can for example use `git commit --amend` to modify the commit message of the last commit. You can use `git commit rebase -i HEAD~2` to modify the latest 2 commits and change the commit message with `reword`.

Ana06 (approved):
This looks great, thanks @VinceQu! 💐

PR #1411, opened 2025-05-27
Fixed and updated memprocfs.vm to latest version

socketz:


packages/memprocfs.vm/tools/chocolateyinstall.ps1
```diff
@@ -4,7 +4,7 @@ Import-Module vm.common -Force -DisableNameChecking
 $toolName = 'MemProcFS'
 $category = VM-Get-Category($MyInvocation.MyCommand.Definition)
 
-$zipUrl = 'https://github.com/ufrisk/MemProcFS/releases/download/v5.14/MemProcFS_files_and_binaries_v5.14.11-win_x64-20250501.zip'
-$zipSha256 = '40ce4ac604933dd919444a9cee42431b3fe43813fa4155bef17466b7db03d764'
+$zipUrl = 'https://github.com/ufrisk/MemProcFS/releases/download/v5.14/MemProcFS_files_and_binaries_v5.14.13-win_x64-20250520.zip'
```
Ana06, line 7: The URL have changed again. We have the same issue in https://github.com/mandiant/VM-Packages/pull/1420.

We need to either downgrade to v5.13.4 (as the asset of that version doesn't seem to change anymore) or the help of MemProcFS by changing how the releases are done. I have sent an issue to see if they are able to help: https://github.com/ufrisk/MemProcFS/issues/389

Ana06 (changes requested):
Thanks for the contribution @socketz.

packages/memprocfs.vm/tools/chocolateyinstall.ps1
```diff
@@ -3,8 +3,41 @@ Import-Module vm.common -Force -DisableNameChecking
 
 $toolName = 'MemProcFS'
 $category = VM-Get-Category($MyInvocation.MyCommand.Definition)
+# Get latest version from GitHub releases with a curl equivalent in powershell
+$latestFileName = 'MemProcFS_files_and_binaries-win_x64-latest.zip'
+$latestVersionUrl = 'https://api.github.com/repos/ufrisk/MemProcFS/releases/latest'
```
Ana06, line 8: I have concerns about using `https://api.github.com/repos/ufrisk/MemProcFS/releases/latest`, as this means we potentially update to major versions without any tool testing, which could cause issues with newer dependency requirements. I think we should use a concrete version and allow our bot to update it so that we can test it by using `https://github.com/ufrisk/MemProcFS/releases/download/v5.15/MemProcFS_files_and_binaries-win_x64-latest.zip`.

packages/memprocfs.vm/tools/chocolateyinstall.ps1
```diff
@@ -3,8 +3,41 @@ Import-Module vm.common -Force -DisableNameChecking
 
 $toolName = 'MemProcFS'
 $category = VM-Get-Category($MyInvocation.MyCommand.Definition)
+# Get latest version from GitHub releases with a curl equivalent in powershell
```
Ana06, line 6: Use `Get-ChocolateyWebFile` chocolatey helper instead of curl, as [we do when we use signature verification](https://github.com/mandiant/VM-Packages/blob/8d2ee4f6704f7745b330dd4d5ac4f939885f321f/packages/sysinternals.vm/tools/chocolateyinstall.ps1#L15)

packages/memprocfs.vm/tools/chocolateyinstall.ps1
```diff
@@ -3,8 +3,41 @@ Import-Module vm.common -Force -DisableNameChecking
...
+$latestFileName = 'MemProcFS_files_and_binaries-win_x64-latest.zip'
+$latestVersionUrl = 'https://api.github.com/repos/ufrisk/MemProcFS/releases/latest'
 
-$zipUrl = 'https://github.com/ufrisk/MemProcFS/releases/download/v5.14/MemProcFS_files_and_binaries_v5.14.11-win_x64-20250501.zip'
-$zipSha256 = '40ce4ac604933dd919444a9cee42431b3fe43813fa4155bef17466b7db03d764'
+if (-not (Get-Command Invoke-RestMethod -ErrorAction SilentlyContinue)) {
+    Write-Error "Invoke-RestMethod is not available. Please ensure you are running PowerShell 3.0 or later."
```
Ana06, line 11: We require PowerShell >= 5: https://github.com/mandiant/flare-vm?tab=readme-ov-file#requirements

packages/memprocfs.vm/tools/chocolateyinstall.ps1
```diff
@@ -3,8 +3,41 @@ Import-Module vm.common -Force -DisableNameChecking
...
+    $latestVersionResponse = (Invoke-WebRequest -Uri $latestVersionUrl -Headers @{ 'User-Agent' = 'VM-Installer'}).content | ConvertFrom-Json
+}else {
+    $latestVersionResponse = Invoke-RestMethod -Uri $latestVersionUrl -Headers @{ 'User-Agent' = 'VM-Installer' }
+}
+
+# Get the latest version's SHA256 from the assets searching in the array by the name key and getting the digest
```
Ana06, line 17: Normally we have a hardcoded hash we use for verification, while here it is being proposed to extract the hash for the GitHub release assets. Providing the hashes seems to be a new GH feature and the hash is calculated automatically: https://github.blog/changelog/2025-06-03-releases-now-expose-digests-for-release-assets/ I have concerns about verifying the download in this way and would prefer to use a hardcoded URL (we may need to clarify what we need in https://github.com/ufrisk/MemProcFS/issues/389 or use an older version that is not updated anymore). what do others think? @vm-packages @emtuls @naacbin

If we decide it is ok to verify the download in this way, we need to document in this package WHY we are doing it such a special way in this case: MemProcFS updates the release assests without increasing the version when Microsoft updates the cached symbol offsets.

emtuls: From reading @Ana06's issue for MemProcFS (https://github.com/ufrisk/MemProcFS/issues/389), it sounds like the developer isn't willing to provide the tool in a way that is preferred for our scripts, which is unfortunate.

I might be okay with grabbing the sha256 dynamically from the asset release though I would usually prefer not to do it this way for consistencies sake, but this seems to be a slightly special case for a highly used tool and this might be the easiest way to keep the tool up to date (if that is our intent).
I don't use the tool enough to know if using an older version would work consistently, as it sounds like things are based off of offsets which can change often. If an older version works well for most recent Windows versions, then I'd prefer to stick with a slightly older version for a longer period of time. Not sure what the trade-off is if we don't keep the tool at the latest update.
Could someone chime in and let me know how crucial it is to have the most recent update for this tool?

If we do go the route of the dynamic hash grabbing, then the minor version may need to be modified as well due to the releases only being something like `X.XX` rather than the `X.XX.X` style that some assets have in the name.

Ana06: @emtuls I think the developer was very supportive and willing to help us, but maybe it was not clear what we exactly need.

PR #1414, opened 2025-05-27
Updated binwalk from python version to rust version

socketz:


emtuls:
@socketz Thank you for this!

It looks like there are just a couple of small conflicts that have occurred. If you could fix those, I can look into merging this ASAP :)

PR #1417, opened 2025-06-05
WxTCmd updated hash

ninyya:
The download for WxTCmd has been updated while staying on version 1.1.0.
Updated to handle the new hash.


Ana06 (dismissed):
Thanks for the hash fix @ninyya! @ninyya can you please squash your commits into one to help us keeping a clean commit history?

ninyya:
Fixed the version as requested and squashed the commits.
I believe the current task failures is due to the version downgrade that you mentioned @Ana06

emtuls (approved):
Thanks for the update and commits @ninyya!

Ana06:
@emtuls the option is also gone for me, it could be because the run is too old :confused:  To force the CI to re-run you (or @ninyya) can also re-push after forcing a new commit hash (for example rebasing will change the hash of the commit). It is important to do this fast as this package is now broken (users can't install it as it is not in MyGet anymore).

Ana06:
Thanks @emtuls @ninyya! :rocket:

PR #1420, opened 2025-06-06
fixed dead link of memprocfs and updated hash

Jinsakai-25:
Updated the MemprocFs dead github zip link to the new one and calculated and stored the new hash value too.


Ana06 (changes requested):
Thanks for updating the contribution @Jinsakai-25, but note this is a duplicated of https://github.com/mandiant/VM-Packages/pull/1411 and that both PR are not working, as the URL has changed again.

We need to either downgrade to v5.13.4 (as the asset of that version doesn't seem to change anymore) or the help of MemProcFS by changing how the releases are done. I have sent an issue to see if they are able to help: https://github.com/ufrisk/MemProcFS/issues/389

Closing this PR in favour of https://github.com/mandiant/VM-Packages/pull/1411 as it was created before.

Issue #1398, opened 2025-05-07
update_package.py: Always increase version


Ana06:
### Details

The bot wrongly updated the version of memprocfs.vm from 5.14.9 to 5.14 in https://github.com/mandiant/VM-Packages/pull/1361.

https://github.com/mandiant/VM-Packages/commit/077ee1aff091cec3996f487ea21313063905aff7

I think we should add a check in [`update_package.py`](https://github.com/mandiant/VM-Packages/blob/main/scripts/utils/update_package.py) that ensures the version is increased.

Jinsakai-25:
hello Mam , I have got some experience working with Vm Packages and would love to contribute by fixing this issue .Could you kindly assign it to me ?

PR #1426, opened 2025-06-15
Fix to issue #1398 in update_package.py

Jinsakai-25:
the script Now directly pastes Memprocfs the link extracted from github api and fetches the version directly from link and not from tag_name . I have done it now only for Memprocfs as it is the problematic one , i am currently testing for others

### DIFF
--- a/scripts/utils/update_package.py
+++ b/scripts/utils/update_package.py
@@ -25,9 +25,7 @@ def replace_version(latest_version, nuspec_content):
         except ValueError:
             # not all tools use symver, observed examples: `cp_1.1.0` or `current`
             print(f"unusual version format: {latest_version}")
-            print(
-                "reusing old version with updated date, manual fixing may be appropriate"
-            )
+            print("reusing old version with updated date, manual fixing may be appropriate")
             latest_version = version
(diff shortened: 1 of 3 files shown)


scripts/utils/update_package.py
```diff
@@ -39,6 +43,15 @@ def get_latest_version(org, project, version):
     if not response.ok:
         print(f"GitHub API response not ok: {response.status_code}")
         return None
+    if org == "ufrisk" and project == "MemProcFS":
```
Ana06, line 46: It is not desirable to add code for only two packages, the code should be generic.

Ana06 (commented):
@Jinsakai-25 I appreciate your work on this, but with https://github.com/mandiant/VM-Packages/issues/1398 we want to ensure that the version is increased in automatic updates.This PR adds code to handle only two concrete packages, which is not desirable as we want the code to be generic, making it easy to maintain.

 https://github.com/mandiant/VM-Packages/pull/1467 runs the linter in the automatic updates to avoid this and potentially other issues, closing ttps://github.com/mandiant/VM-Packages/issues/1398.

Issue #1366, opened 2025-04-22
nmap fails to update


sara-rn:
### Details

The package nmap fails to update, as per the package update action see the log below:
```
2025-04-16T23:04:46.0679134Z nmap.vm v7.94.0 (forced)
2025-04-16T23:04:46.0694004Z nmap.vm package files upgrade completed. Performing other installation steps.
2025-04-16T23:04:47.0336048Z Downloading nmap.vm 
2025-04-16T23:04:47.0336524Z   from 'https://nmap.org/dist/nmap-7.94-setup.exe'
2025-04-16T23:04:48.5419115Z 
2025-04-16T23:04:48.5427256Z Download of nmap-7.94-setup.exe (30.96 MB) completed.
2025-04-16T23:04:51.6276368Z Hashes match.
2025-04-16T23:04:51.6317452Z C:\Users\runneradmin\AppData\Local\Temp\chocolatey\nmap-7.94-setup.exe
2025-04-16T23:05:08.0308063Z Added C:\ProgramData\chocolatey\bin\ncat.exe shim pointed to 'c:\tools\nmap\ncat.exe'.
2025-04-16T23:05:08.2198490Z Added C:\ProgramData\chocolatey\bin\nmap.exe shim pointed to 'c:\tools\nmap\nmap.exe'.
2025-04-16T23:05:08.4187128Z Added C:\ProgramData\chocolatey\bin\nping.exe shim pointed to 'c:\tools\nmap\nping.exe'.
2025-04-16T23:05:08.4904826Z Cannot find path 'C:\Tools\nmap\zenmap.exe' because it does not exist.
2025-04-16T23:05:08.4905535Z At C:\ProgramData\chocolatey\lib\nmap.vm\tools\chocolateyinstall.ps1:47 char:23
2025-04-16T23:05:08.4906156Z +     $executablePath = Join-Path $toolDir "zenmap.exe" -Resolve
2025-04-16T23:05:08.4906594Z +                       ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
2025-04-16T23:05:08.5912132Z ERROR: Cannot find path 'C:\Tools\nmap\zenmap.exe' because it does not exist.
2025-04-16T23:05:08.7842524Z   nmap.vm may be able to be automatically uninstalled.
2025-04-16T23:05:08.7904157Z Environment Vars (like PATH) have changed. Close/reopen your shell to
[...]

PR #1439, opened 2025-07-09
Fix(ci): Ensure clean state between package tests

emtuls:
The package update workflow was vulnerable to state pollution, causing false-positive test results. A successful test of one version could leave artifacts on the file system that would cause a subsequent test of a new, broken version to incorrectly pass.

This was observed when a successful installation of `nmap v7.93` left on the system, which then caused the test for the broken `v7.94` package to succeed because the file already existed.

## Original (outdated) change
The original change/fix I came up with was to add a `choco uninstall` command after each test run. This ensures every package test starts in a clean environment, preventing artifacts from one test from influencing another and guaranteeing that installation failures are accurately detected.

Working change in logs showing `nmap v7.94` fails after uninstalling `v7.93`:
https://github.com/emtuls/VM-Packages/actions/runs/16158436802/job/45605392190#step:5:1028

New Workflow: https://github.com/emtuls/VM-Packages/actions/runs/16158436802/workflow

Previous version that shows `v7.93` and `v7.94` working in logs:
https://github.com/emtuls/VM-Packages/actions/runs/16158000438/job/45604210877#step:5:1138

Old Workflow: https://github.com/emtuls/VM-Packages/actions/runs/16158000438/workflow

### Problem
This left a secondary issue of taking the version of last `$UPDATE_TYPE` to succeed rather than the highest one if multiple types succeeded, which led to my second change, which is a bit more robust.
## 

## New (updated) change
I updated the workflow file to be slightly more improved so that it checks for and tests only the highest found version, as there is a possibility that multiple `$UPDATE_TYPE` could work and result in different versions, and the current workflow would select the last working update type even if the version was not the highest.

New run with update for finding and testing only highest version (linking directly to `nmap` to show that it correctly fails after finding and testing the latest version `7.94`):
https://github.com/emtuls/VM-Packages/actions/runs/16159022853/job/45606992656#step:5:587

Workflow File: https://github.com/emtuls/VM-Packages/actions/runs/16159022853/workflow

New Package Update PR that works: https://github.com/emtuls/VM-Packages/pull/2
##

This addresses the primary issue in https://github.com/mandiant/VM-Packages/issues/1366, but there is still the underlying issue that `nmap v7.94` fails to properly install `zenmap`, which seems to possibly be a known issue?

The PR is not from my fork due to some other issues in my repo I need to resolve after pulling the latest commits that would have caused issues with this PR.

### DIFF
--- a/.github/workflows/update_packages.yml
+++ b/.github/workflows/update_packages.yml
@@ -27,7 +27,7 @@ jobs:
             $package = $packagePath.Name
             $newVersion = 0
             # Test independently every type of update and commit what works
-            foreach ($UPDATE_TYPE in ('DEPENDENCIES', 'GITHUB_URL', 'VERSION_URL', 'DYNAMIC_URL', 'MSIXBUNDLE_URL')) {
+            foreach ($UPDATE_TYPE in ('GITHUB_URL', 'MSIXBUNDLE_URL', 'VERSION_URL', 'DYNAMIC_URL', 'DEPENDENCIES')) {
               $version = python scripts\utils\update_package.py $package --update_type $UPDATE_TYPE
               $updated = $?
               echo "$package $version"
@@ -49,6 +49,10 @@ jobs:
               # Clean changes and built packages
               git restore .
               Remove-Item built_pkgs -Recurse -ErrorAction Ignore
+              if ($tested) {
+                # Only allow 1 successful package install
+                break
+              }
             }
             if ($newVersion) {
               if ($finalUpdateType -eq 'DYNAMIC_URL') {


emtuls:
This may still need a little bit of thought as the current state of `update_packages.yml` will cycle through each `$UPDATE_TYPE` and will take the version of whichever the last one that works in the order it performs.

It is uncommon that multiple types would work, but as we can see, it does occur. I think we should we consider a small refactor to account for this and to only test and commit the highest version that we find.

I'll resubmit the PR with this change.

Ana06 (changes requested):
Thanks for all the research and providing as much details @emtuls! 

The issue I see with the proposed solution is that it complicates the code and keeping the code of the actions simply makes it easier to maintain and fix future bugs. Alternative simpler proposal: **Do not allow more than one type of update for the sample package in the same PR update**. The same package should only work for one type (or produce the same result if it works for more than one type), maybe with the exception of the dependencies, which could be updated in a future update. For this:
- Exit the `foreach` loop after one type of package has succeeded to install
- Sort the updates types by priority (try more reliable first and update dependencies at the end):
```
foreach ($UPDATE_TYPE in ('GITHUB_URL', 'MSIXBUNDLE_URL', 'VERSION_URL', 'DYNAMIC_URL', 'DEPENDENCIES')) {
```
- Document we only support 1 type of update at once to prevent issues and that the types are ordered by priority

.github/workflows/update_packages.yml
```diff
@@ -41,6 +41,8 @@ jobs:
                   $newVersion = $version
                   # Save the update type to use it in the commit
                   $finalUpdateType = $UPDATE_TYPE
+                  # Only allow 1 successful package install
+                  break
```
Ana06, line 45: this skips the cleaning of built packages, doesn't it?

Ana06 (changes requested):

emtuls:
> @emtuls
> 
> > This may cause an issue where a package fails to install at a higher version but completes at a lower version (assuming 2 types produce different versions) and we may miss that a higher version fails to install and why due to the completion of a lower version.
> 
> I am not sure if I understand this part. can you please elaborate? That two types of updates work for different versions is very unlikely (except for dependencies maybe, that we have placed at the end). If the update fails we miss it independently of another lower version succeeding, don't we?

Ah, you might be right. I assumed that we could have multiple types of potential ways to update 1 package, but it's really only dependency and one type, so that changes things. :)

.github/workflows/update_packages.yml
emtuls: That is correct. I missed that! Thank you. :)

Ana06 (approved):
LGTM, thanks @emtuls! I'll merge it and trigger the package update to see it working :wink:

PR #1453, opened 2025-07-17
[ci: fix] Linter fails for new package

Ana06:
If the package is not found in MyGet, the local version is the first version of the package (new package) and the `VersionNotIncreased` linter should not fail. The bug was introduced in https://github.com/mandiant/VM-Packages/pull/1437 when we started checking that the local version is higher than the one in MyGet, instead of only that the version is modified.


Ana06:
As said above, ALL PRs fail including this one because of this bug (as I have deleted vb-decompiler-lite.vm from MyGet). As this PR is not updating any packages (just the linter that is run in the CI), we can merge it by bypassing the checks.

Issue #354, opened 2023-05-25
Add INSTALLER_EXE package type


Ana06:
Add a new `INSTALLER_EXE` package type to `create_package_template.py` and the issue template type to create packages like the following automatically:
- [vbdec.vm](https://github.com/mandiant/VM-Packages/blob/e5e526c9738e8535bda49063496d87dfc5abc647/packages/vbdec.vm/tools/chocolateyinstall.ps1)

PR #1461, opened 2025-07-28
[doc] Add installer type to new_package.yml

Ana06:
There is no type for installer as we do not support it in the automation (yet). The current issue template documentation does not mention installers, causing that users select the wrong type (`SINGLE_EXE`) which makes the automation fail when trying to create the package. Improve the documentation to mention installers explicitly until we support this package type (we have an issue for this: https://github.com/mandiant/VM-Packages/issues/354)


Ana06:
@mlhaigh let me know if you have any feedback

stevemk14ebr (approved):

PR #1470, opened 2025-07-29
Install vcredist140.vm before visualstudio2017-workload-vctools

### LINKED ISSUE #1440: packages fail to install due to newer version of vcredist140
### Details

As @emtuls has explained in https://github.com/mandiant/VM-Packages/issues/1371#issuecomment-3067276106, running `choco install pdbs.pdbresym.vm` after the FLARE-VM installation fails because a newer version of vcredist140 is installed. All `.vm` packages that require vcredist140 use the [vcredist140.vm](https://github.com/mandiant/VM-Packages/blob/b9fe8c66f433b0631692ecf36f1e7b15ab63483a/packages/vcredist140.vm/vcredist140.vm.nuspec#L10) package to install the same version of vcredist140 and prevent issues like this. This means that likely a package in the default FLARE-VM configuration is indirectly installing a newer version. We need to investigate which package this is to adapt the vcredist140 version we require in vcredist140.vm.

d35ha:
Closes #1440.
Still needs to be tested on a freshly installed FLARE-VM.


emtuls:
This seems to mostly work but I think to finalize it, each of the packages that requires `vcbuildtools.vm` as a dependency will need to be updated to ensure they use `version 0.0.0.20250729`.

`<dependency id="vcbuildtools.vm" version="0.0.0.20250729"/>`

The packages in question seem to be:
- binaryninja.vm
- evtxecmd.vm
- googlechrome.vm
- libraries.python3.vm
- metasploit.vm
- pecmd.vm
- recmd.vm
- registry_explorer.vm
- rla.vm
- sqlecmd.vm
- sysinternals.vm

packages/vcbuildtools.vm/vcbuildtools.vm.nuspec
```diff
@@ -2,13 +2,13 @@
...
     <id>vcbuildtools.vm</id>
-    <version>0.0.0.20250228</version>
+    <version>0.0.0.20250729</version>
     <description>Metapackage that requires the dependencies visualstudio2017buildtools and visualstudio2017-workload-vctools</description>
     <authors>Mandiant, Microsoft</authors>
     <dependencies>
       <dependency id="common.vm"  version="0.0.0.20250206" />
       <dependency id="visualstudio2017buildtools" version="[15.9.58, 15.9.59)" />
-      <dependency id="visualstudio2017-workload-vctools" version="[1.3.3]" />
+      <dependency id="visualstudio2017-workload-vctools.vm" />
```
Ana06, line 11: Adding a metapackage for `visualstudio2017-workload-vctools` to require `vcredist140.vm` in it should be the same as adding it directly in `vcbuildtools.vm` (here as dependency) as it is the only place where you are using it. This is what I proposed in https://github.com/mandiant/VM-Packages/issues/1440#issuecomment-3132826642 and I think it is a better solution as adding extra packages makes following dependencies harder and increase the risk of dependency loops. Or am I missing something @d35ha?

```suggestion
      <!-- Ensure visualstudio2017-workload-vctools installs a vcredist140 version compatible with the one required in vcredist140.vm -->
      <dependency id="vcredist140.vm" />
```

Ana06 (changes requested):

packages/vcbuildtools.vm/vcbuildtools.vm.nuspec
d35ha: Done.

emtuls (approved):
Looks good to me! A quick test with installing the new `vcbuildtools.vm` followed by `binaryninja.vm` and then `bindiff.vm` looks to work. :)

Issue #1371, opened 2025-04-23
Improve shortcuts in visualstudio.vm and microsoft-office.vm


Ana06:
### Details

Improve shortcuts in visualstudio.vm and microsoft-office.vm fixing https://github.com/mandiant/flare-vm/pull/660#issuecomment-2657169466

Ana06:
Improving the shortcuts does not fix the issue encountered in https://github.com/mandiant/flare-vm/pull/660#issuecomment-2657169466. Sometimes the shortcuts are created as expected, but other times the shortcuts are not created or are created "broken" (without a target). This does not happen during the FLARE-VM installation, only when using VirtualBox scripting to install the packages from outside the VM.

Ana06:
@emtuls 
> I was not able to get the VM to be exported (even with creating the expected default folder), so I couldn't test if it happened somehow during that process.

I experienced the issue in the snapshots too, so the issue is not related to the VM export.

Ana06:
The issue with the vcredist140 version that causes pdbs.pdbresym.vm to install is important, although unrelated to the shortcuts. I have created a separate issue to track it, as we need to solve it ensuring all our packages install the same version of vcredist140: https://github.com/mandiant/VM-Packages/issues/1440

Ana06:
I have just ran the build script and I still get broken shortcuts 😕

Ana06:
The issue is not with the shortcut creating but with the build process. I'll send a fix.

PR #1487, opened 2025-08-04
[VM-Install-Shortcut] Enhance code and ensure shortcut is created 

Ana06:
Re-create shortcut using `ComObject` as in some weird cases the shortcut created by `Install-ChocolateyShortcut` does not have arguments.

To implement it I am refactoring `VM-Install-Shortcut`. Any bug introduced in this refactoring could break the icons of the tools, so we need proper testing before merging.

Aims to fix https://github.com/mandiant/VM-Packages/issues/1371.

### DIFF
--- a/packages/common.vm/tools/vm.common/vm.common.psm1
+++ b/packages/common.vm/tools/vm.common/vm.common.psm1
@@ -269,16 +269,26 @@ function VM-Install-Shortcut{
     Install-ChocolateyShortcut @shortcutArgs
     VM-Assert-Path $shortcut
 
+    # Ensure shortcut has target as in some weird cases (e.g. with virtualbox scripting) the shortcut is created without the arguments
+    $shell = New-Object -ComObject ("WScript.Shell")
+    $shortcutObject = $shell.CreateShortcut($shortcut)
+    if (-Not $shortcutObject.TargetPath) {
+        VM-Write-Log "WARN" "$shortcut has no target, trying to update shortcut using ComObject"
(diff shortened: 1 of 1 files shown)


packages/common.vm/tools/vm.common/vm.common.psm1
```diff
@@ -231,63 +231,62 @@ function VM-Install-Shortcut{
...
+    $shell = New-Object -ComObject ("WScript.Shell")
+    $shortcutObject = $shell.CreateShortcut($shortcut)
+    if (-Not $shortcutObject.TargetPath) {
+        VM-Write-Log "WARN" "$shortcut has no target, trying to update shortcut using ComObject"
+        $shortcutObject.TargetPath = $executablePath
+        $shortcutObject.Arguments = $arguments
+        $shortcutObject.WorkingDirectory = $executableDir
+        $shortcutObject.IconLocation = $iconLocation
+        $shortcutObject.Save()
+    }
```
emtuls, line 282: I think this is still missing adding the `RunAsAdmin` part for the new shortcut creation like [here](https://github.com/mandiant/VM-Packages/pull/1487/files#diff-42c583682dc7cee57079eab399b94fbe5a60b1fcec5626a25a7092e197cb7308R265-R267)

emtuls (changes requested):
Hard to really test this, but it looks good except for I think it's missing the `RunAsAdmin` for the backup shortcut creation part. :)

PR #1502, opened 2025-08-20
Update x64dbg

mrexodia:
- Since the previous release we started using CalVer and actual GitHub releases that do not get overwritten
- For a few years it has been possible to create a `userdir` file next to x64dbg.exe (as well as x32dbg.exe). This allows package managers to install x64dbg in a non-writable location and it will put the databases in appdata. I did not implement this here, but might be helpful. Source code: https://github.com/x64dbg/x64dbg/blob/85e0ff85796891ebfd3cc5bb46daedbe7ef6098c/src/bridge/bridgemain.cpp#L113


Ana06 (changes requested):
@mrexodia thanks for the contribution and sorry for the late reply. I have tested locally and it works as expected :rocket: I have just a small nitpick to make the description following our documentation.

PR #1515, opened 2025-10-20
Check if the issue templates package categories are updated

### LINKED ISSUE #298: Keep categories.txt and issue templates in sync
Discussion from https://github.com/mandiant/VM-Packages/pull/291. Ideas to ensure that `categories.txt` and issue templates are kept in sync :
- Migrate `categories.txt` to Python so that we can add a comment (about the need of updating the issue template)  and simplify the reading code (e.g. by making it a list directly) @mr-tz 
- Add a check to the CI that ensure the issue templates and `categories.txt` are in sync. 

We are not sure if the effort is worth as we don't plan to update the categories often. Tracking the ideas here in case we see this is needed.

d35ha:
Fixes #298.


emtuls (approved):
I did not properly test this, but it seems fine. I'm not sure how the `exit(1)` and `exit(0)` will effect things (technically, the `exit(0)` part is redundant and not needed), but if there is an issue, we can revert back.

emtuls:
I think there may be some issues with this @d35ha 

See the following: 
https://github.com/mandiant/VM-Packages/actions/runs/19484039748/job/55762142469 
https://github.com/mandiant/VM-Packages/actions/runs/19484101428/job/55762333313#step:7:20


Going to revert back for now and test it better later. :)

PR #1517, opened 2025-10-20
Improve the output logging for VM-Set-Legal-Notice and VM-Clean-Up

### LINKED ISSUE #1372: Improve `vbox-build-flare-vm.py` output
### Details

From https://github.com/mandiant/flare-vm/pull/660: Reduce noisy output from certain commands within the build VM script, such as `VM-Clean-Up` and `VM-Set-Legal-Notice`, to improve readability of `vbox-build-flare-vm.py` output.

d35ha:
Fixes #1372.

### DIFF
--- a/packages/common.vm/common.vm.nuspec
+++ b/packages/common.vm/common.vm.nuspec
@@ -2,7 +2,7 @@
 <package xmlns="http://schemas.microsoft.com/packaging/2015/06/nuspec.xsd">
   <metadata>
     <id>common.vm</id>
-    <version>0.0.0.20250828</version>
+    <version>0.0.0.20251015</version>
     <description>Common libraries for VM-packages</description>
     <authors>Mandiant</authors>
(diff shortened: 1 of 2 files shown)


d35ha:
> Thank you for the improvements!
> 
> I think there's only a single small change I would make. For `VM-Set-Legal-Notice`, we should have a single message saying the note has been applied and maybe the legal notice text. :)

Done, the legal notice might be multiple lines though.

emtuls:
> the legal notice might be multiple lines though.

No worries, and thank you! We can look into adjusting it later if it is still an annoyance. I wanted to make sure we hit the ask that @Ana06 mentioned in https://github.com/mandiant/VM-Packages/issues/1372#issuecomment-3158443663 of:
> We should replace it by a single message saying the note has been applied and maybe the legalnotice text.

emtuls (approved):

